PR #127, opened 2022-12-07
Advanced query params in url

baiirun:


apps/web/modules/services/import.ts
```diff
@@ -971,14 +971,14 @@ function convertSanFranciscoData(
...
-      row[4] ? [row[4], 'type', 'attribute'] : null,
+      row[4] ? [row[4], 'type', 'related to'] : null,
       row[8] ? [row[2], 'author', row[8]] : null,
       row[8] ? [row[8], 'name', row[8]] : null,
       row[8] ? [row[8], 'type', 'author'] : null,
 
       ...isAboutTuples.map(({ name }): EavRow => [row[2], 'is about', name.toLowerCase()]),
       ...isAboutTuples.map(({ name }): EavRow => [name.toLowerCase(), 'name', name]),
-      ...isAboutTuples.map(({ name }): EavRow => [name.toLowerCase(), 'type', 'attribute']),
+      ...isAboutTuples.map(({ name }): EavRow => [name.toLowerCase(), 'type', 'is about']),
```
baiirun, line 981: We were incorrectly setting the is about and related to relations to type attribute

PR #131, opened 2022-12-13
Edit mode

baiirun:


apps/web/next.config.js
```diff
@@ -7,7 +7,7 @@ const withBundleAnalyzer = analyzer({
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
-  reactStrictMode: true,
+  // reactStrictMode: true,
```
baiirun, line 10: Rainbowkit currently suffers from the double-effect running issue that was introduced in Strict Mode in React 18. This resulted in connect/disconnect dialogs popping up when dealing with wallets when they shouldn't have appeared. Turning it off fixes it for now until we implement a better wallet solution.

apps/web/pages/dev.tsx
```diff
@@ -59,7 +59,7 @@ export default function Dev() {
         <a>Spaces</a>
       </Link>
 
-      <ConnectButton accountStatus="avatar" />
+      <ConnectButton />
```
baiirun, line 62: Leaving the default Connect button in the dev route so we can change chains there.

apps/web/modules/utils.ts
```diff
@@ -26,18 +27,29 @@ export function titleCase(string: string): string {
     .join(' ');
 }
 
-export const navUtils = {
+export const NavUtils = {
```
baiirun, line 30: In the near future I'd like us to move towards better namespacing of modules. Right now we kinda have things all over the place, mostly as a side-effect of the file structure of the repo being leftovers from the early prototyping days. This and the Params change is the first step.

PR #135, opened 2022-12-15
fetchEntities combines both name_starts_with_nocase and name_contains_nocase

makeitrein:


apps/web/modules/services/network.ts
```diff
@@ -217,11 +226,36 @@ export class Network implements INetwork {
...
+      }
+      return a.length - b.length;
+    };
+
+    const primaryResults = startEntities.sort((a, b) => sortLengthThenAlphabetically(a.name, b.name));
+    const secondaryResults = containEntities
+      .filter(entity => !startEntityIds.includes(entity.id))
+      .sort((a, b) => sortLengthThenAlphabetically(a.name, b.name));
+
+    return [...primaryResults, ...secondaryResults];
```
dabbott, line 258: I might move all the sorting logic to a separate function like

```ts
// Assume entities starting with the term are most relevant
function sortSearchResultsByRelevance(startEntities, containEntities) {
  ...
  return [...primaryResults, ...secondaryResults]
}
```

If we were gonna keep this behavior longer term I'd add a unit test, but probably not needed

dabbott (approved):
Seems like a reasonable improvement for now. My wallet doesn't have edit access at the moment so I didn't try it, but I assume it's better

apps/web/modules/services/network.ts
makeitrein: Done!

dabbott (approved):

PR #137, opened 2022-12-19
Env cookie

baiirun:
This change sets a cookie if the developer has set the environment config using the `?env=<environment>` URL parameter.

We've changed to doing SSR for most of our pages, and setting the configuration using the wallet is especially annoying when developing using the non-production chain. The `?env` escape hatch allows the developer to manually set the configuration they want to use for a given page. Unfortunately, we would have to do this for every single route that used SSR, often manually having to add the parameter after navigating.

Now, once the developer has the configuration, it gets stored in a cookie that the other SSR'd routes read from when fetching data.

This is the order of priority when reading the environment:
1. The `?env` parameter
2. The cookie
3. Defaults to production if neither the parameter or cookie are set.


apps/web/pages/space/[id].tsx
```diff
@@ -33,7 +33,7 @@ export default function TriplesPage({ spaceId, spaceName, spaceImage, initialTri
 export const getServerSideProps: GetServerSideProps<Props> = async context => {
   const spaceId = context.params?.id as string;
   const initialParams = Params.parseQueryParameters(context.resolvedUrl);
-  const config = Params.getConfigFromUrl(context.resolvedUrl);
+  const config = Params.getConfigFromUrl(context.resolvedUrl, context.req.cookies[Params.ENV_PARAM_NAME]);
```
Comment (baiirun) [inline apps/web/pages/space/[id].tsx:36]: We may be able to make some getServerSideProps wrapper that handles common middleware like getting the config, but I'm leaving it as is for now since there's not a _ton_ of duplication.

PR #139, opened 2022-12-20
Entities Table (Read-Only / Limited Filtering)

makeitrein:
High level notes:

image

- Displays the new entities table on the hidden `/entities` route
- Breaks up monolithic TriplesTable into SpaceHeader, SpaceNavbar, SpaceActions
- New fetchEntityTableData method on Network for fetching rows and columns
- New Row and Column and Cell types
- Entity Table component and store forked from Triple Table store, made to support arbitrary columns, multiple triple values in a cell, typeID filtering
- New UUID entities ATTRIBUTES and SCHEMA_TYPE defined in both constants and subgraph bootstrap files


apps/web/modules/components/entity-table/entity-table.tsx
```diff
@@ -0,0 +1,202 @@
...
+    }, [initialCellData]);
+
+    if (!cellData) {
+      return <div />;
+    }
+
+    return (
+      <Entities>
+        {cellData.triples.map(({ value, attributeId, entityId, entityName }) => {
+          if (attributeId === 'name') {
```
baiirun, line 96: Should we be using the `columnId` to determine what to render instead of the `attributeId`?

apps/web/modules/components/space/space-navbar.tsx
```diff
@@ -0,0 +1,89 @@
...
+    {
+      name: 'Triples',
+      href: `/space/${query.id}/triples`,
+      selected: tabTriplesSelected,
+    },
+  ];
+
+  return (
+    <NavbarContainer>
+      <div>
```
baiirun, line 79: Can we use `gap` or a `Spacer` here instead of encoding spacing in the `TabLink` component?

apps/web/modules/design-system/tab-link.tsx
```diff
@@ -0,0 +1,29 @@
+import styled from '@emotion/styled';
+import Link from 'next/link';
+import React from 'react';
+
+const StyledLink = styled.a<{ isActive: boolean }>(({ theme, isActive }) => ({
+  ...theme.typography.largeTitle,
+  color: isActive ? theme.colors.text : theme.colors['grey-04'],
+  paddingRight: `${theme.space * 4}px`,
```
baiirun, line 8: See the comment in [space-navbar](apps/web/modules/components/space/space-navbar.tsx)

apps/web/modules/services/network.ts
```diff
@@ -290,6 +301,98 @@ export class Network implements INetwork {
...
   };
+
+  fetchEntityTableData = async ({ spaceId, params, config }: FetchEntityTableDataParams) => {
+    /* TODO: Explore moving this method into another layer of the codebase responsible for both data querying and transformation  */
+
+    const storage = new StorageClient(config.ipfs);
+    const subgraph = config.subgraph;
+
+    /* To get our columns, fetch the all attributes from that type (e.g. Person -> Attributes -> Age) */
+    const columnsTriples = await new Network(storage, subgraph).fetchTriples({
```
baiirun, line 312: Now that I'm seeing this here I don't like it. We're having to instantiate a new network inside of the Network in order to avoid cancellation. We can leave it for now to get this merged, but let's figure out how we want to solve this before we release this feature to the wild.

apps/web/modules/services/network.ts
```diff
@@ -290,6 +301,98 @@ export class Network implements INetwork {
...
+      skip: 0,
+      filter: [
+        { field: 'entity-id', value: params.typeId },
+        { field: 'attribute-id', value: SYSTEM_IDS.ATTRIBUTES },
+      ],
+    });
+
+    /* To get our rows, first we get all of the entity IDs of the selected type */
+    const rowEntityIds = (
+      await new Network(storage, subgraph).fetchTriples({
```
baiirun, line 325: Can we put this network call and the one on line 312 together in a `Promise.all` so they happen in parallel?

packages/subgraph/src/bootstrap.ts
```diff
@@ -7,13 +7,19 @@ import {
...
+
 export function bootstrap(space: string, createdAtBlock: BigInt): void {
   log.debug(`Bootstrapping space ${space}!`, [])
 
   handleAction(new CreateEntityAction('type'), space, createdAtBlock)
   handleAction(new CreateEntityAction('name'), space, createdAtBlock)
   handleAction(new CreateEntityAction('attribute'), space, createdAtBlock)
   handleAction(new CreateEntityAction('space'), space, createdAtBlock)
+  handleAction(new CreateEntityAction(ATTRIBUTES_ID), space, createdAtBlock)
+  handleAction(new CreateEntityAction(SCHEMA_TYPE_ID), space, createdAtBlock)
```
baiirun, line 22: We should redeploy the subgraph after we merge so these get created with the correct ID in every space.

apps/web/modules/components/entity-table/entity-table-container.tsx
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export function EntityTableContainer({ spaceId, initialColumns, initialRows }: Props) {
+  const entityTableStore = useEntityTable();
+
+  return (
+    <PageContainer>
+      <Spacer height={20} />
+
+      <motion.div style={{ maxWidth: '100%' }} layout="position" transition={{ duration: 0.1 }}>
```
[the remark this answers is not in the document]
makeitrein: 👍 Resolved

apps/web/modules/components/entity-table/entity-table.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+    // We need to keep and update the state of the cell normally
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const [cellData, setCellData] = useState<Cell>(initialCellData);
+
+    // If the initialValue is changed external, sync it up with our state
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useEffect(() => {
+      setCellData(initialCellData);
+    }, [initialCellData]);
```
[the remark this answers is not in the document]
makeitrein: Is it worth keeping if table editing will next on our docket? Definitely want to pick your brain about how you structured the table editing for the triples table when that feature existed

apps/web/modules/components/triple-table.tsx
```diff
@@ -168,9 +125,9 @@ export const TripleTable = memo(function TripleTable({ triples, space }: Props)
           {table.getHeaderGroups().map(headerGroup => (
             <tr key={headerGroup.id}>
               {headerGroup.headers.map(header => (
-                <TableHeader width={header.column.getSize()} key={header.id}>
+                <TableHead width={header.column.getSize()} key={header.id}>
                   {flexRender(header.column.columnDef.header, header.getContext())}
-                </TableHeader>
+                </TableHead>
```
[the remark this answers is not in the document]
makeitrein: Excellent catch

apps/web/modules/components/entity-table/entity-table.tsx
makeitrein: yup, that feels more descriptive of what's actually going on - resolved

apps/web/modules/components/space/space-navbar.tsx
makeitrein: Yup agreed - outside component should be responsible for managing spacing

apps/web/modules/design-system/tab-link.tsx
makeitrein: ^^ Done

apps/web/modules/services/network.ts
makeitrein: Agreed, this can be the area of first refactor

apps/web/modules/services/network.ts
makeitrein: Nice optimization 👍

packages/subgraph/src/bootstrap.ts
makeitrein: Agreed, also need to bootstrap all of those other custom types like VALUE_TYPE, RELATION_TYPES, and MULTIPLE_ALLOWED

apps/web/modules/components/entity-table/entity-table.tsx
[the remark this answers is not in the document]
baiirun: I'd probably remove it for now just to simplify the code for now. It's easy enough to add it back whenever we get to editing.

apps/web/pages/space/[id]/entities.tsx
```diff
@@ -0,0 +1,142 @@
...
+    attributeId: '',
+    attributeName: '',
+    value: {
+      type: 'string',
+      value: '',
+      id: '',
+    },
+  };
+
+  const initialSelectedType =
```
Comment (baiirun) [inline apps/web/pages/space/[id]/entities.tsx:95]: I'd opt to set `initialSelectedType` to `null` if there's no initial type selected. We can figure out how we want to handle this on the client. Maybe we just select the first type in the list instead.

This works basically the same as the blessed `notFoundType` Triple, but without having to encode what the `notFoundType` means on the client and server.

apps/web/pages/space/[id]/entities.tsx
```diff
@@ -0,0 +1,142 @@
...
+      <SpaceHeader spaceId={spaceId} spaceImage={spaceImage} spaceName={spaceName} />
+
+      <Spacer height={34} />
+      <SpaceNavbar spaceId={spaceId} />
+
+      <EntityTableStoreProvider
+        space={spaceId}
+        config={config}
+        initialRows={initialRows}
+        initialSelectedType={initialSelectedType}
```
Comment (baiirun) [inline apps/web/pages/space/[id]/entities.tsx:51]: For example, here we can default to the first type in the list of there's no initial type selected. Maybe it makes sense to that on the server as well instead of on the client 🤔

baiirun (changes requested):

apps/web/modules/components/entity-table/entity-table.tsx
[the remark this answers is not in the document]
makeitrein: Done!

apps/web/pages/space/[id]/entities.tsx
Comment (makeitrein) [reply apps/web/pages/space/[id]/entities.tsx:51]: Made initialSelectedType nullable - feels like a decent paradigm, what say you good ser?

baiirun (approved):

PR #140, opened 2022-12-21
Batch editing

baiirun:
This PR ended up getting fairly large, mostly because of `modules/` changes that were necessarily to make the code for global actions tracking make more sense. You can check the high level changes below.

# Changeset
### Global Actions
User edits/actions are now tracked globally instead of just within an entity page. We now have a separate `ActionsStore` to track local changes. Stores that need local actions can now consume the `ActionsStore`. See `EntityStore` and `TriplesStore`. This greatly decouples and simplifies stores that consume user Actions.

The FlowBar is also now global. Change counts that appear in the FlowBar are now scoped to the space you're currently viewing. Changes being published from the FlowBar are also scoped to the space.

### Module organization
Various changes to the way modules are organized and namespaced. A lot of our file organization was left over from the prototyping days and didn't make sense with the current state of the repo. Hopefully the new organization with namespacing makes sense.

### Bug fixes
There were some bugs when aggregating the change count in the FlowBar. We now keep IDs of all locally changed triples stable to make tracking and diffing easier. This ID gets updated on the subgraph when changes are published.

We now squash all local actions before publishing to just the relevant changes. This reduces the work that IPFS and the subgraph has to do when there are large action sets.


apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -298,20 +295,19 @@ function EntityAttributes({
...
-        attributeName: groupedTriples[attributeId][0].attributeName,
-        value: {
-          type: 'entity',
-          id: linkedEntity.id,
-          name: linkedEntity.name,
-        },
-      })
-    );
+    create({
+      ...groupedTriples[attributeId][0],
```
baiirun, line 299: We want to keep IDs stable when being edited locally. See the PR description and `/modules/id/README` for more info.

apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -428,7 +424,10 @@ function EntityAttributes({
                     },
                   ]}
                 />
-                <SquareButton icon="trash" onClick={() => remove(triples.filter(t => t.attributeId === attributeId))} />
+                <SquareButton
+                  icon="trash"
+                  onClick={() => triples.filter(t => t.attributeId === attributeId).forEach(t => remove(t))}
+                />
```
baiirun, line 430: We unified the CRUD APIs for changing triples to all accept a single triple instead of an array of triples.

apps/web/modules/entity/entity-store-provider.tsx
```diff
@@ -13,16 +14,17 @@ interface Props {
 }
 
 export function EntityStoreProvider({ id, spaceId, children, initialTriples }: Props) {
-  const { network } = useServices();
+  const { network } = Services.useServices();
+  const ActionsStore = useActionsStoreContext();
```
baiirun, line 18: Some namespaced usage like this is inconsistent, but I'll fix in an upcoming PR

PR #141, opened 2022-12-22
Enable development on local chain

makeitrein:
Just uncommenting some prior code and updating the types. Do we want to the same for staging?

### DIFF
--- a/apps/web/modules/wallet.tsx
+++ b/apps/web/modules/wallet.tsx
@@ -9,21 +9,19 @@ import { Unlink } from './design-system/icons/unlink';
 import { Spacer } from './design-system/spacer';
 import { Text } from './design-system/text';
 
-// const LOCAL_CHAIN: Chain = {
-//   id: Number(configOptions.development.chainId),
-//   name: 'Geo Genesis Dev', // Human-readable name
-//   network: 'ethereum', // Internal network name
-//   nativeCurrency: {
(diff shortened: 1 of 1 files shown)


apps/web/modules/wallet.tsx
```diff
@@ -67,7 +65,10 @@ const DEFAULT_CHAIN: Chain = {
   },
 };
 
-const { chains, provider, webSocketProvider } = configureChains([DEFAULT_CHAIN, TESTNET_CHAIN], [publicProvider()]);
+const { chains, provider, webSocketProvider } = configureChains(
+  [LOCAL_CHAIN, DEFAULT_CHAIN, TESTNET_CHAIN],
```
baiirun, line 69: The DEFAULT_CHAIN should be first in the list, or the Connect button will attempt to connect to LOCAL_CHAIN when connecting your wallet.

baiirun, line 69: The DEFAULT_CHAIN should be first in the list, or the Connect button will attempt to connect to LOCAL_CHAIN when connecting your wallet.

baiirun (changes requested):

apps/web/modules/wallet.tsx
makeitrein: Good catch, resolved

baiirun (approved):

PR #145, opened 2023-01-03
Resolve filtering for 'asdfasdf' bug showing initial state for entity and triples table

makeitrein:
When filtering for 'asdfasdf' or other gibberish, both entity table and triples table would display initial rows fetched during SSR.

Rather than using filter state to determine whether or not to show SSR rows, I instead used a hydrated$ flag to mark when client side should be responsible for displaying data. 

This feels closer to the intent of our table filtering logic and less prone to bugs, especially as we begin to add table editing functionality.

Also used this PR as an opportunity to pass in the new abortController to fetchEntityTableData to prevent race conditions.


baiirun (approved):
🚀

PR #147, opened 2023-01-03
Exploration: Functional option/result types and higher-order-abstractions with ts-belt

baiirun:
This exploration is to hopefully solve for a couple things:
1. Using higher-order-abstractions for certain pieces of code in order to avoid easy-to-write bugs. This should hopefully make the code easier to understand as well by avoiding more "low level" implementation of things.
2. Better handling of branching code using `Option` types. i.e., handling different states that a value can be in as a first-class language construct.
3. Same thing for values/errors using `Result` types.
4. Making code more declarative and follow-able by using functions composition where it makes sense.
5. Avoid maintainance for some commonly used utilities like `groupBy` or `partition`

I'm not a functional purist by any means, and mostly just want to use its strengths where it's beneficial. I made some changes in the codebase just to provide some basic usage examples. There's still lots of places that could benefit from `ts-belt`, but we can address those as we run into them.

### DIFF
--- a/apps/web/modules/components/entity-table/entity-table.tsx
+++ b/apps/web/modules/components/entity-table/entity-table.tsx
@@ -169,7 +169,7 @@ export const EntityTable = memo(function EntityTable({ rows, space, columns }: P
                 return (
                   <TableCell
                     isExpandable={isExpandable}
-                    isExpanded={expandedCells[cellId]}
+                    isExpanded={expandedCells[cellId] ?? false}
                     width={cell.column.getSize()}
                     key={cell.id}
                     toggleExpanded={() =>
--- a/apps/web/modules/components/entity-table/entity-table.tsx
+++ b/apps/web/modules/components/entity-table/entity-table.tsx
@@ -169,7 +169,7 @@ export const EntityTable = memo(function EntityTable({ rows, space, columns }: P
                 return (
                   <TableCell
                     isExpandable={isExpandable}
-                    isExpanded={expandedCells[cellId] ?? false}
+                    isExpanded={expandedCells[cellId]}
                     width={cell.column.getSize()}
                     key={cell.id}
                     toggleExpanded={() =>
--- a/apps/web/modules/components/entity-table/entity-table.tsx
+++ b/apps/web/modules/components/entity-table/entity-table.tsx
@@ -1,4 +1,5 @@
 import styled from '@emotion/styled';
+import { A } from '@mobily/ts-belt';
 import {
   ColumnDef,
   createColumnHelper,
@@ -153,7 +154,7 @@ export const EntityTable = memo(function EntityTable({ rows, space, columns }: P
           ))}
         </thead>
         <tbody>
-          {table.getRowModel().rows.length === 0 && (
+          {A.isEmpty(table.getRowModel().rows) && (
             <tr>
               <EmptyTableText>No results found</EmptyTableText>
             </tr>
--- a/apps/web/modules/components/entity/editable-entity-page.tsx
+++ b/apps/web/modules/components/entity/editable-entity-page.tsx
@@ -16,7 +16,7 @@ import { EntityTextAutocomplete } from './entity-text-autocomplete';
 import { Entity, useEntityStore } from '~/modules/entity';
 import { useActionsStore } from '~/modules/action';
 import { useEditEvents } from './edit-events';
-import { A, B, pipe } from '@mobily/ts-belt';
+import { A, B, D, F, pipe } from '@mobily/ts-belt';
 
 const PageContainer = styled.div({
   display: 'flex',
@@ -218,7 +218,7 @@ function EntityAttributes({
       type: 'REMOVE_ENTITY',
       payload: {
         triple,
-        isLastEntity: groupedTriples[triple.attributeId]?.length === 1,
+        isLastEntity: groupedTriples[triple.attributeId].length === 1,
       },
     });
   };
--- a/apps/web/modules/components/entity/editable-entity-page.tsx
+++ b/apps/web/modules/components/entity/editable-entity-page.tsx
@@ -16,7 +16,7 @@ import { EntityTextAutocomplete } from './entity-text-autocomplete';
 import { Entity, useEntityStore } from '~/modules/entity';
 import { useActionsStore } from '~/modules/action';
 import { useEditEvents } from './edit-events';
-import { A, B, D, F, pipe } from '@mobily/ts-belt';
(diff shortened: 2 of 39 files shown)


apps/web/tsconfig.json
```diff
@@ -14,6 +14,7 @@
     "isolatedModules": true,
     "jsx": "preserve",
     "incremental": true,
+    "noUncheckedIndexedAccess": true,
```
baiirun, line 17: This is required for `Option` to work in `ts-belt`. IMO this results in _super_ defensive code and often the TS compiler won't narrow correctly for the existence of an object in a Record, resulting in more complexity to correctly handle the undefined cases.

apps/web/pages/space/[id]/[entityId].tsx
```diff
@@ -86,9 +87,15 @@ export const getServerSideProps: GetServerSideProps<Props> = async context => {
     .flatMap(entity => entity.triples)
     .reduce((acc, triple) => {
       if (!acc[triple.entityId]) acc[triple.entityId] = { triples: [], name: null, id: triple.entityId };
-      acc[triple.entityId].id = triple.entityId;
-      acc[triple.entityId].name = triple.entityName;
-      acc[triple.entityId].triples = [...acc[triple.entityId].triples, triple]; // Duplicates?
```
Comment (baiirun) [inline apps/web/pages/space/[id]/[entityId].tsx:91]: The check on line 88 doesn't narrow as expected here, so we have to rewrite to handle the undefined values. Using Option and mapping to the new value may make it more ergonomic, but I didn't want to put too many changes in this PR.

apps/web/modules/components/entity-table/entity-table.tsx
```diff
@@ -153,7 +154,7 @@ export const EntityTable = memo(function EntityTable({ rows, space, columns }: P
           ))}
         </thead>
         <tbody>
-          {table.getRowModel().rows.length === 0 && (
+          {A.isEmpty(table.getRowModel().rows) && (
```
baiirun, line 157: We can use a higher-order abstraction instead of manually writing length checks.

apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -80,7 +81,13 @@ export function EditableEntityPage({ id, name: serverName, space, triples: serve
...
   // We hydrate the local editable store with the triples from the server. While it's hydrating
   // we can fallback to the server triples so we render real data and there's no layout shift.
-  const triples = localTriples.length === 0 && actions.length === 0 ? serverTriples : localTriples;
+  const triples = pipe(
+    A.isEmpty(localTriples) && A.isEmpty(actions),
+    B.ifElse(
+      () => serverTriples,
+      () => localTriples
+    )
+  );
```
baiirun, line 90: We can do some functional composition to get ternaries. Not sure if this is really better but it's an example.

apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -195,9 +202,9 @@ function EntityAttributes({
   send: ReturnType<typeof useEditEvents>;
   name: string;
 }) {
-  const groupedTriples = groupBy(triples, t => t.attributeId);
+  const groupedTriples = A.groupBy(triples, t => t.attributeId);
```
baiirun, line 205: No longer need to maintain our own groupBy function.

apps/web/modules/components/entity/readable-entity-page.tsx
```diff
@@ -260,7 +261,7 @@ function LinkedEntityCard({
 }) {
   const [isExpanded, setIsExpanded] = useState(false);
 
-  const [linkedTriples, unlinkedTriples] = partition(
+  const [linkedTriples, unlinkedTriples] = A.partition(
```
baiirun, line 264: No longer need to maintain our own partition function.

apps/web/modules/components/flow-bar.tsx
```diff
@@ -102,12 +102,16 @@ interface ReviewProps {
...
+    Action.squashChanges,
+    A.groupBy(action => {
       if (action.type === 'deleteTriple' || action.type === 'createTriple') return action.entityId;
       return action.after.entityId;
-    })
-  ).length;
+    }),
+    D.keys,
+    A.length
+  );
```
baiirun, line 114: This feels like a more readable way of composing `Array` and `Object` built-ins instead of what we had before.

apps/web/modules/entity/autocomplete.tsx
```diff
@@ -1,25 +1,40 @@
...
+
+          // Right now we don't do anything with the error state.
+          R.catchError(() => R.Error('Failed to fetch autocomplete entities'))
+        );
       })
     );
+
+    // If the result of the computed fetch is an error we default to an empty array.
+    // Otherwise return the data.
+    this.results$ = computed(() => R.getWithDefault(networkData$.get(), []));
```
baiirun, line 37: We can represent the result of functions that can either succeed or error by using the `Result` type. This kinda looks like `.then/.catch`, but with first-class support for the resulting value either being the successful value or the error.

apps/web/modules/params/params.ts
```diff
@@ -1,20 +1,23 @@
...
-  }, []);
+  const filterStateResult = pipe(
+    [...params.keys()],
+    A.filter(key => key !== 'query' && key !== 'page'),
+    A.reduce([] as FilterState, (acc, key) => {
+      const value = params.get(key);
+      if (!value) return acc;
+      return [...acc, { field: key as FilterField, value }];
+    })
+  );
```
baiirun, line 20: Another example of function composition using `pipe` instead of dot notation. `pipe` is a bit more composable since we can put our own functions inside the `pipe`. With dot notation we're stuck with whatever `.functionCall` exists in the builder.

apps/web/package.json
```diff
@@ -20,6 +20,7 @@
     "@geogenesis/contracts": "workspace:*",
     "@legendapp/state": "^0.20.0",
     "@mailchimp/mailchimp_marketing": "^3.0.80",
+    "@mobily/ts-belt": "4.0.0-rc.2",
```
baiirun, line 23: Using the next major version's RC for a couple reasons
1. It has more ergonomic types for things like `groupBy`
2. It doesn't rely on `noUncheckedIndexedAccess` flag in `tsconfig` in order to use `Option`. This flag is super un-ergonomic if we use a lot of Records, which we do.
3. Has support for Async data types to represent async operations. e.g., `AsyncResult` or `AsyncOption.`

baiirun:
Stale and no longer relevant

PR #154, opened 2023-01-04
Fix: Hydration mismatches due to behavior mismatch between mobile wallets and extension wallets

baiirun:
To fix the hydration issues with our wallet and fix some of the MM issues we've been having when switching chains I'm migrating from Rainbowkit to Connectkit for our wallet integrations. Rainbowkit seems to be falling behind based on their github issues and activity as their main developer has moved on to work at Remix/Shopify. We've felt some this with not being able to update wagmi versions to solve some MM issues we've been having while continuing to use Rainbowkit.

Connectkit has a similar quality level of UX and they seem to be moving a lot faster and staying up-to-date with dependencies compared to Rainbow. They are in a similar place to the Rainbow team in that they are developing their own wallet.

### DIFF
--- a/apps/web/modules/wallet.tsx
+++ b/apps/web/modules/wallet.tsx
@@ -1,7 +1,7 @@
 import styled from '@emotion/styled';
-import { ConnectButton, getDefaultWallets, RainbowKitProvider } from '@rainbow-me/rainbowkit';
 import '@rainbow-me/rainbowkit/styles.css';
 import { Chain, configureChains, createClient, WagmiConfig } from 'wagmi';
+import { polygon, polygonMumbai } from 'wagmi/chains';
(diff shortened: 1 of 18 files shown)


apps/web/modules/wallet.tsx
```diff
@@ -103,19 +89,25 @@ const StyledConnectButton = styled.button(props => ({
 }));
 
 export function GeoConnectButton() {
+  // There's currently no mechanisms in connectkit to handle disconnecting their APIs
+  // without going through the modal. It uses wagmi internally so we can escape-hatch
+  // into wagmi-land to disconnect.
+  const { disconnect } = useDisconnect();
```
baiirun, line 95: Instead of having to open the modal to disconnect we're just disconnecting directly whenever the user clicks "Disconnect wallet."

PR #158, opened 2023-01-05
Editable table cells

makeitrein:
@baiirun One thing that is a bit funky is name rendering: 

Not sure how to handle the situation when there is both an entityName and a name triple.

Screen Shot 2023-01-05 at 11 40 22 AM
Screen Shot 2023-01-05 at 11 40 27 AM

You can view this bug on testnet on this URL: http://localhost:3000/space/0xb5E2cD8A5F88517d3576ba99d52C005b19351A43/entities?typeId=fec429e7-0f17-4dd4-a39e-2e60f91fcd5d

### DIFF
--- a/apps/web/modules/components/entity-table/editable-entity-table-cell.tsx
+++ b/apps/web/modules/components/entity-table/editable-entity-table-cell.tsx
@@ -1,0 +1,146 @@
+import styled from '@emotion/styled';
+import { useActionsStore } from '~/modules/action';
+import { useEntityStore } from '~/modules/entity';
+import { groupBy } from '~/modules/utils';
+import { ChipButton } from '../../design-system/chip';
+import { Cell, Triple } from '../../types';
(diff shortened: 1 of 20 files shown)


baiirun (changes requested):
Looks good! I fixed the text-wrapping issue on the editable StringFields. You had wrapped the StringField in `editable-entity-page` and `editable-entity-table-cell` with a `div` which broke the width.

There's a couple console logs left in the code if you can remove those.

makeitrein:
> @makeitrein For the name field maybe we just default to the name triple since that's the actual source of truth. Whenever the name changes locally we probably need to update all local triples for that entity to reflect the new entityName. I realized this while working on the persistent local actions.

Sounds good, updated to use Entity.name in both places. 

Agreed on your use case, that sounds important for this local state push.

baiirun (approved):
🔥

PR #160, opened 2023-01-06
feat: type attributes autofilling when creating an entity

MercuricChloride:


apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -307,6 +310,17 @@ function EntityAttributes({
...
 
+  //@ts-ignore
+  // type inference being stupid
+  const entityTypes = groupedTriples.type.filter(t => t.value.type === 'entity').map(t => t.value.name);
+
+  const requiredAttributes = attributes
+    .filter(t => entityTypes.includes(t.entityName) && t.value.type === 'entity')
+    .filter((value, index, self) => self.indexOf(value) === index);
+
+  console.log('Entity Types: ', entityTypes);
```
yanivtal, line 321: Let's not check in console.logs

apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -307,6 +310,17 @@ function EntityAttributes({
     }
   };
 
+  //@ts-ignore
+  // type inference being stupid
```
yanivtal, line 314: Let's not check in messages like this

apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -307,6 +310,17 @@ function EntityAttributes({
...
+  const entityTypes = groupedTriples.type.filter(t => t.value.type === 'entity').map(t => t.value.name);
+
+  const requiredAttributes = attributes
```
yanivtal, line 317: The attributes aren't required. Maybe typeAttributes

yanivtal (commented):
Just some code cleanliness things

apps/web/modules/components/entity/editable-entity-page.tsx
baiirun: In TS filter annoying doesn't type narrow. You can add a flatMap after the filter to narrow the type in the filtered arrow like this:

```ts
const entityTypes = groupedTriples.type
    .filter(t => t.value.type === 'entity')
    .flatMap(t => t.value.type === 'entity' ? t.value.name : [])
```

apps/web/pages/space/[id]/create-entity.tsx
```diff
@@ -3,28 +3,56 @@ import { useMemo } from 'react';
...
+      filter: [],
+    }),
+    network.fetchTriples({
+      space,
+      query: '',
+      skip: 0,
+      first: 100,
+      filter: [{ field: 'attribute-name', value: 'Attributes' }],
+    }),
+  ]);
```
Comment (baiirun) [inline apps/web/pages/space/[id]/create-entity.tsx:49]: If I'm understanding this correctly:
1. We fetch the first 100 triples whose attribute name is Attributes.
2. We pass those triples to `EditableEntityPage`
3. When we select Type in the edit UI, we check if the name of that Type is the entity name of any of these attributes.
4. For every one of these matches, we render a placeholder for the attribute triple.

What if there's more than 100 Attribute Triples? We can fetch more but there's an upper limit where it stops making sense to do it all ahead of time. Maybe it makes more sense to query for the attributes at selection-time as the user selects a type. This does add a lot more complexity.
1. User selects Type
2. We query for the attributes that match that type
3. We render the placeholders

This is partly why I suggested moving the Types triple field to it's own area in the UI (and the code) so it's easier to do these more complex interactions that will require network requests at runtime without mixing in with the other editable triples whose behavior is different.

baiirun:
@makeitrein @MercuricChloride is this PR good to close in favor of [Entity autofill v2](https://github.com/geobrowser/geogenesis/pull/165)?

baiirun:
[An updated version of this PR was merged here](https://github.com/geobrowser/geogenesis/pull/165)

PR #161, opened 2023-01-06
Logrocket tracking setup

makeitrein:
Only tracks users in production. We plan to turn this off post-beta.

Invites sent to Yaniv and Byron, we can pay for more seats and session tracking if we like the product.


baiirun (approved):
🔥

PR #163, opened 2023-01-06
feat: disambiguate entity search

baiirun:
This PR disambiguates entities in the entity search list by listing either their `Types` or their `Description string value` in a priority order where Types takes precedence over Description. It also formalizes some opinions on how we should handle descriptions from the knowledge graph in the UI.

### Description handling
Previously we were handling several cases where a Description could exist on an entity:
1. The description triple's `attributeId === SYSTEM_IDS.DESCRIPTION_SCALAR` and the `valueType: STRING`
2. The description triple's `attributeId === SYSTEM_IDS.DESCRIPTION_SCALAR` and the `valueType: ENTITY`.
3. The description triple's `attributeId !== SYSTEM_IDS.DESCRIPTION_SCALAR` and the `valueType: STRING`
3. The description triple's `attributeId !== SYSTEM_IDS.DESCRIPTION_SCALAR` and the `valueType: ENTITY`

Instead of handling every combination of attributeId:valueType, we are now only handling these cases:
- `attributeId: SYSTEM_IDS.DESCRIPTION_SCALAR` and `valueType: STRING`
- `attributeName: Description` and `valueType: STRING`

**This formalizes some assumptions where our UI assumes that the value of the Description triple should be the `STRING` type.** For now we'll handle edge-cases where someone assigned a different Description attribute than the system one, but eventually we may want to enforce that users select the correct Description attribute when adding a description to an Entity.


apps/web/integration/pages/space/[id]/[entityId].test.tsx
```diff
@@ -59,16 +59,6 @@ describe('Entity page', () => {
...
-  it('Renders entity description from description that is an entity type', () => {
-    render(
-      <Providers>
-        <EntityPage id="1" name="Banana" space="1" triples={[linkedDescriptionTriple]} linkedEntities={{}} />
-      </Providers>
-    );
-
-    expect(screen.queryAllByText('Description of a Banana').length).toEqual(2);
-  });
-
```
Comment (baiirun) [inline apps/web/integration/pages/space/[id]/[entityId].test.tsx:71]: We removed some tests where we were testing that the entity page would render description triples whose value was `valueType: ENTITY`

apps/web/modules/components/entity/autocomplete/entity-autocomplete.tsx
```diff
@@ -48,43 +48,6 @@ const SearchIconContainer = styled.div(props => ({
...
-    outline: 'none',
-    backgroundColor: props.theme.colors['grey-01'],
-  },
-
-  ...(props.existsOnEntity && {
-    backgroundColor: props.theme.colors['grey-01'],
-    cursor: 'not-allowed',
-  }),
-}));
-
```
baiirun, line 87: I abstracted some shared UI components from the two autocomplete dialogs into the same definitions

apps/web/modules/services/network-local-mapping.ts
```diff
@@ -0,0 +1,59 @@
+import { Action, Entity, Space, Triple, Value } from '../types';
```
baiirun, line 1: The network file was starting to get overload with types and functions for mapping from Network versions of types to App-specific versions of types.

apps/web/modules/value/value.ts
```diff
@@ -0,0 +1,5 @@
+import { Triple } from '../types';
+
+export function entityName(triple: Triple) {
+  return triple?.value?.type === 'entity' ? triple.value.name : null;
+}
```
baiirun, line 5: I moved the entityName helper to a `Value` module since it was kinda confusing being in the `Entity` module. The intended usage of this helper is to parse the name of an `EntityValue` @makeitrein @badgeth-dao

apps/web/modules/services/network.ts
```diff
@@ -371,7 +378,7 @@ export class Network implements INetwork {
 
     /* ...and then we can format our user-defined schemaColumns */
     const schemaColumns = columnsTriples.triples.map(triple => ({
-      name: Entity.entityName(triple) || triple.value.id,
+      name: Value.entityName(triple) || triple.value.id,
```
baiirun, line 381: See [this comment](https://github.com/geobrowser/geogenesis/pull/163/files#r1063852871)

apps/web/modules/value/value.ts
```diff
@@ -0,0 +1,5 @@
+import { Triple } from '../types';
+
+export function nameOfEntityValue(triple: Triple) {
+  return triple?.value?.type === 'entity' ? triple.value.name : null;
```
baiirun, line 4: I moved the entityName helper to a Value module since it was kinda confusing being in the Entity module. The intended usage of this helper is to parse the name of an EntityValue @makeitrein @badgeth-dao

apps/web/modules/services/network.ts
```diff
@@ -371,7 +378,7 @@ export class Network implements INetwork {
 
     /* ...and then we can format our user-defined schemaColumns */
     const schemaColumns = columnsTriples.triples.map(triple => ({
-      name: Entity.entityName(triple) || triple.value.id,
+      name: Value.nameOfEntityValue(triple) || triple.value.id,
```
baiirun, line 381: See [this comment](https://github.com/geobrowser/geogenesis/pull/163/files#r1063853419)

apps/web/modules/components/triple-table.tsx
```diff
@@ -70,8 +70,9 @@ const defaultColumn: Partial<ColumnDef<Triple>> = {
         );
       }
       case 'attributeId': {
-        const attributeId = cellData as string;
-        const value = triple.attributeName ?? attributeId;
+        const attributeName = cellData as string;
+        const value = attributeName ?? triple.attributeId;
```
baiirun, line 74: This surfaced in a bug where the Description attribute name would not render in the triples table since it does not have a `name` triple atm.

We changed the accessor to `triple.attributeName` so in the previous version `triple.attributeName` and `attributeId` were the same thing (see line 74). Now we correctly default to ID if name doesn't exist.

makeitrein (approved):
LGTM

PR #165, opened 2023-01-07
Entity autofill v2

MercuricChloride:
Gonna wrap this up on with entity schema defaults selection on Monday.


apps/web/modules/entity/entity-store.tsx
```diff
@@ -64,6 +68,55 @@ export class EntityStore implements IEntityStore {
...
+                },
+              ],
+            });
+          })
+        );
+
+        return attributes
+          .map(attribute => attribute.triples)
+          .flat()
+          .map(triple => ({
```
baiirun, line 112: We can use `flatMap` here if we really want to save a loop. Probably doesn't matter in most cases since there's not many schemaTypes for a given type.

apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -206,18 +210,45 @@ const GroupedAttributesList = styled.div(({ theme }) => ({
...
   send,
 }: {
   entityId: string;
   triples: Props['triples'];
+  schemaTriples: Props['schemaTriples'];
   send: ReturnType<typeof useEditEvents>;
   name: string;
 }) {
-  const groupedTriples = groupBy(triples, t => t.attributeId);
+  const [deletedPlaceholders, setDeletedPlaceholders] = useState<string[]>([]);
```
baiirun, line 223: Does it make more sense to keep deletedPlaceholders in the EntityStore? Right now most of our state lives in the EntityStore so it's a little confusing for some of it to live in React state and some to live in EntityState, especially since we're already tracking placeholder-related state in the store.

apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -206,18 +210,45 @@ const GroupedAttributesList = styled.div(({ theme }) => ({
...
+
+  const unusedSchemaTriples = schemaTriples.filter(t => !triples.some(t2 => t2.attributeId === t.attributeId));
```
baiirun, line 225: Can we use more meaningful names for the triples in these callbacks? Like `schemaTriple` and `triple` or something.

apps/web/modules/components/entity/edit-events.ts
```diff
@@ -96,6 +103,8 @@ interface ListenerConfig {
 const listener =
   ({ api: { create, update, remove }, context }: ListenerConfig) =>
   (event: EditEvent) => {
+    console.log(event);
```
baiirun, line 106: Can remove this lil guy

baiirun (changes requested):
Looks good. We should also add some tests for all of the placeholder-related code.
- Tests in the`EntityStore` to ensure `typeTriples` and `schemaTriples` populated as expected. You can use the `MockNetwork` to test `schemaTriples` without using the real subgraph.
- Tests to make sure we're rendering the correct UI for the placeholder vs. non-placeholder fields. And that manipulating one or the other correctly updates the UI.

MercuricChloride:
Just an fyi, still working out these bugs, these changes aren't the final submission for this PR

apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -213,22 +228,20 @@ function EntityAttributes({
...
+  deleteSchemaId,
+  deletedSchemaIds,
 }: {
   entityId: string;
   triples: Props['triples'];
   schemaTriples: Props['schemaTriples'];
   send: ReturnType<typeof useEditEvents>;
   name: string;
+  deleteSchemaId: (id: string) => void;
+  deletedSchemaIds: (String | undefined)[];
```
baiirun: Should this be lowercase `string | undefined`? We can also filter out the `undefined` values if that makes more sense.

apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -324,14 +337,16 @@ function EntityAttributes({
...
+            onBlur={e => {
+              updateValueFromPlaceholder({ ...triple }, e.target.value);
+            }}
           />
         ) : (
           <StringField
             key={triple.id}
             variant="body"
-            placeholder="Add value..."
+            placeholder="String value!"
```
baiirun: We should change this back to "Add value..."

apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -375,6 +390,9 @@ function EntityAttributes({
         const attributeName = triples[0].attributeName;
         const isPlaceholder = triples[0].placeholder;
 
+        console.log('TRIPLES PER GROUP', triples);
```
baiirun: Can get rid of this lil guy

apps/web/modules/entity/entity-store.tsx
```diff
@@ -106,9 +107,11 @@ export class EntityStore implements IEntityStore {
           })
         );
 
+        const deletedSchemaIds = this.deletedSchemaIds$.get();
+
         return attributes
-          .map(attribute => attribute.triples)
-          .flat()
+          .flatMap(attribute => attribute.triples)
+          .filter(triple => !deletedSchemaIds.includes(triple.attributeId))
```
baiirun: Is this related to the `@goose` comment in `editable-entity-page`? If so we should probably just do one or the other instead of both.

apps/web/modules/entity/entity-store.tsx
```diff
@@ -44,6 +44,7 @@ export class EntityStore implements IEntityStore {
   triples$: ObservableComputed<TripleType[]>;
   typeTriples$: ObservableComputed<TripleType[]> = observable([]);
   schemaTriples$: ObservableComputed<TripleType[]> = observable([]);
+  deletedSchemaIds$: Observable<String[]> = observable<String[]>([]);
```
baiirun: Should this be lowercase `string`?

baiirun (commented):
Looks good. Had a few questions

apps/web/modules/components/entity/editable-entity-page.tsx
makeitrein: Done!

apps/web/modules/components/entity/editable-entity-page.tsx
makeitrein: Removed all logs!

apps/web/modules/entity/entity-store.tsx
makeitrein: Yup yup

apps/web/modules/entity/entity-store.tsx
makeitrein: Which comment are you referring to?

apps/web/modules/components/entity/editable-entity-page.tsx
makeitrein: Done!

apps/web/modules/components/entity/editable-entity-page.test.tsx
```diff
@@ -0,0 +1,65 @@
+import { render, screen } from '@testing-library/react';
```
baiirun, line 1: 🔥

apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -203,19 +230,83 @@ const GroupedAttributesList = styled.div(({ theme }) => ({
...
+    return notHidden && notInTriples;
+  });
+
+  const filteredTriples = [...triples, ...visibleSchemaTriples];
+
+  const entityValueTriples = triples.filter(triple => triple.value.type === 'entity');
+
+  const schemaAttributeIds = schemaTriples.map(schemaTriple => schemaTriple.attributeId);
+
+  const sortedTriples = filteredTriples.sort((tripleA, tripleB) => {
```
baiirun, line 263: Can we move this to its own function? Would make this component more scannable without having to scroll past 40 lines of sorting code. Would make it possible to test this sorting algo as well.

apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -338,45 +479,50 @@ function EntityAttributes({
...
+                        ),
+                        onClick: () => onChangeTripleType('string', triples),
+                        disabled: !isEntityGroup,
+                      },
+                      {
+                        label: (
+                          <div style={{ display: 'flex', alignItems: 'center' }}>
+                            <Relation />
+                            <Spacer width={8} />
+                            <Text variant="button">Relation</Text>
```
baiirun, line 508: can you change the labels back to not have the `Text` component wrapping the field? It'll break styling of the type dropdown.

apps/web/modules/components/entity/editable-entity-page.tsx
```diff
@@ -338,45 +479,50 @@ function EntityAttributes({
...
-                      ),
-                      onClick: () => onChangeTripleType('string', triples),
-                      disabled: !isEntityGroup,
-                    },
-                    {
-                      label: (
-                        <div style={{ display: 'flex', alignItems: 'center' }}>
-                          <Relation />
-                          <Spacer width={8} />
-                          Relation
```
baiirun, line 366: Related to X: The field label should look like this

apps/web/modules/entity/entity-store.test.ts
```diff
@@ -0,0 +1,74 @@
+import { when } from '@legendapp/state';
```
baiirun, line 1: 🔥

apps/web/modules/entity/entity-store.tsx
```diff
@@ -61,11 +71,116 @@ export class EntityStore implements IEntityStore {
...
+
+    observe<string[]>(e => {
+      const typeIds = this.typeIds$.get();
+      const previous = e.previous || [];
+
+      if (!A.eq(previous, typeIds, (a, b) => a === b)) {
+        this.setSchemaTriples(typeIds);
+      }
+
+      return typeIds;
```
baiirun, line 96: Might be worth adding a comment as to why we're using `observe` here instead of `computed`

baiirun (commented):
Looking really good! Just a couple small comments

apps/web/modules/entity/entity-store.tsx
makeitrein: Good call, adding now...

apps/web/modules/components/entity/editable-entity-page.tsx
makeitrein: Agreed, good call

apps/web/modules/components/entity/editable-entity-page.tsx
makeitrein: Yup, mucked up during a merge

apps/web/modules/components/entity/editable-entity-page.tsx
makeitrein: Good catch

baiirun (approved):
🎉

PR #176, opened 2023-01-10
ids package

baiirun:
Moved system IDs from the web codebase to a separate `@geogenesis/ids` package so they can be shared between the subgraph and the web app.


makeitrein (approved):
LGTM

PR #179, opened 2023-01-10
feat: search locally created entities

baiirun:
As part of the editing flow users may create entities locally that they want to reference from other entities. Right now they need to publish these new entities before they are able to reference them.

This PR adds the ability to search locally created and edited entities with the search results from the network when searching for entities.

### DIFF
--- a/apps/web/modules/entity/entity-store.tsx
+++ b/apps/web/modules/entity/entity-store.tsx
@@ -62,7 +62,7 @@ export class EntityStore implements IEntityStore {
         return isCreate || isDelete || isRemove;
       });
       // We want to merge any local actions with the network triples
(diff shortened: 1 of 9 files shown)


apps/web/modules/entity/entity-store.tsx
```diff
@@ -62,7 +62,7 @@ export class EntityStore implements IEntityStore {
         return isCreate || isDelete || isRemove;
       });
       // We want to merge any local actions with the network triples
-      return Triple.fromActions(spaceId, entitySpecificActions, initialDefaultTriples);
+      return Triple.fromActions(entitySpecificActions, initialDefaultTriples);
```
baiirun, line 65: `spaceId` wasn't being used in `fromActions` so it's been removed.

PR #182, opened 2023-01-12
feat: entity table advanced filtering

baiirun:


baiirun:
Closing for now since the work for sorting and filtering table blocks is coming up and will probably deviate from this implementation

PR #189, opened 2023-01-17
Add entity dropdown

MercuricChloride:


baiirun (changes requested):
It looks like there's a bug where an empty relation gets added to the Types field after creating a new type. Here's the repro
1. Create a new type from the dropdown
2. Navigate to the new type by searching for the name in the triples table.
3. Change to edit mode to see the new triples.
4. There should be an empty types relation under the Types attribute.

There's also several places where we have imports that aren't being used (like in `type-dialog`) and some logs (like in `entity-table-store`. Looks like linting in CI isn't working correctly for some reason so I'll fix that this week so CI catches these.

apps/web/modules/components/entity-table/entity-input.tsx
```diff
@@ -72,32 +77,37 @@ export function EntityInput() {
...
-              <AdvancedFilterPill
-                key={filter.field}
-                filterClause={filter}
-                onClick={() => onAdvancedFilterClick(filter.field)}
-              />
-            ))}
-          </AdvancedFilters>
-        )}
-      </SearchInputContainer>
+      <EntityStoreProvider id={newId} spaceId={spaceId} initialTriples={[]} initialSchemaTriples={[]}>
```
baiirun, line 84: Since we're just creating two triples, we probably don't need to wrap this in an EntityStoreProvider. As far as I can tell you're really just trying to access the `create` API from the `ActionsStore`. We can probably just use `useActionStoreContext()` here instead.

It should clean up a lot of the code around the `EntityStore` that you added as well as the code in `entity-input`


```tsx
const ActionStore = useActionsStoreContext()

const onCreateType = () => {
  const newId = ID.createEntityId();
  ActionStore.create(nameTriple)
  ActionStore.create(typeTriple)
}
```

apps/web/modules/components/filter/type-dialog.tsx
```diff
@@ -116,23 +175,38 @@ export function TypeDialog({ inputContainerWidth }: Props) {
...
+                onChange={e => {
+                  setFilter(e.target.value);
+                }}
+              />
+              <AddEntityContainer>
+                <Text variant="button">All types</Text>
+                {searchMode ? (
+                  <SmallButton variant="secondary" icon="createSmall" onClick={handleCreateType}>
+                    New Type
+                  </SmallButton>
```
baiirun, line 191: You might've missed the convo, but we decided to not have the small `Create Type` button and instead just have the big blue one that appears when there's no types in the search results.

baiirun (commented):
Suggested another way of solving this that might affect less of the codebase.

baiirun (approved):
🔥

PR #199, opened 2023-01-19
Shift columnSchema into column triples, add basic column header text editing

makeitrein:


apps/web/modules/services/network.ts
```diff
@@ -426,18 +419,18 @@ export class Network implements INetwork {
     /* Name is the default column... */
     const defaultColumns = [
```
baiirun, line 420: Can we add this to the variable definition instead of casting?

```ts
const defaultColumns: Column[] = [ ... ]

```

apps/web/modules/services/network.ts
```diff
@@ -426,18 +419,18 @@ export class Network implements INetwork {
...
-    ];
+    ] as Column[];
 
     /* ...and then we can format our user-defined schemaColumns */
-    const schemaColumns = columnsTriples.triples.map(triple => ({
-      name: Value.nameOfEntityValue(triple) || triple.value.id,
+    const schemaColumns = columnsTriples.triples.map((triple, i) => ({
       id: triple.value.id,
+      triples: relatedColumnTriples[i].triples,
     })) as Column[];
```
baiirun, line 431: Same thing here

baiirun (approved):
Looks good! Just one small nitpick-y thing

PR #202, opened 2023-01-25
feat: add keyboard shortcut for edit mode

baiirun:


apps/web/pages/_app.tsx
```diff
@@ -75,38 +56,68 @@ function MyApp({ Component, pageProps }: AppProps) {
...
     </Relative>
   );
 }
 
+function App({ Component, pageProps }: AppProps) {
+  const router = useRouter();
+  const { id: spaceId } = router.query as { id: string | undefined };
+  const { setEditable, editable } = useEditable();
+  const { isEditor, isAdmin, isEditorController } = useAccessControl(spaceId);
+  const [open, setOpen] = useState(false);
```
baiirun, line 70: Moved all the below code to a new component so we could access `useAccessControl` which relies on the service provider to be initialized.

PR #207, opened 2023-02-01
Subgraph changes

MercuricChloride:
New subgraph schema


baiirun:
These changes ended up not making sense once we tried integrating with the client. Mapping from this data model back to triples added more complexity and we deemed it not necessary now that we have child filters.

PR #214, opened 2023-02-16
Onboarding profile

makeitrein:
Update `apps/web/modules/onboarding/dialog.tsx` on line 199 to be `open={true}` to see the onboarding profile in action!

Stubbed some functions and hooks since the backend is cooking right now.


apps/web/modules/onboarding/dialog.tsx
```diff
@@ -0,0 +1,212 @@
...
+    if (e.target.files) {
+      const file = e.target.files[0];
+      const ipfsUri = await network.uploadFile(file);
+      setAvatar(ipfsUri);
+    }
+  };
+
+  return (
+    <div>
+      <div className="text-center pb-4 text-smallTitle -mt-6">{name}</div>
```
baiirun, line 112: This should probably be some sort of header for accessibility reasons.

apps/web/modules/onboarding/dialog.tsx
```diff
@@ -0,0 +1,212 @@
...
+          <input
+            placeholder="Name..."
+            className="text-center block px-2 py-1"
+            value={name}
+            onChange={e => setName(e.target.value)}
+            autoFocus
+          />
+        </div>
+      </div>
+      <div className="text-bodySemibold text-xl text-center">
```
baiirun, line 74: This should be some sort of header. We also still have the `Text` component for text variants.

apps/web/modules/onboarding/dialog.tsx
```diff
@@ -0,0 +1,212 @@
...
+          />
+        ) : (
+          <Avatar size={154} value={address} />
+        )}
+      </div>
+
+      <div className="flex justify-center">
+        <label className="text-ctaPrimary text-metadataMedium text-center cursor-pointer hover:underline inline-block">
+          Upload photo
+          <input accept="image/png, image/jpeg" onChange={handleChange} type="file" className="hidden" />
```
baiirun, line 131: Shouldn't the input be outside of the label? They should be siblings.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

apps/web/modules/onboarding/dialog.tsx
```diff
@@ -0,0 +1,212 @@
...
+  const showBackButton = step > 0 && step < steps.length - 1;
+  const showTitle = step !== steps.length - 1;
+  const showAnimatedBackground = step === steps.length - 1;
+
+  // Note: set open to true or to isOnboardingVisible.get() to see the onboarding flow
+  // Currently stubbed as we don't have a way to create a profile yet
+  return (
+    <DialogContainer open={false} label="Entity search">
+      <div className="flex justify-between items-center pb-12">
+        <div className={`rotate-180`}>
```
baiirun, line 201: Shouldn't need the curly bois here.

apps/web/modules/onboarding/dialog.tsx
```diff
@@ -0,0 +1,212 @@
...
+  };
+
+  if (!address) return null;
+
+  const steps = [
+    <StepWallet nextStep={nextStep} address={address} key="wallet" />,
+    <StepName nextStep={nextStep} setName={setName} name={name} key="name" />,
+    <StepAvatar nextStep={nextStep} avatar={avatar} setAvatar={setAvatar} name={name} address={address} key="avatar" />,
+    <StepSuccess key="success" />,
+  ];
```
baiirun, line 190: Another way to do this is to name the steps to something semantic.

```tsx
type Steps = "wallet" | "name" | "avatar" | "success"

const [step, setStep] = useStep<Steps>("wallet");
```

Then we can pass in which step that `nextStep` should go to.

```tsx
<StepWallet onNext={() => setStep("name")} onPrevious={undefined} address={address} key="wallet" />
<StepName onNext={() => setStep("avatar")} onPrevious={() => setStep("wallet") address={address} key="wallet" />,
```

This is nice since you don't need all the error checks for the step index value and it's pretty clear how each step progresses to the next or previous. We do the same thing in `FlowBar`

apps/web/modules/onboarding/use-onboarding.tsx
```diff
@@ -0,0 +1,23 @@
+import { useObservable } from '@legendapp/state/react';
+import { useAccount } from 'wagmi';
+import { useGeoProfile } from '../auth/use-geo-profile';
+
+export function useOnboarding() {
+  /* Note! Stub hook! Need to have profile backend ready... */
+  const { profile } = useGeoProfile();
+  const isOnboardingVisible = useObservable(false);
```
baiirun, line 8: Any reason we're using `useObservable` here instead of regular React state?

baiirun (changes requested):

apps/web/modules/onboarding/dialog.tsx
makeitrein: Yee, cleaned it up

apps/web/modules/onboarding/dialog.tsx
```diff
@@ -0,0 +1,212 @@
...
+export function StepName({
+  nextStep,
+  name,
+  setName,
+}: {
+  nextStep: () => void;
+  name: string;
+  setName: (name: string) => void;
+}) {
+  const validName = name.length > 3;
```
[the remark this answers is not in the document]
makeitrein: Nope, made it a minimum length of 1. There is a product requirement of a disabled state w/ no text.

apps/web/modules/design-system/icons/close.tsx
```diff
@@ -0,0 +1,18 @@
...
+  const theme = useTheme();
+  const themeColor = color ? theme.colors[color] : 'currentColor';
+
+  return (
+    <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" fill="none">
+      <path d="M2 14L14 2" stroke={themeColor} />
+      <path d="M2 2L14 14" stroke={themeColor} />
+    </svg>
+  );
+}
```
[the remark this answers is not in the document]
makeitrein: Yup removed!

apps/web/modules/onboarding/use-onboarding.tsx
```diff
@@ -0,0 +1,26 @@
...
+      // If the user autoconnects, we don't want to show the modal.
+      if (address && !isReconnected) {
+        isOnboardingVisible.set(true);
+      }
+    },
+  });
+
+  const hideOnboarding = () => isOnboardingVisible.set(false);
+
+  return { isOnboardingVisible, hideOnboarding };
```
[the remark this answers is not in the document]
makeitrein: Got it

apps/web/modules/utils.ts
```diff
@@ -63,3 +63,7 @@ export function partition<T>(array: T[], predicate: (value: T) => boolean): [T[]
     [[], []]
   );
 }
+
+export function formatAddress(address: string): string {
```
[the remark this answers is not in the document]
makeitrein: Updated

apps/web/modules/onboarding/dialog.tsx
```diff
@@ -0,0 +1,351 @@
...
+  const [avatar, setAvatar] = useState('');
+  const [step, setStep] = useState<Steps>('wallet');
+
+  if (!address) return null;
+
+  // Note: set open to true or to isOnboardingVisible to see the onboarding flow
+  // Currently stubbed as we don't have a way to create a profile yet
+  // Also note that setting open to true will cause SSR issues in dev mode
+  return (
+    <Command.Dialog open={false} label="Onboarding profile">
```
baiirun, line 33: There's a known bug in here somewhere that Nate discovered when disconnecting and reconnecting your wallet. We can revisit later when we're ready to integrate Profiles on the backend.

PR #215, opened 2023-02-16
Sort fields in browse mode, always show name and description in edit mode

makeitrein:


apps/web/modules/components/entity/edit-events.ts
```diff
@@ -350,18 +350,6 @@ const listener =
...
-            {
-              ...triple,
-              entityName: value,
-              placeholder: false,
-              value: { ...triple.value, type: 'string', value },
-            },
-            triple
-          );
-        }
-
```
baiirun, line 364: Are we sure this doesn't break anything for column editing?

apps/web/modules/components/entity/editable-fields.tsx
```diff
@@ -29,6 +29,10 @@ export function StringField({ variant = 'body', ...props }: StringFieldProps) {
   const [localValue, setLocalValue] = React.useState(props.value || '');
   const ref = useRef<HTMLTextAreaElement>(null);
 
+  useEffect(() => {
+    setLocalValue(props.value || '');
+  }, [props.value]);
+
```
baiirun, line 35: Are we sure we need this?

makeitrein: Yup, otherwise the value is cached on initial mount.

makeitrein: But now that I typed that out, let me double check...

apps/web/modules/components/entity/edit-events.ts
baiirun: Ah I think at some point I was using `UPDATE_VALUE` for column name editing but I guess I changed it to `EDIT_ENTITY_NAME`

makeitrein: Bingo! I think we can merge this come Monday

apps/web/modules/components/entity/editable-fields.tsx
makeitrein: Yup, we need a useEffect to reset the input when the value changes from outside the component. Just asked Yaniv and Nate if this is desired behavior, feels a little strange to me atm.

apps/web/modules/components/entity/readable-entity-page.tsx
```diff
@@ -92,7 +95,7 @@ export function ReadableEntityPage({ triples, id, name, space, linkedEntities }:
 
       <Content>
         <Attributes>
-          <EntityAttributes entityId={id} triples={triples} space={space} />
+          <EntityAttributes entityId={id} triples={sortedTriples} space={space} />
```
baiirun, line 98: What is this change doing?

baiirun (commented):
Couple questions

apps/web/modules/components/entity/readable-entity-page.tsx
makeitrein: Ensuring that triples are sorted the same way in both edit mode and read mode

baiirun (approved):

PR #216, opened 2023-02-17
feat: refactor

sandren:
This PR completes the following refactors:

- future proofs the React import style for TSX
- sorts imports according to packages and local code
- updates the ESLint, Prettier, PostCSS, and Tailwind configurations (no rule changes)
- replaces styled components (Emotion) with Tailwind
- adds the classnames package for easier styling with Tailwind
- creates a new `<Icon />` component


baiirun:
There's some places where styles are off. I'll update this comment as I find them.

This should have a divider between the dropdown list items.
CleanShot 2023-02-16 at 17 12 33@2x
https://www.figma.com/file/iWKsYButqqKWd1bqeBrUUj/Geo---Web?node-id=3991%3A38293&t=JcbBVmD1ERN8y1gu-4

apps/web/styles/styles.css
```diff
@@ -1,95 +1,85 @@
...
+  }
 }
 
-*,
-::before,
-::after {
-  border-width: 0;
-  border-style: solid;
+.animate-fade-in {
+  animation: fadeIn 0.15s ease-in-out`;
```
baiirun, line 82: Where are we using this? Can we just define this in tailwind's config instead?

baiirun (commented):
Just a couple small things

apps/web/styles/styles.css
sandren: We're using it in the `card.tsx` component. We can move the definition to the Tailwind config.

PR #219, opened 2023-02-21
feat: user presence on entity page and entity table

baiirun:
This PR adds functionality for users to who else is editing the same entity as them. It uses [Liveblocks](https://liveblocks.io) to handle syncing state between clients.


apps/web/modules/components/avatar.tsx
```diff
@@ -0,0 +1,9 @@
+import BoringAvatar from 'boring-avatars';
```
baiirun, line 1: Grabbed this from the Onboarding PR

PR #226, opened 2023-02-28
feat: image type and upload component

makeitrein:
Demo Video: https://share.cleanshot.com/bHlxMcmk


packages/action-schema/src/action.ts
```diff
@@ -37,9 +37,15 @@ interface StringValue {
   value: string
 }
 
+interface ImageValue {
```
makeitrein, line 40: The files below are mainly ceremony to add a new value type

apps/web/modules/components/entity/edit-events.ts
```diff
@@ -449,6 +494,32 @@ const listener =
...
+        const newValue = { ...triple.value, value: imageSrc };
+
+        return update(
+          Triple.ensureStableId({
+            ...triple,
+            value: newValue,
+          }),
+          triple
+        );
+      }
```
baiirun, line 522: Is the difference between `CREATE_IMAGE_TRIPLE_WITH_VALUE` and `UPLOAD_IMAGE` basically the difference between placeholder and non-placeholder image fields?

apps/web/modules/components/entity/editable-fields.tsx
```diff
@@ -81,3 +84,142 @@ export function PageStringField({ ...props }: PageStringFieldProps) {
...
+    },
+    avatar: {
+      height: 44,
+      width: 44,
+    },
+    banner: {
+      height: 44,
+      width: 240,
+    },
+  };
```
baiirun, line 108: These styles can probably live outside the component so we're not recreated them every time the component re-renders.

apps/web/modules/components/entity/edit-events.ts
makeitrein: Less placeholders, more like in the editable entity page, we have an existing triple to upload to, but in the table, it's everything (triple + value) must be created at once

apps/web/modules/components/entity/editable-fields.tsx
makeitrein: Yup yup

apps/web/modules/components/entity/edit-events.ts
```diff
@@ -247,15 +261,20 @@ const listener =
       case 'CHANGE_TRIPLE_TYPE': {
         const { type, triples } = event.payload;
 
+        const value = emptyValue(type);
```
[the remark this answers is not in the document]
makeitrein: Good catch

apps/web/modules/components/entity/entity-page-utils.tsx
```diff
@@ -9,30 +9,30 @@ export function sortEntityPageTriples(visibleTriples: Triple[], schemaTriples: T
 
   /* Visible triples includes both real triples and placeholder triples */
   return visibleTriples.sort((tripleA, tripleB) => {
```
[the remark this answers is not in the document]
makeitrein: Added some clarity in the comment above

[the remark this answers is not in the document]
baiirun: It's still not clear to me what the actual sort order is without diving through the code

PR #232, opened 2023-03-08
adding and removing typeIds

MercuricChloride:
Here is an example endpoint with these changes deployed: https://thegraph.com/hosted-service/subgraph/mercuricchloride/geo-fulltext-master

For querying and filtering columns you can do something like this:

```
{
  triples(where: { 
    attribute: "Description",
    entity_: {
      typeIds_contains: ["f4f48574-71cd-412d-9769-11fc02c8975f"],
    }
  }, orderBy: stringValue) {
    id
    
    stringValue
    entityValue {
      id
      name
    }
    
    attribute {
      id
      name
    }
    
    entity {
      id
      name
    }
    
  }
}
```

### DIFF
--- a/packages/subgraph/src/actions.ts
+++ b/packages/subgraph/src/actions.ts
@@ -48,6 +48,35 @@ class HandleCreateTripleActionOptions {
   createdAtBlock: BigInt
 }
 
+function addEntityTypeId(entity: GeoEntity, valueId: string): void {
+  log.debug(`Adding type ID ${valueId} to entity ${entity.id}`, [])
+  entity.typeIds = entity.typeIds.concat([valueId]);
+  entity.save()
+}
+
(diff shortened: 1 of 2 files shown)


packages/subgraph/src/actions.ts
```diff
@@ -151,6 +188,24 @@ function handleDeleteTripleAction(
     return
   }
 
+  if (fact.attributeId == TYPES) {
+    const entity = GeoEntity.load(fact.entityId)
```
baiirun, line 192: Will this break if the entity doesn't exist?

Right now it's not possible within _our_ UI to create a triple with an attribute that doesn't exist, but someone could theoretically add something manually without a valid entity as the attribute. Should probably use `getOrCreateEntity` here.

baiirun (approved):

PR #234, opened 2023-03-13
feat: history list

baiirun:
This PR adds the initial infrastructure for reading entity versions on the entity page. Future PRs will add the same functionality to spaces, as well as reviewing versions and reverting/restoring them.

**Demo here**
https://share.cleanshot.com/K9mBwQVg

### TODO
- [x] create history list item
- [x] create history panel
- [x] integrate with entity page

### Future PRs
- [ ] integrate with space page
- [ ] create a "hidden" page for viewing all versions historically (might be cool for visualization)
- [ ] reviewing changes in older versions
- [ ] add revert/reset actions on history item hover


apps/web/modules/design-system/panel.tsx
```diff
@@ -0,0 +1,42 @@
+import { PopoverContent, Root, Trigger } from '@radix-ui/react-popover';
```
baiirun, line 1: Reusable component for rendering a list in a panel. We use this same styling in other places so can probably refactor to use this Panel component with child composition instead of their own implementations of Popover.

apps/web/modules/types.ts
```diff
@@ -121,3 +121,16 @@ export interface Cell {
...
+  id: string;
+  name: string;
+  description?: string;
+  createdBy: {
+    id: string;
+    name?: string;
+    avatarUrl?: string;
+  };
+  createdAt: number;
+  actions: Array<Action>;
```
baiirun, line 135: Close approximation of our versioning type from the subgraph

apps/web/pages/dev/history.tsx
```diff
@@ -0,0 +1,50 @@
+import { HistoryItem } from '~/modules/components/history';
```
baiirun, line 1: This is a temporary, dev page for rendering the history component. I didn't want to set up Storybook, but we probably should at some point.

apps/web/modules/services/io/queries.ts
baiirun, line 1: This is likely the first step in separating the read/write functionality when fetching data and publishing changes

apps/web/modules/types.ts
```diff
@@ -121,3 +121,16 @@ export interface Cell {
...
+
+export type Version = {
+  id: string;
+  name: string;
+  description?: string;
+  createdBy: {
+    id: string;
+    name?: string;
+    avatarUrl?: string;
+  };
```
baiirun, line 133: We'll eventually have a Profile type that encapsulates these properties

apps/web/modules/utils.ts
```diff
@@ -67,3 +67,13 @@ export function partition<T>(array: T[], predicate: (value: T) => boolean): [T[]
 export function formatShortAddress(address: string): string {
   return address.slice(0, 8) + '...' + address.slice(-6);
 }
+
+/**
+ * We return blocktime from the subgraph for createdAt and updatedAt fields.
+ * JavaScript date expects milliseconds, so we need to convert from seconds.
+ */
+export class GeoDate {
```
baiirun, line 75: Making this class to namespace GeoDate functionality. `GeoDate.fromGeoTime(...)`

apps/web/modules/config/config.ts
```diff
@@ -34,7 +34,8 @@ export const options: Record<AppEnv, AppConfig> = {
     chainId: '137',
     rpc: 'https://polygon-rpc.com',
     ipfs: 'https://api.thegraph.com/ipfs',
-    subgraph: 'https://api.thegraph.com/subgraphs/name/baiirun/geo',
+    // subgraph: 'https://api.thegraph.com/subgraphs/name/baiirun/geo',
+    subgraph: 'https://api.thegraph.com/subgraphs/name/mercuricchloride/geo-testnet',
```
baiirun, line 38: @baiirun DON'T FORGET TO UPDATE THIS ONCE THE SUBGRAPH FINISHES BUILDING

PR #237, opened 2023-03-16
feat: versioning in subgraph

MercuricChloride:


packages/subgraph/schema.graphql
```diff
@@ -56,12 +56,68 @@ type GeoEntity @entity {
...
+  entity: GeoEntity! # entity we are describing
+  createdAt: BigInt! # block number
+  actions: [Action!]! # All actions to be taken if this version is finalized
+}
+
+type Version @entity {
+  id: ID!
+  name: String
+  description: String
+  author: Account
```
baiirun, line 95: The author properties should be called `createdBy`. We want to use `author` in a different context in the future for referencing original authors for things like references, books, etc.

packages/subgraph/schema.graphql
```diff
@@ -56,12 +56,68 @@ type GeoEntity @entity {
...
+type ActionCount @entity {
+  id: ID! #always 1
+  count: BigInt!
+}
+
+type ProposedVersion @entity {
+  id: ID!
+  name: String
+  description: String
+  author: Account
```
baiirun, line 85: See https://github.com/geobrowser/geogenesis/pull/237/files#r1139598374

packages/subgraph/src/actions.ts
```diff
@@ -221,38 +318,201 @@ function handleDeleteTripleAction(
...
+      valueId = stringValue.id
+      strValue = stringValue.value
+    }
+    let numberValue = value.asNumberValue()
+    let numValue: string | null = null
+    if (numberValue != null) {
+      valueId = numberValue.id
+      numValue = numberValue.value
+    }
+    let action = getOrCreateAction(
```
baiirun, line 399: Should we return the action (and whatever else) so we can add it to the proposal we created in `handleRoot`?

packages/subgraph/src/actions.ts
```diff
@@ -72,11 +82,85 @@ export function getOrCreateEntity(id: string): GeoEntity {
...
+    version = new ProposedVersion(versionId)
+    version.createdBy = getOrCreateAccount(createdBy).id
+    version.actions = actions // action ids
+    version.entity = entityId
+    version.createdAt = createdAt
+    version.save()
+  }
+  let proposal = Proposal.load(proposalId)
+  if (proposal != null) {
+    proposal.proposedVersions = proposal.proposedVersions.concat([versionId])
```
baiirun, line 110: For some reason the subgraph data for every proposal is an empty array
image

baiirun: Also is `createdAt` supposed to be a string? Is that how graph-node represents BigInt?

packages/subgraph/src/actions.ts
MercuricChloride: We could follow a similar pattern, but not sure if it is needed since we can grab the actions from the proposedVersion

packages/subgraph/src/actions.ts
```diff
@@ -221,38 +318,201 @@ function handleDeleteTripleAction(
...
+      value.type.toUpperCase(),
+      valueId,
+      numValue,
+      strValue,
+      entValue
+    )
+    let proposed = createProposedVersion(
+      entityId + '-' + getOrCreateActionCount().count.toString(),
+      createdAt,
+      [action.id],
```
baiirun, line 413: We should be creating a proposed version from _all_ of the actions that operate on the same entity in a given entry. Right now we're creating a proposed version for each action.

I'd probably expect that we have a function called in `handleRoot` that does all the proposal related stuff.

```ts
function handleRoot(root: Root, space: string, createdAtBlock: BigInt): void {
  handleProposalRoot(...)

  for (let i = 0; i < root.actions.length; i++) {
    const action = root.actions[i]

    handleAction(action, space, createdAtBlock)
  }
}

function handleProposalRoot(...) {
  // group actions by entity
  // create proposed versions by entity
  // create version by entity
  // create proposal from actions
}
```

MercuricChloride: The changes I made should resolve this

PR #239, opened 2023-03-17
feat: review changes

sandren:
This PR implements the new review changes UI and a few other features.

### Changes
- new flowbar, review changes panel, and statusbar
- `localStorage` persistence for actions
- restores action squashing on review/publish
- new `<ClientOnly>` component for client-side rendering
- upgrade to Framer Motion 10

### Notes
- the proposal name functionality hasn't been implemented


apps/web/modules/components/main.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export const Main = ({ children }: MainProps) => {
+  const { isReviewOpen } = useReview();
+
+  return (
+    <motion.main
+      variants={variants}
+      animate="animate"
+      transition={transition}
+      custom={isReviewOpen}
```
baiirun, line 18: How does this `custom` prop work?

apps/web/modules/design-system/dropdown.tsx
```diff
@@ -11,10 +11,11 @@ const MotionContent = motion(DropdownPrimitive.Content);
 
 interface Props {
   trigger: React.ReactNode;
+  align?: 'end' | 'center' | 'start' | undefined;
```
baiirun, line 14: Shouldn't need the `undefined` state since the `?` adds it automatically

apps/web/modules/design-system/icon.tsx
```diff
@@ -51,7 +52,10 @@ export const Icon = ({ icon, ...rest }: IconProps) => {
   return <Component {...rest} />;
 };
 
+const Blank = () => <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg" />;
```
baiirun, line 55: What's this blank svg used for?

apps/web/modules/providers.tsx
```diff
@@ -1,14 +1,17 @@
...
 import { WalletProvider } from './wallet';
 
 export function Providers({ children }: { children: React.ReactNode }) {
   return (
     <Services.Provider>
       <ActionsStoreProvider>
-        <WalletProvider>{children}</WalletProvider>
+        <ReviewProvider>
+          <WalletProvider>{children}</WalletProvider>
+        </ReviewProvider>
```
baiirun, line 14: We can move this `ReviewProvider` to be more closely scoped with the `FlowBar` and `Review` components

apps/web/modules/review.tsx
```diff
@@ -0,0 +1,67 @@
+import * as React from 'react';
+import { useState, useCallback, createContext, useContext } from 'react';
+
+type ReviewState = {
+  isReviewOpen: boolean;
+  setIsReviewOpen: (value: boolean) => void;
+  toggleIsReviewOpen: () => void;
```
baiirun, line 7: Do we need both the setter and a toggle function?

apps/web/modules/review.tsx
```diff
@@ -0,0 +1,67 @@
...
+};
+
+const initialReviewState = {
+  isReviewOpen: false,
+  setIsReviewOpen: (value: boolean) => null,
+  toggleIsReviewOpen: () => null,
+  activeSpace: '',
+  setActiveSpace: (value: string) => null,
+  isReadyToPublish: false,
+  setIsReadyToPublish: (value: boolean) => null,
```
baiirun, line 21: Do we need all the initial state? Can probably just set the ones that aren't overwritten by the React body directly in the initial `value` on the Provider.

```tsx
<ReviewContext.Provider
      value={{
        isReviewOpen,
        setIsReviewOpen,
        toggleIsReviewOpen,
        activeSpace,
        setActiveSpace,
        isReadyToPublish,
        setIsReadyToPublish,
      }}
    >
      {children}
    </ReviewContext.Provider>
```

apps/web/modules/review.tsx
```diff
@@ -0,0 +1,67 @@
...
+const ReviewContext = createContext<ReviewState>(initialReviewState);
+
+type ReviewProviderProps = {
+  children: React.ReactNode;
+};
+
+export const ReviewProvider = ({ children }: ReviewProviderProps) => {
+  const [isReviewOpen, setIsReviewOpen] = useState<boolean>(false);
+
+  const toggleIsReviewOpen = useCallback(() => {
```
baiirun, line 33: Don't need memoization for this really

apps/web/modules/components/review.tsx
```diff
@@ -0,0 +1,494 @@
...
+
+  // Set a new default active space when active spaces change
+  useEffect(() => {
+    if (allSpacesWithActions.length === 0) {
+      setIsReviewOpen(false);
+      return;
+    }
+
+    setActiveSpace(allSpacesWithActions[0] ?? '');
+  }, [allSpacesWithActions]); // eslint-disable-line react-hooks/exhaustive-deps
```
baiirun, line 63: Why are we disabling the lint rule? Usually that's a code smell and will generate bugs that we don't expect.

apps/web/modules/components/review.tsx
```diff
@@ -0,0 +1,494 @@
...
+        disabled: activeSpace === spaceId,
+        onClick: () => setActiveSpace(spaceId),
+      })),
+    [activeSpace, allSpacesWithActions] // eslint-disable-line react-hooks/exhaustive-deps
+  );
+
+  // Proposal state
+  const [reviewState, setReviewState] = useState<ReviewState>('idle');
+  const [proposalName, setProposalName] = useState<string>('');
+  const isReadyToPublish = proposalName.length > 3;
```
baiirun, line 94: Why are we limiting the name length? Did this come up as a product limitation? We probably want some UI to denote that there's a condition on the name in order to publish content.

apps/web/modules/components/review.tsx
```diff
@@ -0,0 +1,494 @@
...
+              {allSpacesWithActions.length === 1 && (
+                <span className="inline-flex items-center gap-2 text-button text-text ">
+                  <span className="relative h-4 w-4 overflow-hidden rounded-sm">
+                    <img
+                      src={
+                        spaces.find(({ id }) => id === activeSpace)?.attributes[SYSTEM_IDS.IMAGE_ATTRIBUTE] ??
+                        'https://via.placeholder.com/600x600/FF00FF/FFFFFF'
+                      }
+                      className="absolute inset-0 h-full w-full object-cover object-center"
+                      alt=""
```
baiirun, line 124: This logic for finding the image value or setting a default is used in lots of places. Can probably abstract it so it's easier to change in the future.

apps/web/modules/components/review.tsx
```diff
@@ -0,0 +1,494 @@
...
+  before?: Before;
+  after?: After;
+  differences?: Array<Difference>;
+};
+
+type AttributeName = string;
+type Before = string[];
+type After = string[];
+
+const useChanges = (actions: Array<any>) => {
```
baiirun, line 240: Why is this `Array<any>` instead of `Array<Action>`?

apps/web/modules/components/review.tsx
```diff
@@ -0,0 +1,494 @@
...
+  return useMemo(() => getChanges(actions as Array<Action>), [actions]);
+};
+
+const getChanges = (actions: Array<any>): Changes => {
```
baiirun, line 244: We should def write a test for this.

apps/web/modules/components/review.tsx
```diff
@@ -0,0 +1,494 @@
...
+const Chip = ({ status = 'unchanged', children }: ChipProps) => {
+  return (
+    <span
+      className={cx(
+        status === 'added' && 'bg-green/10',
+        status === 'removed' && 'bg-grey-02 line-through',
+        status === 'unchanged' && 'bg-white',
+        'inline-flex min-h-[1.5rem] items-center rounded-sm px-2 py-1 text-left text-metadataMedium shadow-inner shadow-text'
+      )}
+    >
```
baiirun, line 478: This is a usecase where I'd probably use `cva` since the syntax for expressing different states is a lot more explicit IMO

apps/web/modules/action/action.ts
```diff
@@ -94,6 +94,32 @@ function getFirstAndLastChanges(actions: ActionType[]) {
...
+
+      // This doesn't need to go to the subgraph at all.
+      if (changeTuple[0].type === 'createTriple' && changeTuple[1].type === 'deleteTriple') {
+        return null;
+      }
+
+      return changeTuple[1];
+    })
+    .flatMap(changeTuple => (changeTuple ? changeTuple : []));
+}
```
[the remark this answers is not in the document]
sandren: I'll look into that!

apps/web/modules/action/use-actions-store.ts
```diff
@@ -9,26 +11,40 @@ import { useActionsStoreContext } from './actions-store-provider';
  * on a dev route or the root /spaces page.
  */
 export function useActionsStore(spaceId?: string) {
-  const { actions$, publish, clear, create, update, remove } = useActionsStoreContext();
+  const { actions$, publish, clear, create, update, remove, deleteActions } = useActionsStoreContext();
   const actions = useSelector(actions$);
+  const allActions: Array<Action> = useMemo(
+    () => Object.entries(actions).flatMap(([, actions]) => actions) ?? [],
```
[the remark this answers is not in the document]
sandren: True! 😅

apps/web/pages/_app.tsx
```diff
@@ -72,23 +75,16 @@ function App({ Component, pageProps }: AppProps) {
...
         spaceId=""
       />
-      <main className="mx-auto max-w-[1200px] pt-10 pb-20 xl:pt-[40px] xl:pr-[2ch] xl:pb-[4ch] xl:pl-[2ch]">
+      <Main>
         <Component {...pageProps} />
         <Analytics />
-      </main>
-      <GlobalFlowBar spaceId={spaceId ?? ''} />
+      </Main>
+      <ClientOnly>
```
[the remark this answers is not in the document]
sandren: Yes, I'll add a note of that.

apps/web/modules/design-system/icons/close.tsx
```diff
@@ -1,4 +1,6 @@
-import { ColorName, colors } from '~/modules/design-system/theme/colors';
+import * as React from 'react';
```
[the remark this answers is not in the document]
sandren: The new JSX transform works for JavaScript files, but in TypeScript, I actually get an ESLint error when _not_ importing the `React` namespace in files where JSX is used.

If you prefer to remove it, I can open a separate PR to do it globally and try and resolve the ESLint issue.

apps/web/modules/components/main.tsx
sandren: The `custom` prop passes its value to any variants that are specified as functions. 

See: https://www.framer.com/motion/component/###custom

apps/web/modules/design-system/icon.tsx
sandren: It's currently only used for the staging/unstaging changes checkbox, but it may come in handy elsewhere. It lets you hide an icon while preserving the amount of space it takes up without messing with CSS classes.

apps/web/modules/components/review.tsx
sandren: I didn't want the selected space to inadvertently change on re-render, but I'll see if there's another way to implement this.

apps/web/modules/components/review.tsx
sandren: Yeah. I can move this to utils.

apps/web/modules/components/review.tsx
sandren: It _should_ be `Array<Action>`, but it was causing errors. I can try again and see what I find.

apps/web/modules/components/flow-bar.tsx
```diff
@@ -1,158 +1,47 @@
...
+import { Action, useActionsStore } from '~/modules/action';
+import { Button } from '~/modules/design-system/button';
+import { useReview } from '~/modules/review';
+import { groupBy } from '~/modules/utils';
+
+export const FlowBar = () => {
+  const { allActions, allSpacesWithActions } = useActionsStore();
+  const { setIsReviewOpen } = useReview();
+
+  const actionsCount = useMemo(() => Action.getChangeCount(allActions), [allActions]);
```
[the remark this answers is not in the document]
sandren: Not really. Just habit. 😅

apps/web/modules/components/review.tsx
baiirun: We probably want it to be required. More of a Yaniv question though.

apps/web/modules/providers.tsx
sandren: The review context is used in the `<Main>` component as well so the only part of the app that would benefit from a finer scope would be the navbar.

apps/web/modules/design-system/icons/close.tsx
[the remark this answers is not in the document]
baiirun: Weird. I _don't_ get an ESLint error when using JSX without importing the namespace. Maybe it's a VSCode setting? Probably worth us spending a little time on eslint tooling in general since it's not working exactly in CI either. We can just keep adding them for now until we do that. Maybe next week.

apps/web/modules/components/flow-bar.tsx
```diff
@@ -1,158 +1,69 @@
...
+    opacity: custom ? 1 : 0,
+    y: custom ? '0px' : '-4px',
+    transition: {
+      type: 'spring',
+      duration: 0.5,
+      bounce: 0,
+      delay: custom ? 0.5 : 0,
+    },
+  }),
+};
```
baiirun, line 71: Should this be animating from the bottom instead of the top? Makes more sense with the flowbar being at the bottom of the page.

apps/web/pages/_app.tsx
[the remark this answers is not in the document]
baiirun: Pinging this one 😺

baiirun (commented):
Looks good! Just some minor things to clean up then looks good to go!

apps/web/modules/design-system/dropdown.tsx
baiirun: Pinging this one again

apps/web/modules/review.tsx
baiirun: Pinging this one again

apps/web/modules/review.tsx
baiirun: Pinging for the memoization stuff

apps/web/modules/components/review.tsx
baiirun: If we're keeping this we want a comment to describe why we're ignoring the lint rule.

apps/web/modules/components/review.tsx
```diff
@@ -0,0 +1,568 @@
...
+const RevisedEntity = ({
+  spaceId,
+  entityId,
+  entityName,
+  entityRevisions,
+  unstagedChanges,
+  setUnstagedChanges,
+}: RevisedEntityProps) => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [entity, setEntity] = useState<any>(null);
```
baiirun, line 338: This should probably be typed.

apps/web/modules/components/review.tsx
baiirun: Pinging this one again for adding tests, especially since the Array is `any` I'm not confident we can change this in the future without causing regressions.

baiirun (commented):
Some comments on types and tests in the review component

apps/web/modules/components/review.tsx
baiirun: We can fix the type in a follow-up PR if we want to get this one pass the finished line.

apps/web/modules/components/flow-bar.tsx
sandren: Oh, good catch!

apps/web/modules/components/review.tsx
baiirun: Pinging this one again

baiirun (approved):
🚀

baiirun (approved):

PR #240, opened 2023-03-20
feat: page editor

makeitrein:


apps/web/pages/space/[id]/[entityId].tsx
```diff
@@ -45,9 +44,12 @@ export default function EntityPage(props: Props) {
   return (
     <EntityStoreProvider
       id={props.id}
+      name={props.name}
```
Comment (makeitrein) [inline apps/web/pages/space/[id]/[entityId].tsx:47]: Name and the blockIdsTriple and blockTriples are now attributes of EntityStoreProvider

Perhaps name should be computed off triples, would that be cleaner @baiirun

apps/web/modules/components/entity/editor/command-extension.tsx
```diff
@@ -0,0 +1,92 @@
...
+    ];
+  },
+});
+
+// Inspired By https://github.com/wenerme/wode/blob/b66696f9ba60038e9b86b62e2624aa36e4e91524/apps/demo/src/contents/TipTap/TipTapPageContent.tsx
+export const ConfiguredCommandExtension = CommandExtension.configure({
+  suggestion: {
+    items: ({ query }) => {
+      return commandItems
+        .filter(v => v.command)
```
baiirun, line 40: What is this filter doing? Don't all our commands in `command-items` have a command?

apps/web/modules/components/entity/editor/command-extension.tsx
```diff
@@ -0,0 +1,92 @@
...
+          popup = tippy('body', {
+            getReferenceClientRect: props.clientRect as any, // fixme
+            appendTo: () => document.body,
+            content: reactRenderer.element,
+            showOnCreate: true,
+            interactive: true,
+            trigger: 'manual',
+            placement: 'bottom-start',
+          });
+        },
```
baiirun, line 65: Can we somehow use Radix's popover elements instead of bringing in a new dependency?

apps/web/modules/components/entity/editor/command-list.tsx
```diff
@@ -0,0 +1,129 @@
...
+}
+
+type CommandListMode = 'select-block' | 'select-table';
+
+export const CommandList = forwardRef<CommandListRef, CommandListProps>(({ command, items }, ref) => {
+  const entityStore = useEntityStore();
+  const [selectedIndex, setSelectedIndex] = useState(0);
+  const [mode, setMode] = useState<CommandListMode>('select-block');
+
+  const tableItem = items.find(item => item.title === 'Table') as CommandSuggestionItem;
```
baiirun, line 29: Could this ever return undefined? When casting like this we gotta be careful that we aren't accidentally hiding a bug.

apps/web/modules/components/entity/editor/command-list.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  const handleTableSelect = (selectedType: SelectedEntityType) => {
+    command({ ...tableItem, selectedType: selectedType, spaceId: entityStore.spaceId });
+  };
+
+  const invokeItem = (item: CommandSuggestionItem) => {
+    if (!item) {
```
baiirun, line 36: When would we not pass an item? Based on the type parameter I'd expect it to always be non-null/undefined.

apps/web/modules/components/entity/editor/command-list.tsx
```diff
@@ -0,0 +1,129 @@
...
+    } else if (item.title === 'Table' && mode === 'select-block') {
+      setMode('select-table');
+    } else {
+      command(item);
+    }
+  };
+
+  const containerRef = useRef<HTMLDivElement>(null);
+  useEffect(() => setSelectedIndex(0), [items]);
+  useImperativeHandle(ref, () => ({
```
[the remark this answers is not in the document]
baiirun: Why are we using `useIMperativeHandle` instead of a `useEffect`?

apps/web/modules/components/entity/editor/command-list.tsx
```diff
@@ -0,0 +1,129 @@
...
+    const min = $ele.offsetTop;
+    if (min < top) {
+      $div.scrollTop = min;
+      return;
+    }
+    const max = min + $ele.clientHeight;
+    const h = $div.clientHeight;
+    if (max > top + h) {
+      $div.scrollTop = max - h;
+    }
```
baiirun, line 87: Why do we need this?

apps/web/modules/components/entity/editor/editor.tsx
```diff
@@ -0,0 +1,54 @@
...
+    },
+    [editable]
+  );
+
+  return editor ? (
+    <>
+      <EditorContent editor={editor} />
+      <FloatingMenu editor={editor}>
+        <div className="absolute -left-12 -top-3">
+          <SquareButton onClick={() => editor.chain().focus().insertContent('/').run()} icon="plus" />
```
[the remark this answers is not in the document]
baiirun: Can you turn this into a function called `openCommandMenu` or something so it's clear what this chain of calls does?

apps/web/modules/components/entity/editor/id-extension.tsx
```diff
@@ -0,0 +1,37 @@
...
+        attributes: {
+          id: {
+            default: null,
+            keepOnSplit: false,
+          },
+        },
+      },
+    ];
+  },
+  onBlur() {
```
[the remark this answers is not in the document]
baiirun: Can you add a comment for this one?

apps/web/modules/components/entity/editor/table-node.tsx
```diff
@@ -0,0 +1,78 @@
...
+  },
+});
+
+export const TableNodeComponent = React.memo(function TableNodeComponent({ node }: NodeViewRendererProps) {
+  const { spaceId, typeId } = node.attrs;
+
+  const { types } = useEntityTable();
+
+  const selectedType = useMemo(() => {
+    return types.find(type => type.entityId === typeId) as Triple;
```
baiirun, line 53: Are we sure this cast isn't hiding a bug?

apps/web/modules/components/entity/editor/table-node.tsx
```diff
@@ -0,0 +1,78 @@
...
+  }, [types.length, typeId]); // eslint-disable-line react-hooks/exhaustive-deps
```
baiirun, line 54: Are we sure this won't cause stale data if we get two `types` arrays that are the same length? We could stringify the `types` array if we know it's never very large.

apps/web/modules/components/entity/editor/table-node.tsx
```diff
@@ -0,0 +1,78 @@
...
+  return (
+    <NodeViewWrapper className="react-component-with-content">
+      <div contentEditable="false">
+        <TableNodeChildren spaceId={spaceId} selectedType={selectedType} />
+      </div>
+    </NodeViewWrapper>
+  );
+});
+
+export const TableNodeChildren = React.memo(function TableNodeComponent({
```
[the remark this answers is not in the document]
baiirun: Do you know why it was re-rendering? I'd rather not memoize everything if we know the root cause and if performance isn't actually an issue.

apps/web/pages/space/[id]/[entityId].tsx
Comment (baiirun) [reply apps/web/pages/space/[id]/[entityId].tsx:47]: This is probably okay, but now there's places in the codebase where we read the entity name using triples and places where we don't.

apps/web/pages/space/[id]/[entityId]/v2.tsx
Comment (baiirun) [inline apps/web/pages/space/[id]/[entityId]/v2.tsx:1]: I'm making changes to the main `[entityId].tsx` page which will probably cause issues when you try and migrate this v2 page over. We might want to get in my versioning changes on the page before we migrate the editor over.

apps/web/modules/entity/entity-table-store/entity-table-store.ts
```diff
@@ -37,10 +37,10 @@ interface IEntityTableStoreConfig {
...
   initialParams?: InitialEntityTableStoreParams;
   pageSize?: number;
-  initialRows: Row[];
-  initialSelectedType: TripleType | null;
-  initialTypes: TripleType[];
-  initialColumns: Column[];
+  initialRows?: Row[];
+  initialSelectedType?: TripleType | null;
+  initialTypes?: TripleType[];
+  initialColumns?: Column[];
```
baiirun, line 43: Why are we making these optional?

baiirun (changes requested):

apps/web/modules/components/entity/editor/editor.tsx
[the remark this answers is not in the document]
makeitrein: Yee

apps/web/modules/components/entity/editor/id-extension.tsx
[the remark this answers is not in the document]
makeitrein: Yee

apps/web/pages/space/[id]/[entityId].tsx
Comment (makeitrein) [reply apps/web/pages/space/[id]/[entityId].tsx:47]: I'll switch it to a computed for cleanliness sake

apps/web/modules/components/entity/editor/command-extension.tsx
makeitrein: After the command menu opens, we can filter the menu options by typing in it! I left a comment, but we can also just remove this feature...

apps/web/modules/components/entity/editor/table-node.tsx
makeitrein: JSON.stringify works!

apps/web/modules/components/entity/editor/table-node.tsx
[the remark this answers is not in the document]
makeitrein: Good point. The problem is with `const { types } = useEntityTable();` - if we remove the useMemo and memoized TableNodeChildren component, the table will flicker on every blur. It looks like we can safely remove the other memoizations.

apps/web/modules/components/entity/editor/command-list.tsx
makeitrein: Good point!

apps/web/modules/components/entity/editor/command-list.tsx
makeitrein: Nuked - helps with scrolling to the proper item, but since we only have a few blocks to select from, we really don't need it

apps/web/modules/entity/entity-table-store/entity-table-store.ts
makeitrein: Whoops, that should not have been committed

apps/web/modules/components/entity/editor/command-extension.tsx
makeitrein: I attempted to follow the code used in https://github.com/ueberdosis/tiptap/issues/2305 for a more Reacty way of doing things, but Radix doesn't seem to expose a way to position a popover w/o a Popover.Trigger element.

apps/web/modules/components/entity/editor/command-list.tsx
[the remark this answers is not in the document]
makeitrein: It feels a little better to hook into ProseMirror's [handleKeyDown](https://prosemirror.net/docs/ref/#view.EditorProps.handleKeyDown) event and call it from their plugin rather than having a global document listener, but both could work!

baiirun (approved):

PR #242, opened 2023-03-23
feat: merge profile with versions on the server when fetching proposed versions

baiirun:
Right now there is no wallet -> profile mapping in the subgraph until we added permissionless spaces and profiles. We can create an ad-hoc profile system using GeoEntities and associated them with a Wallet entity.

This PR sets up infrastructure for querying a Profile GeoEntity based on the wallet address in a version's createdBy field. We then add the profile information to the version to populate the name and avatar.

---
**Demo here**
https://share.cleanshot.com/pn6v2wXS

### DIFF
--- a/apps/web/modules/services/io/queries.ts
+++ b/apps/web/modules/services/io/queries.ts
@@ -1,5 +1,5 @@
 export const proposedVersionsQuery = (entityId: string) => `query {
-  proposedVersions(where: {entity: ${JSON.stringify(entityId)}}, first: 10, orderBy: createdAt, orderDirection: desc) {
+  proposedVersions(where: {entity: ${JSON.stringify(entityId)}}, orderBy: createdAt, orderDirection: desc) {
     id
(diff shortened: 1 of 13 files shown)


apps/web/modules/services/io/queries.ts
```diff
@@ -1,5 +1,5 @@
 export const proposedVersionsQuery = (entityId: string) => `query {
-  proposedVersions(where: {entity: ${JSON.stringify(entityId)}}, first: 10, orderBy: createdAt, orderDirection: desc) {
+  proposedVersions(where: {entity: ${JSON.stringify(entityId)}}, orderBy: createdAt, orderDirection: desc) {
```
baiirun, line 2: We no longer take the first 10 versions and instead do it on the client. Only taking the first 10 at the query level will result in inaccurate contributor counts since we'd only be counting the contributors in the first 10 results.

PR #248, opened 2023-03-24
feat: enhance error logs in network.ts

baiirun:
We now log additional information for each fetch type, both for runtime errors and for request errors from our dependencies (like the hosted network).

Eventually we want to pipe these errors into some other source for us to consume and be alerted by.


apps/web/modules/components/filter/type-dialog.tsx
baiirun, line 1: Not sure how this file got pulled into this PR. It should've been resolved when I merged main into this branch.

PR #249, opened 2023-03-24
feat: integrate editor into entity page

baiirun:
This PR includes the initial integration of the text editor onto entity pages.
- There may be jank, particularly when switching between edit and browse modes
- This only includes support for text content types. The styling should be correct for all text content.
- There is probably some unknown edge case bugs that we haven't encountered or handled yet
- This includes some miscellaneous changes to styling in some places to align on shadows and animatons

Future PRs will include support for table and image blocks as well as cover and page images.


apps/web/modules/components/entity/editor/command-items.tsx
```diff
@@ -61,5 +61,5 @@ export const commandItems: CommandSuggestionItem[] = [
       editor.chain().focus().deleteRange(range).setNode('heading', { level: 3 }).run();
     },
   },
-  tableCommandItem,
+  // tableCommandItem,
```
baiirun, line 64: For now we aren't allowing table blocks until we work on migrating tables to the new design

PR #253, opened 2023-03-29
feat: entity page images

sandren:
This PR adds avatar and cover images to entity pages.

localhost_3000_space_0xe3d08763498e3247EC00A481F199B018f2148723_02834927-38c2-435d-b85b-979fa15c6d78

### DIFF
--- a/apps/web/pages/space/[id]/[entityId].tsx
+++ b/apps/web/pages/space/[id]/[entityId].tsx
@@ -1,5 +1,6 @@
-import type { GetServerSideProps } from 'next';
+import * as React from 'react';
 import { useEffect } from 'react';
+import type { GetServerSideProps } from 'next';
 import Head from 'next/head';
 import { SYSTEM_IDS } from '@geogenesis/ids';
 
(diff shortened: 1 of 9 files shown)


apps/web/pages/space/[id]/[entityId].tsx
```diff
@@ -77,6 +82,44 @@ export default function EntityPage(props: Props) {
...
+  avatarUrl: string | null;
+  coverUrl: string | null;
+};
+
+const Intro = ({ avatarUrl, coverUrl }: IntroProps) => {
+  if (coverUrl) {
+    return (
+      <div className="relative mx-auto mb-[80px] h-[320px] w-[1192px]">
+        <div className="h-full w-full overflow-hidden rounded bg-grey-01">
+          <img src={coverUrl} className="h-full w-full object-cover" alt="" />
```
Comment (baiirun) [inline apps/web/pages/space/[id]/[entityId].tsx:95]: We should be using the `next/image` component for Images for the optimization benefits. There's a quirk with the next/image component where we need to wrap it with a element that has `relative` styling to use the `layout="fill"` prop. There's several examples in the codebase.

apps/web/pages/space/[id]/[entityId].tsx
```diff
@@ -77,6 +82,44 @@ export default function EntityPage(props: Props) {
...
+    return (
+      <div className="mx-auto w-[784px]">
+        <div className="h-[80px] w-[80px] overflow-hidden rounded border border-white bg-grey-01 shadow-lg">
+          <img src={avatarUrl} className="h-full w-full object-cover" alt="" />
+        </div>
+      </div>
+    );
+  }
+
+  return <React.Fragment />;
```
Comment (baiirun) [inline apps/web/pages/space/[id]/[entityId].tsx:120]: It's probably more idiomatic to return null instead of an empty fragment.

Comment (sandren) [reply apps/web/pages/space/[id]/[entityId].tsx:120]: > It's probably more idiomatic to return null instead of an empty fragment.

I tried that at first, but it causes issues with TypeScript. I'll see if an explicit return type can help.

baiirun (approved):

PR #260, opened 2023-03-31
feat: image blocks

sandren:
This PR adds support for image blocks to the page editor.


apps/web/modules/entity/entity-store/entity-store.ts
```diff
@@ -181,6 +182,22 @@ export class EntityStore implements IEntityStore {
           const rowTypeTriple = blockTriples.find(
             triple => triple.entityId === blockId && triple.attributeId === SYSTEM_IDS.ROW_TYPE
           );
+          const imageTriple = blockTriples.find(
+            triple => triple.entityId === blockId && triple.attributeId === SYSTEM_IDS.IMAGE_BLOCK
```
baiirun, line 186: I don't think data model is correct. I'd expect to be reading from the `IMAGE` attribute off an `IMAGE_BLOCK` type since that's the schema for an Image Block.

Each Image Block has an attribute called `Image` that has a specific attributeId associated with it. Right now it looks like you're looking for the attributeId === IMAGE_BLOCK _on_ the Image Block 

CleanShot 2023-04-04 at 13 50 21@2x

apps/web/modules/entity/entity-store/entity-store.ts
```diff
@@ -183,16 +183,19 @@ export class EntityStore implements IEntityStore {
             triple => triple.entityId === blockId && triple.attributeId === SYSTEM_IDS.ROW_TYPE
           );
           const imageTriple = blockTriples.find(
-            triple => triple.entityId === blockId && triple.attributeId === SYSTEM_IDS.IMAGE_BLOCK
+            triple => triple.entityId === blockId && triple.attributeId === SYSTEM_IDS.IMAGE_ATTRIBUTE
           );
 
+          // @TODO remove console.info
+          console.info('imageTriple:', imageTriple);
```
baiirun, line 190: Still got this in here

apps/web/modules/triple/triple.ts
```diff
@@ -182,3 +182,16 @@ export function withLocalNames(actions: ActionType[], triples: Triple[]) {
...
+  switch (triple.value.type) {
+    case 'number':
+      return triple.value.value;
+    case 'string':
+      return triple.value.value;
+    case 'entity':
+      return triple.value.name;
+    case 'image':
+      return triple.value.value;
+  }
```
baiirun, line 196: We need to be a little careful with Entity values because the name isn't actually the value of the Entity value. The actual value is the `value.id`. `value.name` is just a helper to extract the name from that entity. The name _could_ be `null` if the entity doesn't have a `name` set.

apps/web/modules/entity/entity-store/entity-store.ts
```diff
@@ -533,32 +536,27 @@ export class EntityStore implements IEntityStore {
     }
   };
 
-  /* Helper function for upserting a new block image triple for IMAGE_BLOCKs only  */
-  upsertBlockImageTriple = (node: JSONContent) => {
+  /* Helper function for creating a new block image triple for IMAGE_BLOCKs only  */
+  createBlockImageTriple = (node: JSONContent) => {
```
baiirun, line 540: I'm assuming there's no way to replace an image in an`ImageBlock` from the editor right now? You'd have to delete the block then add a new one?

baiirun (approved):

apps/web/modules/entity/entity-store/entity-store.ts
sandren: Yeah. That's why I renamed it and got rid of the upsert logic.

baiirun (approved):

PR #263, opened 2023-04-03
feat: filter blocks from entity search

baiirun:
For now we don't want to clutter entity search with all the block entities within Geo. Eventually, once we have transclusion and other block-based reference features we may want to index entities or handle network filtering a different way. For now this works.


baiirun:
Creating this PR again since we reverted it because of a potential bug. Turns out the bug wasn't related to these changes.

PR #273, opened 2023-04-05
feat: client-side history fetch

baiirun:
This query can take longer since it requires several nested child entities. Since the history is hidden behind a button, we can wait to fetch the data on the client side instead of blocking server fetching.

Right now this use a placeholder, static loading state until Nate designs a more fleshed out loading state. Additionally, we don't animate the data loading in for now.


apps/web/package.json
```diff
@@ -55,11 +56,10 @@
     "react": "^18.1.0",
     "react-dom": "^18.1.0",
     "react-medium-image-zoom": "^5.1.3",
-    "react-query": "^3.39.2",
     "react-use-measure": "^2.1.1",
     "showdown": "^2.1.0",
     "tippy.js": "^6.3.7",
-    "wagmi": "^0.10.4"
+    "wagmi": "^0.9.6"
```
baiirun, line 62: We reverted to an old version of wagmi to align on the correct peer dependency for connectkit.

PR #288, opened 2023-04-24
fix: opengraph images

sandren:
This PR adjusts the social media card preview images to use the avatar (if available) for entity pages.


apps/web/pages/space/[id].tsx
```diff
@@ -33,14 +33,16 @@ export default function EntitiesPage({
   initialRows,
   initialTypes,
 }: Props) {
+  const opengraphUrl = spaceImage ?? 'https://www.geobrowser.io/static/geo-social-image.png';
```
Comment (baiirun) [inline apps/web/pages/space/[id].tsx:36]: We should probably add this to constants so we don't have to hardcode the fallback on every route.

baiirun (approved):
One comment, other than that looks good!

PR #324, opened 2023-05-17
feat: invert hidden columns to shown columns and add cursor gap

sandren:
This PR inverts the behavior of the previous `HIDDEN_COLUMNS` attribute to `SHOWN_COLUMNS` and provides a gap for cursors between and below table blocks.


sandren:
> Can we name this PR something more meaningful? Right now there's several "table enhancements" and "ux fixes" commits on main and it's not clear what each of them are really changing without going through each of the PRs individually.

Yeah. Sorry for all the "UX fixes" one. We were moving really fast before the demo and many of those PRs addressed multiple things. 😅

PR #325, opened 2023-05-17
feat: automatically populate new table names

sandren:
This PR automatically names table blocks with the plural of their type.


sandren:
Tests wil pass again once https://github.com/geobrowser/geogenesis/pull/328 is merged.

PR #326, opened 2023-05-17
feat: add support for multiline title editing

sandren:
This PR adds auto resizing for textareas.


baiirun (commented):
Can we make sure that the spacing between the editable text area aligns with the Figma designs? There were some edge cases with our custom textarea expansion that meant that the spacing we used in edit mode was different than in browse mode. See Line 78 in `EditableHeading`

CleanShot 2023-05-18 at 12 23 02@2x

sandren:
I revised the custom spacing so browse and edit modes now prefectly match. ☺️

baiirun (approved):

PR #331, opened 2023-05-19
feat: table block space filter option

baiirun:
This PR adds `Space` as a special option for filtering rows in a Table.


apps/web/package.json
```diff
@@ -58,7 +58,7 @@
     "react": "^18.1.0",
     "react-dom": "^18.1.0",
     "react-medium-image-zoom": "^5.1.3",
-    "react-textarea-autosize": "^8.4.1",
+    "react-textarea-autosize": "^8.4.0",
```
baiirun, line 61: 8.4.1 has type errors, so reverting to 8.4.0 for now https://github.com/Andarist/react-textarea-autosize/issues/368

PR #350, opened 2023-05-31
fix: revise table block styles

sandren:
This PR revises the table block styles to increase minimum cell width and number of default lines.

### DIFF
--- a/apps/web/modules/components/table/cell-content.tsx
+++ b/apps/web/modules/components/table/cell-content.tsx
@@ -11,7 +11,7 @@ interface Props {
   href?: string;
 }
 
-export function CellContent({ isExpanded, value, isEntity, href }: Props) {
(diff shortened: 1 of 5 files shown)


apps/web/modules/components/table/cell-content.tsx
```diff
@@ -25,7 +25,7 @@ export function CellContent({ isExpanded, value, isEntity, href }: Props) {
   );
 
   return (
-    <Truncate maxLines={isEntity ? 1 : 3} shouldTruncate={!isExpanded}>
+    <Truncate maxLines={3} shouldTruncate={!isExpanded}>
```
baiirun, line 28: Can we remove `isEntity` from the Props definition too?

baiirun (approved):

PR #354, opened 2023-06-02
fix: remove redirects

baiirun:
Redirects might be causing issues with hydration mismatches.


baiirun:
Not related to redirects

PR #361, opened 2023-06-06
feat: find or create

baiirun:
This PR adds new functionality for creating an entity directly in the entity search flow.

https://github.com/geobrowser/geogenesis/assets/26263630/8aa1be29-ab28-4106-9221-7b0eea8a1f04


apps/web/modules/components/flow-bar.tsx
```diff
@@ -28,32 +30,35 @@ export const FlowBar = () => {
 
   const spacesCount = allSpacesWithActions.length;
 
-  if (actionsCount === 0 || !editable) return null;
+  // Don't show the flow bar if there are no actions, if the user is not in edit mode, or if there is a toast
+  const hideFlowbar = actionsCount === 0 || !editable || toast;
```
baiirun, line 34: We now short circuit within the AnimatePresence so we have exit animations on the flow bar

PR #363, opened 2023-06-07
fix: ordering of table rows being re-sorted incorrectly client side

baiirun:
Previously we were applying client-side sorting similarly to how we do when we fetch startsWith and contains for fetchEntities. This wasn't applicable to table rows and was incorrectly re-ordering content during pagination.


baiirun:
missing commits for some reason. I blame github. Going to recreate and hopefully have all the commits.

PR #369, opened 2023-06-13
feat: compare versions and proposals

sandren:
This PR adds a new compare versions UI to entity pages and compare proposals UI to space pages.


apps/web/modules/components/compare.tsx
```diff
@@ -0,0 +1,698 @@
...
+        <div className="inline-flex items-center gap-4">Compare versions</div>
+        <div>
+          <Button variant="secondary" onClick={() => setIsCompareOpen(false)}>
+            Cancel
+          </Button>
+        </div>
+      </div>
+      <div className="mt-3 h-full overflow-y-auto overscroll-contain rounded-t-[32px] bg-bg shadow-big">
+        <div className="mx-auto max-w-[1200px] pt-10 pb-20 xl:pt-[40px] xl:pr-[2ch] xl:pb-[4ch] xl:pl-[2ch]">
+          {isLoading || typeof data !== 'object' ? (
```
baiirun, line 59: Usually if we have these types of checks it's a code smell. Can we check if it's `undefined` instead since that's what the actual type of `data` is when there is no data.

apps/web/modules/components/compare.tsx
```diff
@@ -0,0 +1,698 @@
...
+            <div className="text-metadataMedium">Loading...</div>
+          ) : (
+            <Versions
+              data={
+                data as [
```
baiirun, line 64: Same thing here, if we're casting it usually means that we aren't setting types correctly somewhere. `selected` and `previous` should probably shouldn't be `any`. Right now it's hiding any bugs that TypeScript normally would catch.

apps/web/modules/components/compare.tsx
```diff
@@ -0,0 +1,698 @@
...
+  while (markdownContent.startsWith('#')) {
+    markdownContent = markdownContent.substring(1);
+    markdownLevel++;
+  }
+
+  markdownType = markdownComponent[markdownLevel];
+  markdownContent = markdownContent.trim();
+
+  return { markdownType, markdownContent };
+}
```
baiirun, line 620: What's this code doing?

apps/web/modules/components/history/history-item.tsx
```diff
@@ -8,13 +8,14 @@ interface Props {
...
   // the same triple, we want to make sure that only renders as a single change.
+  onClick?: () => void;
   changeCount: number;
   createdAt: number;
   createdBy: Profile;
   name: string | null;
 }
 
-export function HistoryItem({ changeCount, createdAt, createdBy, name }: Props) {
+export function HistoryItem({ onClick = () => null, changeCount, createdAt, createdBy, name }: Props) {
```
baiirun, line 18: You can probably leave this as an undefined prop since `button` accepts `onClick` as undefined.

apps/web/modules/io/data-source/network.ts
```diff
@@ -256,7 +261,93 @@ export class Network implements INetwork {
...
+
+  fetchEntity = async (
+    id: string,
+    abortController?: AbortController | null,
+    blockNumber?: number
+  ): Promise<EntityType | null> => {
+    if (!id) return null;
+
+    if (typeof blockNumber === 'number' && blockNumber < 36472399) {
+      console.error(`36472399 is the earliest block number`);
```
baiirun, line 336: Probably want it to say "36472399 is the earliest block number _that we support_" or "earlier block number in Geo" or something like that.

apps/web/modules/components/compare.tsx
sandren: It parses the raw markdown into the element that should be rendered and the content to place within it. It's used when rendering diffs of markdown block content.

apps/web/modules/components/compare.tsx
sandren: By default data can also be of type `boolean` so I'll see if generics can help here.

apps/web/modules/io/data-source/network.ts
```diff
@@ -256,7 +261,93 @@ export class Network implements INetwork {
...
+              stringValue
+              valueType
+              valueId
+            }
+            entity {
+              id
+              name
+            }
+          }
+        }`,
```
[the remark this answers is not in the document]
sandren: Yep!

apps/web/modules/change/change.ts
```diff
@@ -402,3 +410,431 @@ const getEntities = async (actions: ActionType[], network: INetwork) => {
...
+  const selectedEntityBlockIds: string[] = selectedEntityBlockIdsTriple
+    ? JSON.parse(Value.stringValue(selectedEntityBlockIdsTriple) || '[]')
+    : [];
+
+  const previousEntityBlockIdsTriple = previousEntity?.triples.find(t => t.attributeId === SYSTEM_IDS.BLOCKS) ?? null;
+  const previousEntityBlockIds: string[] = previousEntityBlockIdsTriple
+    ? JSON.parse(Value.stringValue(previousEntityBlockIdsTriple) || '[]')
+    : [];
+
+  const [maybeRemoteSelectedEntityBlocks, maybeRemotePreviousEntityBlocks, maybeAdditionalRemotePreviousEntityBlocks] =
```
baiirun, line 447: What's an additional remote previous entity block?

apps/web/modules/components/compare.tsx
baiirun: Might be useful adding a comment as to why doing it this way is easier than using the Tiptap editor to render the content.

apps/web/modules/change/change.ts
```diff
@@ -402,3 +410,431 @@ const getEntities = async (actions: ActionType[], network: INetwork) => {
...
+  };
+
+  const selectedBlock = parseInt(selectedProposal.createdAtBlock, 10);
+  const previousBlock = selectedBlock - 1;
+
+  const entitySet = new Set<EntityId>();
+
+  if (selectedProposal) {
+    selectedProposal.proposedVersions.forEach((proposedVersion: any) => {
+      proposedVersion.actions.forEach((action: any) => {
```
baiirun, line 618: Do these need to be set to `any`?

apps/web/modules/io/data-source/network.ts
baiirun: Pinging this one again

apps/web/modules/change/change.ts
sandren: If we only check the current entity's blocks, we might miss blocks that have been deleted since the previous version. So we check the previous version's blocks as well.

apps/web/modules/io/data-source/network.ts
```diff
@@ -108,7 +108,13 @@ export interface INetwork {
   fetchTriples: (options: FetchTriplesOptions) => Promise<FetchTriplesResult>;
   fetchSpaces: () => Promise<Space[]>;
   fetchProfile: (address: string, abortController?: AbortController) => Promise<[string, Profile] | null>;
-  fetchEntity: (id: string, abortController?: AbortController) => Promise<EntityType | null>;
+  fetchProposedVersion: (id: string, abortController?: AbortController) => Promise<any | null>;
+  fetchProposal: (id: string, abortController?: AbortController) => Promise<any | null>;
```
baiirun, line 112: These should return the actual Proposal and ProposedVersion types as well.

PR #386, opened 2023-06-29
feat: date diffs

sandren:
This PR adds support for date diffs in review changes, compare versions, and compare proposals UI.

It also launches support for the new date type.

### DIFF
--- a/apps/web/modules/components/review.tsx
+++ b/apps/web/modules/components/review.tsx
@@ -750,11 +750,11 @@ export const DateTimeDiff = ({ mode, before, after }: DateTimeProps) => {
   let afterDateTime = null;
 
   if (before) {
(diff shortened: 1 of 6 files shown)


apps/web/modules/components/review.tsx
```diff
@@ -699,13 +702,121 @@ const ChangedAttribute = ({
...
+  hour: string;
+  minute: string;
+};
+
+export const DateTimeDiff = ({ mode, before, after }: DateTimeProps) => {
+  let beforeDateTime = null;
+  let afterDateTime = null;
+
+  if (before) {
+    beforeDateTime = GeoDate.fromISOStringUTC(before as string);
```
baiirun, line 753: Do you need to cast `before` and `after` here? It _should_ be coercing it to `string` already since you have the conditional above.

PR #394, opened 2023-07-05
feat: web url diffs

sandren:
This PR adds support for web url diffs in review changes, compare versions, and compare proposals UI.

It also launches support for the new web url type.


baiirun (commented):
Once we comment out the URL and Date entry points this should be good. Once we add the data migrations we can re-enable them.

baiirun (approved):

PR #397, opened 2023-07-07
feat: tiptap improvements

sandren:
This PR makes the following improvements to the TipTap editor:

- adds a "bullet list" type, which is stored as normal markdown content
- replaces hard breaks with normal paragraphs


apps/web/modules/entity/entity-store/entity-store.ts
```diff
@@ -445,7 +445,7 @@ export class EntityStore implements IEntityStore {
   We don't support changing types of blocks, so all we need to do is create a new block with the new type
   */
   createBlockTypeTriple = (node: JSONContent) => {
-    const blockEntityId = node.attrs?.id;
+    const blockEntityId = node.attrs?.id ?? node?.content?.[0]?.content?.[0]?.attrs?.id;
```
baiirun, line 448: Can we make some helper function that parses the ids from nodes? This is fairly brittle so seems easy to create a bug if we need to change this in several places.

baiirun (approved):

PR #398, opened 2023-07-07
fix: subspace redirect bug

sandren:
This PR fixes the browser back button behavior when navigating to a subspace.


apps/web/pages/space/[id]/[entityId].tsx
```diff
@@ -51,7 +51,7 @@ export default function EntityPage(props: Props) {
   }, []);
 
   if (props.redirect && isMounted) {
-    router.push(props.redirect);
+    router.replace(props.redirect);
```
Comment (baiirun) [inline apps/web/pages/space/[id]/[entityId].tsx:54]: 🚀

PR #443, opened 2023-08-16
fix/upgrade-@graphprotocol/graph-cli

jonathanprozzi:
This PR upgrades the `@graphprotocol/graph-cli` from `@0.33.1` to `@latest` 

This fixes the following error that occurred when running `pnpm install` -- " ERROR  Could not resolve feat/smaller to a commit of https://github.com/hugomrdias/concat-stream.git."

Upgrading this package to latest resolves this and allows for all packages to fully install.

PR #446, opened 2023-08-17
fix/upgrade-@graphprotocol/graph-cli -- with matching pnpm version

jonathanprozzi:
This PR upgrades the @graphprotocol/graph-cli from @0.33.1 to @latest -- 

I think the issue in #443 was that I was using a mismatched `pnpm` version. I started fresh with the correct version this time. Builds and passes lint locally, but unsure if will see the same error in the CI/CD.

---

This fixes the following error that occurred when running pnpm install -- " ERROR  Could not resolve feat/smaller to a commit of https://github.com/hugomrdias/concat-stream.git."

Upgrading this package to latest resolves this and allows for all packages to fully install.

### DIFF
--- a/packages/subgraph/package.json
+++ b/packages/subgraph/package.json
@@ -26,11 +26,11 @@
     "copy-subgraph-config": "docker cp $(docker ps -q --filter=\"name=devnet\"):/app/packages/subgraph/subgraph.yaml ."
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


jonathanprozzi:
@baiirun it looks like the `pnpm` version was the issue. Starting fresh with the correct version resolved it

baiirun (approved):

PR #490, opened 2023-09-13
feat: merge entities (GEO-333)

jonathanprozzi:
## Overview

This adds in ability for users to merge entities. A user chooses a second entity from the entity action menu, and then chooses which properties and ID to merge with.

A preview of the new entity is shown, and then the 'publish' flow kicks off when the user is ready:

- Split out the merged and unmerged triples
- Delete the unmerged triples
- Create the new triples on the selected entity ID
- Run the migrate on the unselected entity ID

Added package:
- `@radix-ui/react-tabs'`

## Questions/Considerations

- I made a custom checkbox of an `input` with type `checkbox` to match the design and handle the selection logic
  - Can pull into the design system if needed or we can consider using the Radix Checkbox
  - I noticed we used a button previously, but I thought a checkbox may work better due to the dependent selection logic
- For pulling in the triples for the second entity:
  - This fetches the triples for the second entity ID from the subgraph and populates
  - @baiirun I wasn't sure if there was a way to also do this locally from the store. For the first set of triples, I pull those from the Entity store since this flow is triggered from within the first entity's page
- The delete deletes the unmerged (unselected) triples -- when I tested, this was showing in changes as expected
## Screencaps
merge-entity-demo-1

merge-entity-demo-2


apps/web/partials/merge-entity/merge-entity-menu.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export function MergeEntityMenu({ entityId }: Props) {
+  const { setIsMergeReviewOpen, setEntityIdOne, setEntityIdTwo } = useMergeEntity();
+
+  const autocomplete = useAutocomplete();
+  const { spaces } = useSpaces();
+  const { address } = useAccount();
+
+  // would we also want to potentially filter out entities that are in spaces where the user doesn't have editor permissions?
+  const spacesUserIsEditor = spaces.filter(space => space.editors.includes(address ?? '')).map(space => space.id);
```
baiirun, line 27: Good question to ask in #geo-team

baiirun: If we _don't_ then we might want to change the move entity to have the same behavior.

apps/web/partials/merge-entity/merge-entity-review-page.tsx
```diff
@@ -0,0 +1,190 @@
...
+                xmlns="http://www.w3.org/2000/svg"
+                viewBox="0 0 24 24"
+                fill="none"
+                stroke="currentColor"
+                strokeWidth="4"
+                strokeLinecap="round"
+                strokeLinejoin="round"
+              >
+                <polyline points="20 6 9 17 4 12"></polyline>
+              </svg>
```
baiirun, line 71: What is this SVG?

apps/web/partials/merge-entity/merge-entity-review-page.tsx
```diff
@@ -0,0 +1,190 @@
...
+            <LinkableChip href={NavUtils.toEntity(triple.space, triple.value.id)}>
+              {triple.value.name || triple.value.id}
+            </LinkableChip>
+          </div>
+        );
+      }
+      case 'number':
+        return null;
+    }
+  };
```
baiirun, line 134: We should probably abstract this into a design-system component at this point since it's used in so many places now. Let's make another PR after this one that does that.

apps/web/partials/merge-entity/merge-entity-review-page.tsx
```diff
@@ -0,0 +1,190 @@
...
+                    xmlns="http://www.w3.org/2000/svg"
+                    viewBox="0 0 24 24"
+                    fill="none"
+                    stroke="currentColor"
+                    strokeWidth="4"
+                    strokeLinecap="round"
+                    strokeLinejoin="round"
+                  >
+                    <polyline points="20 6 9 17 4 12"></polyline>
+                  </svg>
```
baiirun, line 180: What's this SVG?

apps/web/partials/merge-entity/merge-entity-review.tsx
```diff
@@ -0,0 +1,344 @@
...
+      queryKey: ['entity-merge-review', entityIdTwo],
+      queryFn: async () => {
+        if (!entityId) return null;
+        return await subgraph.fetchEntity({ endpoint: config.subgraph, id: entityId });
+      },
+    });
+    if (!entityTwoData || isLoading || error) {
+      return [];
+    }
+    return [entityTwoData].flatMap((entity => entity?.triples) ?? []);
```
baiirun, line 63: You can just do `entity?.triples ?? []`

apps/web/partials/merge-entity/merge-entity-review.tsx
```diff
@@ -0,0 +1,344 @@
...
+    </SlideUp>
+  );
+}
+
+function MergeEntityReviewChanges() {
+  const { setIsMergeReviewOpen, entityIdOne, entityIdTwo } = useMergeEntity();
+  const hub = useMigrateHub();
+
+  const { subgraph, config } = Services.useServices();
+  function useEntityById(entityId: string) {
```
baiirun, line 48: We should move this outside of `MergeEntityReviewChanges` since it's a fairly expensive allocation of a function.

apps/web/partials/merge-entity/merge-entity-review.tsx
```diff
@@ -0,0 +1,344 @@
...
+        type: 'DELETE_ENTITY',
+        payload: {
+          entityId: notMergedEntityId,
+        },
+      });
+    };
+
+    onDelete();
+    onCreate();
+    onMigrate();
```
baiirun, line 127: Can we just run the functionality directly here instead of abstracting into functions? Abstracting into functions doesn't do much to make the code easier to understand and just creates indirectness and additional function allocations.

```tsx
batch(() => {
  unmergedTriples.forEach(t => remove(t));
  mergedTriples.forEach(t => {
    create(
    Triple.withId({
       ...t,
       entityId: mergedEntityId,
     })
    );
  });
});

hub.migrate({
  type: 'DELETE_ENTITY',
  payload: {
    entityId: notMergedEntityId,
  },
});

```

apps/web/partials/merge-entity/merge-entity-review.tsx
```diff
@@ -0,0 +1,344 @@
...
+    create,
+    hub,
+  ]);
+
+  function handleCheckboxSelect({
+    attributeId,
+    selectedTriple,
+  }: {
+    attributeId: string;
+    selectedTriple: TripleType[];
```
baiirun, line 147: selectedTriple**s** with an `s`

apps/web/design-system/button.tsx
```diff
@@ -66,7 +77,7 @@ export const Button = forwardRef(function Button(
       disabled={disabled}
       {...rest}
     >
-      {icon && <Icon icon={icon} color={iconColor} />}
+      {icon && <Icon icon={icon} color={finalIconColor} />}
```
baiirun, line 80: Where are we using a Button that requires a special icon color?

apps/web/partials/merge-entity/merge-entity-review-page.tsx
jonathanprozzi: It's an svg for the actual custom checkbox style

apps/web/partials/merge-entity/merge-entity-review-page.tsx
jonathanprozzi: Sounds good!

apps/web/design-system/button.tsx
jonathanprozzi: The back button in the design - it was a secondary variant with a grey-04 color

apps/web/partials/merge-entity/merge-entity-review.tsx
jonathanprozzi: Good idea!

apps/web/partials/merge-entity/merge-entity-review.tsx
jonathanprozzi: Ah, that's much more concise.

apps/web/partials/merge-entity/merge-entity-menu.tsx
jonathanprozzi: Made the change to the _Move Entity_ menu to remove the filter criteria of the user being an editor and updated the empty state message to not be as specific to the user's role

Made the change to remove the filter for spaces where the user is the editor as well. Wondering if we'd want to sort via spaces where user is an editor (show at the top)

baiirun: I think for now it's fine to assume they'll search for the space they're interested in. Grouping by access/no access can be a future enhancement.

apps/web/design-system/button.tsx
baiirun: A better API might be to allow passing a custom Icon instead of adding another control prop. Might look something like this.

```tsx
interface Props {
  icon?: IconName | React.ReactNode
}

function Button({ icon, ...rest }) {
   const Icon = typeof icon === 'string' ? <Icon icon={icon} color={iconColor} /> : typeof icon === 'function' ? icon : null
   
   return (
     {icon}
     // ...
   )
}

function Component() {
  return (
    <div>
      <Button varian="secondary" icon={<LeftArrow color="ctaPrimary />
      <Button variant="secondary" icon="cog" />
    </div>  
}
```

baiirun: Almost might be worth double checking if the button design is intended to replace the current secondary design. I think we're only using the secondary button in like one place in the app right now.

jonathanprozzi: Sounds good! Commented in the Figma -- if that's the case we can update the ternary to use grey instead of the ctaPrimary value

jonathanprozzi: Being able to pass in the Icon as a string or icon component makes sense. If we end up not changing the secondary to use grey over ctaPrimary I'll update to match this component API. 

Another benefit of changing this to take in ReactNodes is that we won't need to import icons into the Icon component -- we can import them directly into the component and pass in like in your second example:
```     <Button variant="secondary" icon={<LeftArrow color="ctaPrimary" />} />```

instead of needing to add a new icon svg to the Icon component so that it's found in the IconName props

File: apps/web/partials/move-entity/move-entity-menu.tsx [hunk 10]
```diff
@@ -84,10 +79,8 @@ export function MoveEntityMenu({ entityId, spaceId }: Props) {
 
       {filteredSpacesForMoveResults.length === 0 && (
         <div className="flex flex-col gap-1 p-2 pt-0">
-          <Text variant="footnoteMedium">You don’t have editor access in any other spaces.</Text>
-          <Text variant="footnote">
-            You will need to become an editor of the space you want to move the entities to.
-          </Text>
+          {/* <Text variant="footnoteMedium">You don’t have editor access in any other spaces.</Text> */}
+          <Text variant="footnoteMedium">No spaces found for your search results.</Text>
```
baiirun, line 83: Do we still need this commented out component?

apps/web/partials/merge-entity/merge-entity-review-page.tsx
```diff
@@ -0,0 +1,190 @@
...
```
baiirun, line 180: Should this be using the new CheckCircleReview icon?

apps/web/partials/merge-entity/merge-entity-review-page.tsx
```diff
@@ -0,0 +1,190 @@
...
+                xmlns="http://www.w3.org/2000/svg"
+                viewBox="0 0 24 24"
+                fill="none"
+                stroke="currentColor"
+                strokeWidth="4"
+                strokeLinecap="round"
+                strokeLinejoin="round"
+              >
+                <polyline points="20 6 9 17 4 12"></polyline>
+              </svg>
```
baiirun, line 71: Should this be using the new CheckCircleReview icon?

apps/web/partials/merge-entity/merge-entity-review-page.tsx
jonathanprozzi: These are different -- that one has a circle around it for the steps (ie Delete triples) whereas this is the checkbox rendering inside of the checkbox itself for the custom checkbox design.

apps/web/partials/merge-entity/merge-entity-review-page.tsx
jonathanprozzi: Same as other checkbox svg

jonathanprozzi:
I went to make some screencaps showing the flows and it looks like something has changed -- I went back to a commit prior to the caching changes and the flow works, so looking into what is different and trying to isolate it.

The local actions taken for the create, delete, and migrate are no longer showing up for me and the review panel no longer opens since there aren't changes. Not sure what could cause that though, and it worked at the end of last week. Going to dig in a bit more @baiirun

baiirun:
> I went to make some screencaps showing the flows and it looks like something has changed -- I went back to a commit prior to the caching changes and the flow works, so looking into what is different and trying to isolate it.
> 
> The local actions taken for the create, delete, and migrate are no longer showing up for me and the review panel no longer opens since there aren't changes. Not sure what could cause that though, and it worked at the end of last week. Going to dig in a bit more @baiirun

The caching should be removed now if you are on latest master.

apps/web/partials/merge-entity/merge-entity-review.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+      batch(() => {
+        unmergedTriples.forEach(t => remove(t)); // delete the triples that aren't merged
+        mergedTriples.forEach(t => {
+          create(
+            Triple.withId({
+              ...t,
+              entityId: mergedEntityId,
+            })
+          );
```
baiirun, line 101: We also need to make sure that the merged triples are being put in the correct space

jonathanprozzi: The correct space would likely be the space for the second entity then? Or would we want to allow someone to select the space?

Entity A - space 1
Entity B - space 2

Would a user always want the desired space to be space 2? Thinking someone may choose Entity A in space 1 and choose properties from Entity B, but intend for it to stay in space 1

jonathanprozzi: Made the update to use Entity B's space and that's working. Still troubleshooting the one case where actions are being wiped out (when there is nothing to migrate), but the other cases work.

jonathanprozzi:
@baiirun 

**Merging an entity into a new space**
merge-entity-new-space

The above screencap shows the user flow for when a user chooses to merge Entity A with Entity B, which is in a different space. When this is the case, the triples created during the merge are specifically created in Entity B's space. There's no data to migrate in this example.

**Merging an entity with a migration from a reference**
merge-entity-with-migration

This screencap shows the user flow for when a user merges entities that requires a migration. In this example, Entity B (Test entity 2) is used referentially in Test Entity 3, so the migration occurs.

**Merging an entity without a migration**
merge-entity-no-migration

This screencap shows the user flow where a user is merging Entity A and Entity B and they're both in the same space, and where no migration is needed.

baiirun (approved):

PR #507, opened 2023-09-21
fix: add sorting to fetch entities query

sandren:
This PR adds automatic sorting by name to the fetch entities query.


jonathanprozzi:
Tested locally and works for me

That's a massive UX improvement for folks!

baiirun (approved):

PR #520, opened 2023-10-16
fix: persistence

sandren:
This PR revises the IndexedDB persistence implementation to only save unpublished actions.

### DIFF
--- a/apps/web/core/state/actions-store/actions-store.ts
+++ b/apps/web/core/state/actions-store/actions-store.ts
@@ -61,7 +61,24 @@ export class ActionsStore implements IActionsStore {
     // https://legendapp.com/open-source/state/persistence/#persistobservable
     if (typeof window !== 'undefined') {
       persistObservable(actions, {
-        local: 'actionsStore',
+        local: {
+          name: 'actionsStore',
+          adjustData: {
+            save: (data: SpaceActions) => {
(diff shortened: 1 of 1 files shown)


apps/web/core/state/actions-store/actions-store.ts
```diff
@@ -61,7 +61,22 @@ export class ActionsStore implements IActionsStore {
...
+            save: (data: SpaceActions) => {
+              const savedData: SpaceActions = {};
+              const spaces = Object.keys(data);
+              spaces.forEach(spaceId => {
+                const persistedActions: ActionType[] = data[spaceId].filter(action => !action.hasBeenPublished);
+                if (persistedActions.length > 0) {
+                  savedData[spaceId] = persistedActions;
+                }
+              });
+              return savedData;
```
baiirun, line 76: It's a minor optimization, but you can use `Object.entries` and access the underlying values directly instead of referecing `data[spaceId` again.

```ts
const savedData: SpaceActions = {};

Object.entries(data).forEach(([spaceId, actions]) => {
  const persistedActions = actions.filter(action => !action.hasBeenPublished);

  if (persistedActions.length > 0) {
    savedData[spaceId] = persistedActions;
  }
});

return savedData;
```

baiirun (approved):
One small optimization comment but looks good

PR #521, opened 2023-10-17
feat: onboarding with backend

baiirun:
This PR integrates (jankily) the e2e workflow for onboarding a new Geo user on Polygon mainnet.
- Deploys personal space
- Configures space with correct roles for the connected user
- Creates profile entity in the new space
- Registers the user in the on-chain `GeoProfileRegistry`
- Integrates network calls in app to attempt querying both the permissionless and permissioned spaces and sets up context to query the correct space on the client depending on the current space.
- Fixes some bugs during the onboarding API workflow

### DIFF
--- a/apps/web/app/space/[id]/[entityId]/layout.tsx
+++ b/apps/web/app/space/[id]/[entityId]/layout.tsx
@@ -37,6 +37,18 @@ export default async function ProfileLayout({ children, params }: Props) {
   // Layouts do not receive search params (hmm)
   const config = Params.getConfigFromParams({}, env);
 
+  let space = await Subgraph.fetchSpace({ endpoint: config.subgraph, id: params.id });
+  let usePermissionlessSubgraph = false;
+
+  if (!space) {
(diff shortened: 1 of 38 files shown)


apps/web/app/space/[id]/[entityId]/layout.tsx
```diff
@@ -37,6 +37,18 @@ export default async function ProfileLayout({ children, params }: Props) {
...
+  let usePermissionlessSubgraph = false;
+
+  if (!space) {
+    space = await Subgraph.fetchSpace({ endpoint: config.permissionlessSubgraph, id: params.id });
+    if (space) usePermissionlessSubgraph = true;
+  }
+
+  if (usePermissionlessSubgraph) {
+    config.subgraph = config.permissionlessSubgraph;
+  }
```
Comment (baiirun) [inline apps/web/app/space/[id]/[entityId]/layout.tsx:50]: This is a janky approach for dynamically querying either the permissioned or permissionless subgraph. Right now we don't know which subgraph has the data for the space we are attempting to view, so we have to try both.

In the future ideally there's some other API (or a gateway) that tells us where the data for the space lives.

PR #533, opened 2023-10-18
feat: performance profiling and improvements

baiirun:


apps/web/core/io/api/space.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { fetcher } from './geo-fetch';
+
+interface NetworkData {
+  space: Space | null;
+  isPermissionlessSpace: boolean;
+}
+
+export async function space(spaceId: string): Promise<NetworkData> {
+  return fetcher<NetworkData>(`/api/space/${spaceId}`);
```
baiirun, line 11: This will eventually have error handling similar to our methods that call `graphql`

PR #570, opened 2023-11-07
feat: table block store refactor

baiirun:
This PR refactors the table block store away from the class-based legendstate approach. This previous approach caused a few problems:
1. All state transitions were asynchronous, so it was easy to get in strange states as a result of race conditions.
2. It was difficult to reason about as we were deriving data from observables all over the file. You would have to constantly context switch across many lines to understand the full picture of the data flow in the table.
3. There was lots of overhead and boilerplate initializing the store. We would pass it from inside of React to outside of React, just to bind back to React again at the end.
4. Caching was not working correctly with our query client inside of async computed observables.

All of the logic in the previous `TableBlockStore` now lives as a simple React hook. It's easier to reason about as you just go top-to-bottom to see the flow of data and how state is dependent on each other.


apps/web/core/merged/merged.ts
```diff
@@ -156,13 +156,13 @@ export class Merged implements IMergedDataSource {
...
-        actions => Entity.mergeActionsWithEntities(actions, maybeNetworkEntity ? [maybeNetworkEntity] : []),
-        A.find(e => e.id === options.id)
-      );
+      // If not networkEntity we need to just return the local entity
+      if (!maybeNetworkEntity) {
+        return Entity.fromActions(this.store.allActions$.get(), options.id);
+      }
+
+      // If the network entity exists, we need to merge the local actions with the network entity.
+      const entity = Entity.mergeActionsWithEntity(this.store.allActions$.get(), maybeNetworkEntity);
```
baiirun, line 165: This fixes a bug where merging an entity with any local state would not merge in the remote triples.

PR #578, opened 2023-11-15
feat: onboarding refinement

sandren:
This PR adds a more robust onboarding experience.

https://github.com/geobrowser/geogenesis/assets/20441876/3b8f1052-99a1-4f06-8692-d549d7122cbf

https://github.com/geobrowser/geogenesis/assets/20441876/6ec7388a-6d59-4fc1-a941-337563c9df1e


baiirun:
I forced the transaction signature to fail (by rejecting it), and then clicked Retry and Accepted and it went straight to the "done" step instead of the 3rd step, which means none of the data was created in their personal space and none of the roles were set up.

baiirun (approved):

PR #603, opened 2023-12-08
feat: support multiple filters on new entity

sandren:
This PR revises the new entity mechanism to support an arbitrary number of filters.

### DIFF
--- a/apps/web/partials/blocks/table/table-block.tsx
+++ b/apps/web/partials/blocks/table/table-block.tsx
@@ -99,7 +99,7 @@ export const TableBlock = React.memo(({ spaceId }: Props) => {
   });
 
   const typeId = type.entityId;
-  const filters =
+  const filters: Array<[string, string]> =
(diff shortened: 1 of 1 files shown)


apps/web/partials/blocks/table/table-block.tsx
```diff
@@ -98,8 +99,8 @@ export const TableBlock = React.memo(({ spaceId }: Props) => {
   });
 
   const typeId = type.entityId;
-  const filterId = filterState?.[0]?.columnId ?? null;
-  const filterValue = filterState?.[0]?.value ?? null;
+  const filters =
```
baiirun, line 102: If you type the `filters` variable as `Array<[string, string]>` you shouldn't have to cast on line `163`. You might need to do `as const` or `as Array<[string, string]>` in the left side of the ternary but it's better to do the casting as close to the initialization as possible.

baiirun (approved):

PR #607, opened 2023-12-12
feat: create new attribute

sandren:
This PR adds the missing "create new attribute" functionality.

localhost_3000_space_0x170b749413328ac9a94762031a7A05b00c1D2e34(HD)

### DIFF
--- a/apps/web/partials/blocks/table/table-block-context-menu.tsx
+++ b/apps/web/partials/blocks/table/table-block-context-menu.tsx
@@ -343,8 +343,7 @@ function AddAttribute() {
   const [, setToast] = useToast();
   const { create } = useActionsStore();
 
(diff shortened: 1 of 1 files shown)


apps/web/partials/blocks/table/table-block-context-menu.tsx
```diff
@@ -338,7 +340,11 @@ const resultsListActionBarStyles = cva(
 );
 
 function AddAttribute() {
+  const [, setToast] = useToast();
+  const { create } = useActionsStore();
+
   const { type } = useTableBlock();
+  const spaceId = type.space;
```
baiirun, line 347: Not sure if the new entity should be created in the same space as the type or the current space the user is in. They could be different spaces. I'd bring this up with Yaniv.

PR #626, opened 2024-01-17
fix: attributes don't auto populate for system entities

baiirun:


baiirun:
Closing to create a new branch without a quote in the name

PR #628, opened 2024-01-17
feat: teams

sandren:


apps/web/app/space/[id]/layout.tsx
```diff
@@ -38,7 +39,26 @@ interface TabProps {
...
+  if (typeIds.includes(SYSTEM_IDS.COMPANY_TYPE)) {
+    const roleTriples = await Subgraph.fetchTriples({
+      space: params.id,
+      query: '',
+      skip: 0,
+      first: 1000,
+      filter: [{ field: 'attribute-id', value: ROLE_ATTRIBUTE }],
+    });
+
+    if (roleTriples) {
```
Comment (baiirun) [inline apps/web/app/space/[id]/layout.tsx:57]: Probably should check length > 0 since it's safer than checking an empty array reference truthiness.

apps/web/app/space/[id]/team/page.tsx
```diff
@@ -1,7 +1,137 @@
...
+    Subgraph.fetchTriples({
+      space: spaceId,
+      query: '',
+      skip: 0,
+      first: 1000,
+      filter: [{ field: 'attribute-id', value: AVATAR_ATTRIBUTE }],
+    }),
+  ]);
+
+  if (!roleTriples) {
```
Comment (baiirun) [inline apps/web/app/space/[id]/team/page.tsx:52]: Same thing here

apps/web/app/space/[id]/team/page.tsx
```diff
@@ -1,7 +1,137 @@
...
+    teamMembers.map(teamMember => {
+      return Subgraph.fetchEntity({ id: teamMember.entityId });
+    })
   );
+
+  entities.forEach(entity => {
+    if (!entity) return;
+
+    const entityId = entity.id;
+    const entitySpaceId = entity.nameTripleSpace;
```
Comment (baiirun) [inline apps/web/app/space/[id]/team/page.tsx:81]: There might be multiple name triples if we've created new triples for the Person in the company space. Which `entitySpaceId` _should_ we be using here in that case?

apps/web/app/space/[id]/team/page.tsx
```diff
@@ -1,7 +1,137 @@
...
+    const teamMemberIndex = teamMembers.findIndex(teamMember => teamMember.entityId === entityId);
+
+    const isLinked = entitySpaceId && spaceId !== entitySpaceId;
+
+    if (isLinked) {
+      teamMembers[teamMemberIndex].space = entitySpaceId;
+      teamMembers[teamMemberIndex].linked = true;
+
+      const nameTriple = nameTriples.find(nameTriple => nameTriple.entityId === entityId);
```
Comment (baiirun) [inline apps/web/app/space/[id]/team/page.tsx:90]: Same thing here, which name will we get if there are multiple name triples?

apps/web/app/space/[id]/team/page.tsx
```diff
@@ -1,7 +1,137 @@
...
+  entities.forEach(entity => {
+    if (!entity) return;
+
+    const entityId = entity.id;
+    const entitySpaceId = entity.nameTripleSpace;
+    const teamMemberIndex = teamMembers.findIndex(teamMember => teamMember.entityId === entityId);
+
+    const isLinked = entitySpaceId && spaceId !== entitySpaceId;
+
+    if (isLinked) {
```
Comment (baiirun) [inline apps/web/app/space/[id]/team/page.tsx:86]: All this code for setting team member metadata should probably be in a separate function that we can test since it's somewhat complex and has lots of branches.

apps/web/app/space/[id]/team/page.tsx
```diff
@@ -1,7 +1,137 @@
...
+        }
+      } else {
+        const name = Entity.name(entity.triples);
+
+        if (name) {
+          teamMembers[teamMemberIndex].name = name;
+        }
+      }
+
+      const avatarTriple = avatarTriples.find(avatarTriple => avatarTriple.entityId === entityId);
```
Comment (baiirun) [inline apps/web/app/space/[id]/team/page.tsx:106]: Same question as the name triple since there might be multiple avatar triples if we create unique ones in the company space.

apps/web/core/utils/value/value.ts
```diff
@@ -37,7 +37,9 @@ export function imageValue(triple: Triple) {
 // format for storing in the triple
 // e.g., https://api.thegraph.com/ipfs/api/v0/cat?arg=HASH -> ipfs://HASH
 export function toImageValue(rawValue: string) {
-  if (rawValue.includes('ipfs') && rawValue.includes('?arg=')) {
+  if (rawValue.startsWith('ipfs://') && rawValue.length === 53) {
```
baiirun, line 40: What does this do?

apps/web/design-system/autocomplete/entity-text-autocomplete.tsx
```diff
@@ -26,7 +27,7 @@ import { ResultContent, ResultsList } from './results-list';
 
 interface Props {
   placeholder?: string;
-  onDone: (result: { id: string; name: string | null }) => void;
+  onDone: (result: { id: string; name: string | null } | EntityType) => void;
```
baiirun, line 30: You can probably just remove the `{ id: string; name: string | null }` type since it's just a subset of `EntityType`. I'm assuming we're using some of the other properties of `EntityType` elsewhere when reading the result of `onDone`?

apps/web/design-system/chip.tsx
```diff
@@ -39,15 +39,18 @@ export function LinkableChip({ href, children }: LinkableChipProps) {
 }
 
 interface ChipButtonProps {
-  onClick: () => void;
+  onClick?: () => void;
   children: React.ReactNode;
-  href: string;
+  href?: string;
```
baiirun, line 44: We should probably verify that the Role chips that we show aren't meant to be clickable/linkable. My guess is that Yaniv would expect them to be clickable like the other entity chips

apps/web/partials/team/find-team-member.tsx
```diff
@@ -0,0 +1,387 @@
...
+    setHasFoundPerson(false);
+    setStep('start');
+  };
+
+  const handleOnChangeEntityId = useCallback(
+    async ({ currentTarget: { value } }: ChangeEvent<HTMLInputElement>) => {
+      if (error) return;
+
+      const entityIdLength = value.trim().length;
+      const isValidEntityId = VALID_ENTITY_ID_LENGTHS.includes(entityIdLength);
```
baiirun, line 160: Why are we checking for valid entity id lengths?

apps/web/partials/team/find-team-member.tsx
```diff
@@ -0,0 +1,387 @@
...
+        setEntityId(entityId);
+
+        const person = await Subgraph.fetchEntity({
+          id: entityId,
+        });
+
+        if (person) {
+          const types = await fetchEntityType({
+            id: entityId,
+          });
```
baiirun, line 175: We can probably parallelize these queries

apps/web/core/utils/value/value.ts
sandren: This check ensures that we return the original value if the value is already a valid image value. Otherwise if `toImageValue` is ever inadvertently used twice in the chain of data processing, it will return an empty string.

apps/web/partials/team/find-team-member.tsx
sandren: This check lets us automatically proceed to link the person when the user pastes a valid entity id. Otherwise we'd have to add a button to the design (or expect that they press `enter`).

apps/web/app/space/[id]/team/page.tsx
```diff
@@ -1,7 +1,145 @@
...
+      filter: [{ field: 'attribute-id', value: NAME }],
+    }),
+    Subgraph.fetchTriples({
+      space: spaceId,
+      query: '',
+      skip: 0,
+      first: 1000,
+      filter: [{ field: 'attribute-id', value: AVATAR_ATTRIBUTE }],
+    }),
+  ]);
```
Comment (baiirun) [inline apps/web/app/space/[id]/team/page.tsx:57]: This is fetching all roles, names, and avatars in the entire _space_? Wonder if there's a way to only get the triples that are part of a specific type in the space.

You also don't need the `first: 1000` with the substream api since it just returns all of the triples.

apps/web/app/space/[id]/team/page.tsx
```diff
@@ -1,7 +1,145 @@
...
+    };
+
+    teamMembers.push(teamMember);
+  });
+
+  const entities = await Promise.all(
+    teamMembers.map(teamMember => {
+      return Subgraph.fetchEntity({ id: teamMember.entityId });
+    })
+  );
```
Comment (baiirun) [inline apps/web/app/space/[id]/team/page.tsx:83]: You can probably write a query that just fetches all the entities that have type role in the space instead of fetching the triples then fetching the entities.

Can try something liek this 

```tsx
Subgraph.fetchEntities({
    spaceId: 'space-id'
    filter: [{
      field: 'attribute-id',
      value: 'role-id'
     }]
  })
```

Comment (baiirun) [reply apps/web/app/space/[id]/team/page.tsx:83]: This _should_ give us all of the triples for an entity that has _some_ triples that match the filter and space.

apps/web/core/utils/value/value.ts
baiirun: What makes an image value "valid"? Just the length?

apps/web/app/space/[id]/team/page.tsx
Comment (sandren) [reply apps/web/app/space/[id]/team/page.tsx:83]: That would work for unlinked members, but linked entities are in a separate space. The actual `spaceId` is only known once we compare `entity.nameTripleSpace` after the entities are fetched.

baiirun (approved):

PR #664, opened 2024-03-06
feat: add nonprofit template

sandren:
This PR adds a starter template for nonprofit spaces.


apps/web/app/api/nonprofit/deploy/make-nonprofit-effect.ts
```diff
@@ -195,6 +195,954 @@ export async function makeNonprofitEffect(
         ...spaceTypeTriple,
       });
 
+      // Add Page foreign type
+      const pageForeignTypeTriple: OmitStrict<Triple, 'id'> = {
```
baiirun, line 199: Are the changes in both files exactly the same?

If so we should probably abstract out into a separate function since these are quite a lot of things happening at once. We should be able to test them easily with vitest snapshots.

sandren: Not quite, since they use different variables for `entityId` and `entityName`, but we could pass them as arguments to `generateTriplesForNonprofit`. 🙂

baiirun: Yeah that sounds good and should be a useful abstraction

baiirun (approved):

PR #675, opened 2024-04-09
feat: table config

sandren:
This PR adds the ability to show/hide columns from the table block context menu.

### DIFF
--- a/apps/web/partials/blocks/table/table-block-context-menu.tsx
+++ b/apps/web/partials/blocks/table/table-block-context-menu.tsx
@@ -1,8 +1,10 @@
 'use client';
 
 import { SYSTEM_IDS } from '@geogenesis/ids';
+import * as Dropdown from '@radix-ui/react-dropdown-menu';
 import { useQuery } from '@tanstack/react-query';
 import { cva } from 'class-variance-authority';
+import cx from 'classnames';
 import { AnimatePresence, motion } from 'framer-motion';
 import { atom, useAtom } from 'jotai';
 import Image from 'next/legacy/image';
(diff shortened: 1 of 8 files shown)


apps/web/partials/blocks/table/table-block-context-menu.tsx
```diff
@@ -229,19 +237,38 @@ function useOptimisticAttributes({
...
+  const { type, spaceId, entityId, name } = useTableBlock();
+  const [isEditingColumns, setIsEditingColumns] = useAtom(editingColumnsAtom);
 
   const isEditing = useUserIsEditing(spaceId);
 
+  useEffect(() => {
+    if (!isEditing) {
+      setIsEditingColumns(false);
+    }
+  }, [isEditing, setIsEditingColumns]);
```
baiirun, line 262: Can't this just be derived during render? Don't think we need an effect to do this.

apps/web/partials/blocks/table/table-block-context-menu.tsx
```diff
@@ -267,56 +289,170 @@ export function TableBlockContextMenu() {
...
+                <button
+                  onClick={() => setIsEditingColumns(false)}
+                  className="flex w-full items-center gap-2 p-2 text-smallButton"
+                >
+                  <LeftArrowLong />
+                  <span>Back</span>
+                </button>
+              </MenuItem>
+              {allColumns.map((column: Column, index: number) => {
+                if (index === 0) return null;
```
baiirun, line 369: Is this specifically for the name column? Might be worth adding a comment for that.

apps/web/partials/blocks/table/table-block-context-menu.tsx
```diff
@@ -267,56 +289,170 @@ export function TableBlockContextMenu() {
...
+  entityId,
+  entityName,
+}: ToggleColumnProps) => {
+  const { create, remove } = useActionsStore(space);
+
+  const { id, name } = column;
+  const isShown = shownIndexes.includes(index);
+
+  const onToggleColumn = useCallback(async () => {
+    const attributeId = '388ad59b-1cc7-413c-a0bb-34a4de48c758';
```
baiirun, line 418: We should add this to SYSTEM_IDS or consume it from there if it already exists

baiirun (approved):

PR #680, opened 2024-04-16
fix: remove duplicate editable fields

sandren:


baiirun (approved):
Should put a comment as to why we're only passing the first triple instead of mapping over triples, but otherwise looks good

PR #717, opened 2024-05-30
feat: templates

sandren:
This PR adds support for space templates.


apps/web/core/utils/contracts/generate-actions-for-company.ts
```diff
@@ -0,0 +1,63 @@
+import { SYSTEM_IDS } from '@geogenesis/ids';
+
+import { cloneEntity } from './clone-entity';
+
+export const generateActionsForCompany = async (
+  spaceConfigEntityId: string,
+  spaceName: string,
```
baiirun, line 7: This should probably be nullable so we don't create the triple if the username is empty. Same for the other space types.

apps/web/core/utils/contracts/generate-actions-for-company.ts
```diff
@@ -0,0 +1,63 @@
...
+    postsPageActions,
+    // productsPageActions,
+    // servicesPageActions,
+    eventsPageActions,
+    jobsPageActions,
+  ] = await Promise.all([
+    cloneEntity({
+      oldEntityId: SYSTEM_IDS.COMPANY_SPACE_CONFIGURATION_TEMPLATE,
+      entityId: spaceConfigEntityId,
+      entityName: spaceName ?? '',
```
baiirun, line 23: See above comment

apps/web/app/space/[id]/spaces/page.tsx
```diff
@@ -0,0 +1,13 @@
...
+  const spaces = await getSpaces();
+
+  return <Spaces spaces={spaces} />;
+}
+
+const getSpaces = async () => {
+  const spaces: Array<any> = [];
+
+  return spaces;
+};
```
Comment (baiirun) [inline apps/web/app/space/[id]/spaces/page.tsx:13]: What is this file doing? Do we have a "Spaces" tab in one of the designs?

apps/web/partials/spaces/spaces.tsx
```diff
@@ -0,0 +1,16 @@
...
+export const Spaces = ({ spaces = [] }: SpacesProps) => {
+  // @TODO remove console.info for spaces
+  console.info('spaces:', spaces);
+
+  return (
+    <div>
+      <div>spaces</div>
+    </div>
+  );
+};
```
baiirun, line 16: What's this file?

packages/ids/system-ids.ts
```diff
@@ -248,5 +263,45 @@ export const MEMBERSHIP_CONTRACT_ADDRESS = '0x34a94160f4B0f86d932927DFfb326354dB
...
+export const NONPROFIT_FINANCES_PAGE_TEMPLATE =
+  '3be01e21-8227-42e0-bd40-868957e3ede2'
+
+export const PERSON_SPACE_CONFIGURATION_TEMPLATE =
+  '25d4b5bb-2f3a-4854-a9fe-df2f5f12b5e4'
+export const PERSON_POSTS_PAGE_TEMPLATE = '026362d4-5d41-4b8d-b6ef-8ed10ecd0d89'
+
 // Legacy attributes
-export const HIDDEN_COLUMNS = '1ed5b976-f8a4-451b-a2cc-531f85b59cab';
+export const HIDDEN_COLUMNS = '1ed5b976-f8a4-451b-a2cc-531f85b59cab'
```
baiirun, line 307: Can you add a comment here that we'll need to migrate all of these types to the new data model? The new data model has a new id scheme where we remove the dashes.

Also the ids are moved to a new `sdk` package and the `ids` package is removed.

Something like `@TODO(migration)`

apps/web/partials/tab/tab.tsx
```diff
@@ -0,0 +1,73 @@
...
+const getEntityId = async (spaceId: string, pageTypeId: string) => {
+  const pageTypeTriples = await Subgraph.fetchTriples({
+    space: spaceId,
+    query: '',
+    skip: 0,
+    first: 1000,
+    filter: [{ field: 'attribute-id', value: SYSTEM_IDS.PAGE_TYPE_TYPE }],
+  });
+
+  const entityId = pageTypeTriples.find(triple => triple.value.id === pageTypeId)?.entityId;
```
baiirun, line 70: Can you add a comment that we need to migrate this? Similar to the other comment. In the new data model this is `triple.value.value === pageTypeId`. I want to track all these spots in the codebase so we can make sure there aren't bugs before we introduce the migration work.

apps/web/core/utils/contracts/clone-entity.ts
```diff
@@ -0,0 +1,187 @@
...
+  const newEntityName = entityName ?? oldEntity.name ?? '';
+  const newTriples: Array<TripleType> = [];
+
+  const triplesToClone: Array<TripleType> = oldEntity.triples.filter(
+    (triple: TripleType) =>
+      ![SYSTEM_IDS.NAME, SYSTEM_IDS.AVATAR_ATTRIBUTE, SYSTEM_IDS.BLOCKS].includes(triple.attributeId)
+  );
+
+  newTriples.push(
+    Triple.withId({
```
baiirun, line 39: FYI all of the places where we create triples are changing in the new data model if we're mapping to "Actions" that we publish on chain. They're now called ops and look like this. So this is something that we'll have to migrate when we merge master into the migration branch.

```ts
type Op = {
  opType: "SET_TRIPLE" | "DELETE_TRIPLE",
  payload: {
    attributeId: string;
    entityId: string;
    value: {
      type: "ENTITY" | "TEXT" // etc
      value: string;
    }
  } 
}
```

baiirun: Not sure if it makes sense for us to merge this into the migration branch together after you merge this into main so the migration branch doesn't drift too much.

apps/web/app/tools/component.tsx
```diff
@@ -201,3 +212,43 @@ const FindEntities = () => {
...
+  const handleAddActions = async () => {
+    addActions(actions);
+  };
+
+  return (
+    <div className="space-y-4">
+      <Input
+        label="space name"
+        value={spaceName}
+        onChange={({ currentTarget: { value } }: any) => setSpaceName(value)}
```
baiirun, line 242: Use the actual types here

apps/web/app/tools/component.tsx
```diff
@@ -201,3 +212,43 @@ const FindEntities = () => {
...
+      />
+      <Input label="space ID" value={spaceId} onChange={({ currentTarget: { value } }: any) => setSpaceId(value)} />
```
baiirun, line 244: Use the actual types here

apps/web/app/tools/component.tsx
```diff
@@ -101,6 +111,7 @@ const FetchEntity = () => {
     <form onSubmit={handleFetchEntityId}>
       <Input label="entity ID" value={entityId} onChange={({ currentTarget: { value } }: any) => setEntityId(value)} />
```
baiirun, line 112: Use the actual types here

apps/web/app/tools/atoms.tsx
```diff
@@ -0,0 +1,16 @@
+import { atom } from 'jotai';
+import { atomWithStorage } from 'jotai/utils';
+
+const isDev = process.env.NODE_ENV === 'development';
+
+const atomWithDevStorage = (key: string, initialValue: any) => {
```
baiirun, line 6: use the actual types here

apps/web/core/utils/contracts/clone-entity.ts
```diff
@@ -0,0 +1,187 @@
...
+    oldEntity.triples.find((triple: TripleType) => triple.attributeId === SYSTEM_IDS.BLOCKS) ?? null;
+
+  if (blockIdsTriple) {
+    const blockIds = blockIdsTriple ? (JSON.parse(Value.stringValue(blockIdsTriple) || '[]') as string[]) : [];
+
+    const newBlockIds = blockIds.map(() => {
+      const newBlockId = ID.createEntityId();
+      return newBlockId;
+    });
+
```
baiirun, line 89: This logic changes in the new model since the blocks triple is a collection value type. Can you add a `@TODO(migration) comment here?

apps/web/core/utils/contracts/clone-entity.ts
```diff
@@ -0,0 +1,187 @@
...
+  if (!oldEntity) return [];
+
+  const newEntityId = entityId ?? ID.createEntityId();
+  const newEntityName = entityName ?? oldEntity.name ?? '';
+  const newTriples: Array<TripleType> = [];
+
+  const triplesToClone: Array<TripleType> = oldEntity.triples.filter(
+    (triple: TripleType) =>
+      ![SYSTEM_IDS.NAME, SYSTEM_IDS.AVATAR_ATTRIBUTE, SYSTEM_IDS.BLOCKS].includes(triple.attributeId)
```
baiirun, line 35: Can you move this tuple to outside of `cloneEntity` so we don't have to recreate the array every time clone entity is called?

apps/web/app/space/[id]/layout.tsx
```diff
@@ -47,13 +47,29 @@ async function buildTabsForSpacePage(types: EntityType[], params: Props['params'
...
+    filter: [{ field: 'attribute-id', value: SYSTEM_IDS.PAGE_TYPE_TYPE }],
+  });
+
+  const hasPostsPage = pageTriples.findIndex(triple => triple.value.id === SYSTEM_IDS.POSTS_PAGE) !== -1;
+  // const hasProductsPage = pageTriples.findIndex(triple => triple.value.id === SYSTEM_IDS.PRODUCTS_PAGE) !== -1;
+  // const hasServicesPage = pageTriples.findIndex(triple => triple.value.id === SYSTEM_IDS.SERVICES_PAGE) !== -1;
+  const hasEventsPage = pageTriples.findIndex(triple => triple.value.id === SYSTEM_IDS.EVENTS_PAGE) !== -1;
+  const hasProjectsPage = pageTriples.findIndex(triple => triple.value.id === SYSTEM_IDS.PROJECTS_PAGE) !== -1;
+  const hasJobsPage = pageTriples.findIndex(triple => triple.value.id === SYSTEM_IDS.JOBS_PAGE) !== -1;
+  const hasFinancesPage = pageTriples.findIndex(triple => triple.value.id === SYSTEM_IDS.FINANCES_PAGE) !== -1;
```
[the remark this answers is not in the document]
Comment (baiirun) [reply apps/web/app/space/[id]/layout.tsx:64]: Also you can probably just do `find` which will return undefined so you don't have to do the additional boolean check to see if `findIndex === -1`

apps/web/app/space/[id]/spaces/page.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { cachedFetchSpace } from '../cached-fetch-space';
+
+export default async function SpacesPage({ params }: { params: { id: string } }) {
+  const spaceId = params.id;
+  const space = await cachedFetchSpace(spaceId);
+
+  if (!space) return null;
+
+  // assumes the order is deterministic
+  const address = space.admins[0];
```
Comment (baiirun) [inline apps/web/app/space/[id]/spaces/page.tsx:18]: Is the spaces tab only for the `Person` type? If so you can use the entity id of the space config which is the person entity id, then query for onchain profiles with that entity id to get the address.

Comment (baiirun) [reply apps/web/app/space/[id]/spaces/page.tsx:18]: Something like

```ts
const query  = `
    {
        onchainProfile(id: entityId) {
            accountId
        }    
    }

`
```

apps/web/app/space/[id]/spaces/page.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const allSpaces = await Promise.all(spaceAddresses.map(spaceId => cachedFetchSpace(spaceId)));
+
+  const spaces = allSpaces
+    .map(space => {
+      if (!space) return null;
+
+      const entity = space?.spaceConfig;
+      const id = space.id;
+      const name = entity?.name ?? '';
+      const image = Entity.cover(entity?.triples) ?? Entity.avatar(entity?.triples) ?? null;
```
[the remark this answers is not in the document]
Comment (baiirun) [reply apps/web/app/space/[id]/spaces/page.tsx:42]: We can also just return the spaces from the `spacesWhereEditor` call instead of making another call to fetch the space again.

apps/web/app/tools/component.tsx
```diff
@@ -201,3 +224,47 @@ const FindEntities = () => {
...
+
+const CloneEntity = () => {
+  const [spaceName, setSpaceName] = useAtom(cloneSpaceNameAtom);
+  const [spaceId, setSpaceId] = useAtom(cloneSpaceIdAtom);
+  const [actions, setActions] = useAtom(cloneActionsAtom);
+
+  const { addActions } = useActionsStore();
+
+  const handleCloneEntity = async () => {
+    console.log(`cloning entity...`);
```
baiirun, line 236: Do we need this log?

apps/web/core/utils/contracts/clone-entity.ts
```diff
@@ -0,0 +1,192 @@
...
+};
+
+export const cloneEntity = async (options: Options) => {
+  if (!options.oldEntityId || !options.spaceId) {
+    throw new Error(`must specify all required options: oldEntityId and spaceId`);
+  }
+
+  const { oldEntityId, entityId = null, entityName = null, spaceId } = options;
+
+  console.info(options.entityName ? `cloning ${entityName}...` : `cloning entity...`);
```
baiirun, line 23: Do we need this log?

apps/web/partials/spaces/spaces.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { PLACEHOLDER_SPACE_IMAGE } from '~/core/constants';
+import { getImagePath } from '~/core/utils/utils';
+
+import { SpaceData } from '~/app/space/[id]/spaces/page';
+
+type SpacesProps = {
+  spaces: Array<SpaceData>;
+};
+
+export const Spaces = ({ spaces = [] }: SpacesProps) => {
```
baiirun, line 16: Don't need to set a default value since `spaces` isn't optional

apps/web/partials/tab/empty-tab.tsx
```diff
@@ -0,0 +1,52 @@
...
+export const EmptyTab = ({ spaceId, children }: EmptyTabProps) => {
+  const isEditing = useUserIsEditing(spaceId);
+  const [hasCreatedEntity, setHasCreatedEntity] = useState<boolean>(false);
+
+  const handleCreateEntity = () => {
+    setHasCreatedEntity(true);
+  };
+
+  if (!isEditing) {
+    redirect(NavUtils.toSpace(spaceId));
```
baiirun, line 28: I don't think `redirect` works in client components https://nextjs.org/docs/app/building-your-application/routing/redirecting

PR #743, opened 2024-06-10
feat: placeholders

sandren:
This PR adds support for customizable placeholder text and images for table blocks.


apps/web/partials/blocks/table/table-block.tsx
```diff
@@ -214,56 +219,89 @@ export const TableBlock = React.memo(({ spaceId }: Props) => {
...
+        )}
       </motion.div>
     </div>
   );
 });
 
+const getPlaceholders = (blockEntity: EntityType | null | undefined) => {
+  // @TODO add defaults for list/gallery views
+  let placeholderText = 'Add your first entity row to get started';
+  let placeholderImage = getImagePath('ipfs://QmfC4DoT7uVNoFRbP6DBYn9T79gpLXw2Uv6qJ2G8wmqT1d');
```
baiirun, line 282: We should move these out of the function so we don't have to recalculate every time the table re-renders. Normally it's not super expensive, but right now Geo does _a lot_ of recalculations across the entire app so it all adds up.

packages/ids/system-ids.ts
```diff
@@ -152,6 +152,8 @@ export const SPACE_CONFIGURATION = '1d5d0c2a-db23-466c-a0b0-9abe879df457';
 export const TABLE_BLOCK = '88d59252-17ae-4d9a-a367-24710129eb47';
 
 export const SHOWN_COLUMNS = '388ad59b-1cc7-413c-a0bb-34a4de48c758';
+export const PLACEHOLDER_TEXT = '0e5f84e4-c85a-4469-8a66-5a7d46fe2786';
```
baiirun, line 155: Can you put a `@TODO(migration)` for these? We'll have to update the existing ids and also add them to the substream bootstrapping

PR #748, opened 2024-06-13
Dockerfiles and docker compose

YaroShkvorets:


YaroShkvorets:
I assumed that old docker compose was an artifact that wasn't being used, so I overwrote it

baiirun (approved):

PR #751, opened 2024-06-18
feat: entity templates

sandren:
This PR adds support for specifying a template for types.


apps/web/partials/entity-page/editable-entity-page.tsx
```diff
@@ -104,6 +105,27 @@ export function EditableEntityPage({ id, spaceId, triples: serverTriples, typeId
...
+                entityName: name,
+                entityId: id,
+                spaceId,
+              });
+
+              newActions.forEach(action => {
+                create(action);
+              });
+            }
+          }
```
baiirun, line 128: More of a UX question, but we're starting to do a lot of async work when users are on the entity page. I wonder if we need some loading states to show it. Otherwise content just starts popping in randomly. Did you experience that when building this feature?

baiirun (approved):

apps/web/partials/entity-page/editable-entity-page.tsx
sandren: The cloned block content populated instantly for me, but for heavier templates, I could see that being an issue in the future.

PR #754, opened 2024-06-19
feat: refine filters

sandren:
This PR refines the filters applied to new entity creation from table blocks to only include attributes.


apps/web/partials/blocks/table/table-block.tsx
```diff
@@ -106,7 +106,9 @@ export const TableBlock = React.memo(({ spaceId }: Props) => {
 
   const typeId = type.entityId;
   const filters: Array<[string, string]> =
-    filterState && filterState.length > 0 ? filterState.map(filter => [filter.columnId, filter.value]) : [];
+    filterState && filterState.length > 0
```
baiirun, line 109: Can you add a comment explaining why we do this? Looking back at this in a year we probably won't remember why we wrote it this way

baiirun (approved):

PR #786, opened 2024-08-12
feat: data source menu

sandren:
This PR adds the new data source menu.

### DIFF
--- a/apps/web/core/state/table-block-store.tsx
+++ b/apps/web/core/state/table-block-store.tsx
@@ -25,7 +25,18 @@ export interface TableBlockFilter {
 
 type SpaceId = string;
 
-export type Source = 'collection' | 'geo' | Array<SpaceId>;
+type SingleSource = {
+  type: 'collection' | 'geo';
+  value: string;
+};
+
+// @TODO add support for `collections` with multiple `collectionId`s
+type MultipleSources = {
+  type: 'spaces';
+  value: Array<SpaceId>;
+};
+
+type Source = SingleSource | MultipleSources;
 
 export function useTableBlock() {
   const { entityId, selectedType, spaceId } = useTableBlockInstance();
@@ -281,17 +292,28 @@ export function useTableBlock() {
   const view = getView(blockEntity);
   const placeholder = getPlaceholder(blockEntity, view);
 
-  let source: Source = 'geo';
+  let source: Source = {
+    type: 'geo',
+    value: '',
+  };
 
   // required to coerce returned source type to `Source`
   // @TODO replace with proper logic for collection blocks
   const isCollection = false;
   if (isCollection) {
-    source = 'collection';
+    source = {
+      type: 'collection',
+      value: 'MOCK_COLLECTION_ID',
+    };
   }
 
   if (filterState && filterState.find(filter => filter.columnId === SYSTEM_IDS.SPACE)) {
-    source = filterState.filter(filter => filter.columnId === SYSTEM_IDS.SPACE).map(filter => filter.value);
+    const spaces = filterState.filter(filter => filter.columnId === SYSTEM_IDS.SPACE).map(filter => filter.value);
+
+    source = {
+      type: 'spaces',
+      value: spaces,
+    };
   }
 
   return {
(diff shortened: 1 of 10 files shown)


apps/web/core/state/table-block-store.tsx
```diff
@@ -23,6 +23,10 @@ export interface TableBlockFilter {
   valueName: string | null;
 }
 
+type SpaceId = string;
+
+export type Source = 'collection' | 'geo' | Array<SpaceId>;
```
baiirun, line 28: It might be simpler for consumers to make this a discriminated type. That would mean we don't need to handle the value differently depending on whether the type is a string or an array. This would make it so components like `DataBlockSourceMenu` don't have to check the type is an array, and instead can just check the source type since TypeScript handles type discrimination better this way.

```ts
type SingleSource = {
  type: 'collection' | 'geo'
  value: string;
}

type MultipleSources = {
  type: 'spaces' | 'collections'
  value: Array<SpaceId | CollectionId>
}

type Source = SingleSource | MultipleSources
```

apps/web/partials/blocks/table/data-block-source-menu.tsx
```diff
@@ -0,0 +1,187 @@
...
+};
+
+export const DataBlockSourceMenu = ({
+  collectionName = 'New collection',
+  onBack = () => null,
+}: DataBlockSourceMenuProps) => {
+  const [view, setView] = useState<View>('initial');
+  const { spaces } = useSpaces();
+  const { filterState, setFilterState, source } = useTableBlock();
+  const sourceIsSpaces = Array.isArray(source);
```
baiirun, line 36: I feel like having to check that the value is an array or not to know what to do kinda leaks the data model into the components which isn't usually a good idea. Using a type discriminator like in this comment should make consuming the sources less magical.

apps/web/partials/blocks/table/data-block-source-menu.tsx
```diff
@@ -0,0 +1,187 @@
...
+};
+
+export const SpacesMenu = ({ onBack }: SpacesMenuProps) => {
+  const [query, setQuery] = useState<string>('');
+  const { filterState, setFilterState } = useTableBlock();
+
+  const { spaces } = useSpaces();
+  const searchableSpaces = spaces.filter(space => space.type === 'PUBLIC' && !!space?.spaceConfig?.name);
+
+  const fuse = new Fuse(searchableSpaces, options);
```
baiirun, line 123: I think we should really be querying the spaces with a query instead of filtering the spaces in `useSpaces`. It's on my list to remove all the places we're reading from useSpaces. The list of spaces is going to grow exponentially once we launch, and right now `useSpaces` fetches every space and all the metadata for each space which is very slow and also consumes a lot of memory in the app.

We probably want a hook that can allow us to just query for the spaces by their name as the user types. I have an ad-hoc one I made in `add-subspace-dialog` that does this for querying subspaces.

apps/web/partials/blocks/table/table-block-context-menu.tsx
```diff
@@ -275,13 +278,15 @@ export function TableBlockContextMenu({ allColumns, shownColumnTriples, shownCol
...
       setIsMenuOpen(false);
+      setIsEditingDataSource(false);
       setIsEditingColumns(false);
     } else {
       setIsMenuOpen(true);
     }
   };
 
   const spaceImage = space?.spaceConfig?.image ?? null;
+  const isInitial = !isEditingDataSource && !isEditingColumns;
```
baiirun, line 289: What does `isInitial` mean?

apps/web/partials/blocks/table/table-block-editable-title.tsx
```diff
@@ -1,20 +1,92 @@
...
-      onBlur={e => setName(e.currentTarget.value)}
-      defaultValue={name ?? undefined}
-      placeholder="Enter a name for this table..."
-      className="w-full shrink-0 grow appearance-none text-smallTitle text-text outline-none placeholder:text-grey-03"
-    />
-  ) : (
-    <h4 className="text-smallTitle">{name}</h4>
+
+  const { name, setName, source } = useTableBlock();
+  const sourceIsSpaces = Array.isArray(source);
```
baiirun, line 14: Same comment as above

baiirun (approved):

PR #793, opened 2024-08-22
feat: personal space cards

sandren:
This PR adds the new personal space onboarding cards.


apps/web/design-system/icons/copy.tsx
```diff
@@ -4,15 +4,16 @@ import { ColorName, colors } from '~/design-system/theme/colors';
 
 interface Props {
   color?: ColorName;
+  fill?: string;
```
baiirun, line 7: This should be derived from our actual color names instead of being a string. So if we want to use `ctaPrimary` or `grey-04` we'll get autocomplete.

apps/web/partials/space-page/space-notices.tsx
```diff
@@ -113,27 +179,30 @@ type NoticeProps = {
...
+        orange: 'bg-gradient-orange',
       },
-    }
-  );
+      format: {
+        normal: 'col-span-1 aspect-square flex-col',
+        wide: 'col-span-3 flex-row gap-6 pr-10',
+      },
+    },
+  });
```
baiirun, line 205: This can be defined outside of the component

apps/web/partials/space-page/space-notices.tsx
```diff
@@ -104,7 +170,7 @@ type NoticesContainerProps = {
   children?: React.ReactNode;
 };
 
-const NoticesContainer = ({ children }: NoticesContainerProps) => {
+export const NoticesContainer = ({ children }: NoticesContainerProps) => {
```
baiirun, line 173: Why are `NoticesContainer` and `Notice` being explorted?

baiirun (approved):

PR #797, opened 2024-08-27
feat: onboarding

sandren:
This PR updates the initial account creation flow to the latest designs.


apps/web/core/hooks/use-onboarding.tsx
```diff
@@ -29,15 +30,15 @@ export function useOnboarding() {
   // @TODO: We should only show onboarding if the user is not a member
   // of any spaces OR there is no profile representing the user in
   // any of the spaces where they are a member.
-  React.useEffect(() => {
-    if (isModalOpen) {
+  useEffect(() => {
+    if (isModalOpen || !!user) {
```
baiirun, line 34: Does this work if the user doesn't sign in with Privy and instead uses an external wallet? My guess is no

baiirun: We need a better heuristic to see if the user has gone through onboarding already. What we decided was that we'll show the onboarding if they aren't a member of a space that has a Person entity with relations pointing to the currently connected wallet. We don't have a query to check for that yet.

apps/web/design-system/tooltip.tsx
```diff
@@ -1,63 +1,57 @@
...
-    <Popover.Root open={isTooltipShown} onOpenChange={setIsTooltipShown}>
-      <Popover.Anchor asChild>
-        <div onClick={() => setIsTooltipShown(true)} onMouseEnter={() => setIsTooltipShown(true)}>
-          {trigger}
-        </div>
-      </Popover.Anchor>
-      <Popover.Portal>
+    <Provider delayDuration={300} skipDelayDuration={300}>
+      <Root open={isOpen} onOpenChange={setIsOpen}>
+        <Trigger className={className}>{trigger}</Trigger>
```
baiirun, line 25: We should probably compose the styles on the trigger itself instead of passing both a trigger and a trigger class name

baiirun (approved):

PR #805, opened 2024-09-10
feat: company space cards

sandren:
This PR updates the company space cards and adds the new team onboarding card.


apps/web/partials/tab/tab.tsx
```diff
@@ -11,7 +13,7 @@ export type Props = {
   params: { id: string };
 };
 
-export const Tab = async (props: Props & { slug: string }) => {
+export const Tab = async (props: Props & { slug: string; notice: ReactNode }) => {
```
baiirun, line 16: Why not add this to the Prop type directly?

baiirun (approved):

packages/sdk/src/system-ids/system-ids.ts
```diff
@@ -275,6 +275,7 @@ export const POSTS_PAGE = 'e73c3db8320042309ae952eddb73b566'
 export const PRODUCTS_PAGE = '6764f3827ff247e2b2ad295791153705'
 export const SERVICES_PAGE = 'e5d69a755ede4a56b43344e5d3fde7bc'
 export const EVENTS_PAGE = 'bb2917434c394223afba91a08aa83478'
+export const TEAM_PAGE = '979eb04c-efa9-42b6-bd10-229bf7f0ce21'
```
baiirun, line 278: This should be the uuid without the dash - like the other ids look. Was this made on the L1 or the L2? If we don't make it on the L1 it won't exist when we migrate the real data to the L2

apps/web/partials/tab/tab.tsx
```diff
@@ -11,7 +13,7 @@ export type Props = {
   params: { id: string };
 };
 
-export const Tab = async (props: Props & { slug: string }) => {
+export const Tab = async (props: Props & { slug: string; notice?: ReactNode }) => {
```
baiirun, line 16: We should probably normalize the props that `Tab` receives instead of trying to combine it with the page props.

for each page route it would receive

```ts
type Props = {
  params: { id: string }
}
```

and Tab would get

```ts
type Props = {
  spaceId: string;
  slug: string;
  notice?: ReactNode;
}
```

baiirun (approved):

PR #830, opened 2024-10-18
fix: QA revisions

sandren:
This PR restores the default data block placeholder images and the new entity button in the create dropdown.

### DIFF
--- a/apps/web/core/state/table-block-store.tsx
+++ b/apps/web/core/state/table-block-store.tsx
@@ -18,6 +18,7 @@ import { Entity } from '../io/dto/entities';
 import { EntityId, SpaceId } from '../io/schema';
 import { Schema, ValueType as TripleValueType } from '../types';
 import { EntityTable } from '../utils/entity-table';
+import { getImagePath } from '../utils/utils';
 import { Values } from '../utils/value';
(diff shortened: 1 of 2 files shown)


apps/web/core/state/table-block-store.tsx
```diff
@@ -268,7 +269,8 @@ const getView = (blockEntity: Entity | null | undefined): DataBlockView => {
 
 const getPlaceholder = (blockEntity: Entity | null | undefined, view: DataBlockView) => {
   let text = DEFAULT_PLACEHOLDERS[view].text;
-  // let image = getImagePath(DEFAULT_PLACEHOLDERS[view].image);
+  // eslint-disable-next-line prefer-const
+  let image = getImagePath(DEFAULT_PLACEHOLDERS[view].image);
```
baiirun, line 273: Does image get changed anywhere? Looks like it can just be const instead of disabling the lint rule

baiirun (approved):

apps/web/core/state/table-block-store.tsx
sandren: Not yet, but will be when we restore the dynamic placeholders.

PR #857, opened 2024-11-11
feat: collection items

sandren:
This PR adds support for linked and verified collection items.

### DIFF
--- a/apps/web/core/io/subgraph/fetch-space.ts
+++ b/apps/web/core/io/subgraph/fetch-space.ts
@@ -76,7 +76,7 @@ export async function fetchSpace(options: FetchSpaceOptions): Promise<Space | nu
 
   const result = await Effect.runPromise(graphqlFetchWithErrorFallbacks);
 
-  if (!result?.space) {
+  if (!result.space) {
     return null;
   }
 
--- a/apps/web/core/io/subgraph/fetch-space.ts
+++ b/apps/web/core/io/subgraph/fetch-space.ts
@@ -76,7 +76,7 @@ export async function fetchSpace(options: FetchSpaceOptions): Promise<Space | nu
 
   const result = await Effect.runPromise(graphqlFetchWithErrorFallbacks);
 
-  if (!result.space) {
+  if (!result?.space) {
     return null;
   }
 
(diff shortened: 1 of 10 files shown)


apps/web/core/io/subgraph/fetch-space.ts
```diff
@@ -76,7 +76,7 @@ export async function fetchSpace(options: FetchSpaceOptions): Promise<Space | nu
 
   const result = await Effect.runPromise(graphqlFetchWithErrorFallbacks);
 
-  if (!result.space) {
+  if (!result?.space) {
```
baiirun, line 79: Do we need this? Does result sometimes return nullish?

apps/web/core/state/table-block-store.tsx
```diff
@@ -81,6 +81,15 @@ export function useTableBlock() {
     }, [blockEntity.relationsOut, source])
   );
 
+  const { data: collectionItemEntities, isLoading: isLoadingCollectionItemEntities } = useQuery({
+    queryKey: ['table-block-collection-item-entities', collectionItems.map(c => c.id).join('-')],
```
baiirun, line 85: We should use the `enabled` property in `useQuery` to only run this if `collectionItems.length > 0`

baiirun: Should also follow the new queryKeys pattern used in the other useQuery hooks in this file. You shouldn't need to turn the key into a string either since RQ deeply hashes arrays and objects internally.

apps/web/core/state/table-block-store.tsx
```diff
@@ -81,6 +83,18 @@ export function useTableBlock() {
     }, [blockEntity.relationsOut, source])
   );
 
+  const collectionItemIds = collectionItems?.map(c => c.id) ?? [];
```
baiirun, line 86: You can use `collectionItems` directly in the key instead of mapping to the ids. RQ hashes the query keys deterministically so we don't have to make our own hashable keys. https://tanstack.com/query/v4/docs/framework/react/guides/query-keys#query-keys-are-hashed-deterministically

baiirun (approved):

PR #877, opened 2024-11-26
fix: image uploads during space creation

baiirun:
We were previously using the incorrect entity id when setting the relation for an avatar or cover when creating spaces. We now return the image id from `Image.make` so consumers have easy access to the underlying image entity that was created.

Additionally, we weren't correctly handling the setting the image string for ad-hoc creation of spaces, specifically for company spaces. We now have a unified `image` state instead of an `avatar` and `cover` state.

Lastly, this fixes a bug where we weren't setting the space configuration on spaces that weren't `company`, `personal`, or `nonprofit` spaces.

### DIFF
--- a/apps/web/core/hooks/use-deploy-space.ts
+++ b/apps/web/core/hooks/use-deploy-space.ts
@@ -12,7 +12,7 @@ type DeployArgs =
   | {
       type: 'personal' | 'company';
       spaceName: string;
-      spaceAvatarUri?: string;
+      spaceImage?: string;
       governanceType?: SpaceGovernanceType;
     }
   | {
@@ -27,7 +27,7 @@ type DeployArgs =
         | 'government-org'
         | 'interest-group';
       spaceName: string;
-      spaceCoverUri?: string;
+      spaceImage?: string;
       governanceType?: SpaceGovernanceType;
     };
 
@@ -54,8 +54,8 @@ export function useDeploySpace() {
       );
 
       if (args.type === 'personal' || args.type === 'company') {
-        if (args.spaceAvatarUri && args.spaceAvatarUri !== '') {
-          url.searchParams.set('spaceAvatarUri', args.spaceAvatarUri);
+        if (args.spaceImage && args.spaceImage !== '') {
+          url.searchParams.set('spaceAvatarUri', args.spaceImage);
         }
       } else if (
         args.type === 'default' ||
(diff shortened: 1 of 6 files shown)


apps/web/core/hooks/use-deploy-space.ts
```diff
@@ -12,7 +12,7 @@ type DeployArgs =
   | {
       type: 'personal' | 'company';
       spaceName: string;
-      spaceAvatarUri?: string;
+      spaceImage?: string;
```
baiirun, line 15: We now just use a single `spaceImage` field for all the different types. Makes it simpler than having to conditionally switch between avatar and cover in consumer code.

apps/web/core/io/dto/entities.ts
```diff
@@ -70,9 +64,9 @@ export function EntityDto(substreamEntity: SubstreamEntity): Entity {
           // The "Renderable Type" for an entity provides a hint to the consumer
           // of the entity to _what_ the entity is so they know how they should
           // render it depending on their use case.
-          renderableType: isCoverOrAvatar ? 'IMAGE' : renderableType,
```
baiirun, line 73: Since we aren't reliant on migrations anymore, we can get rid of the temporary bandaid.

apps/web/core/utils/contracts/generate-ops-for-space-type.ts
```diff
@@ -60,10 +50,20 @@ export const generateOpsForSpaceType = async ({ type, spaceName, spaceAvatarUri,
...
+      ops.push(
+        ...Relation.make({
+          fromId: newEntityId,
+          toId: SYSTEM_IDS.SPACE_CONFIGURATION,
+          relationTypeId: SYSTEM_IDS.TYPES,
+        })
+      );
+      break;
+    }
   }
```
baiirun, line 63: We now correctly set the Space Configuration type on _all_ spaces, not just the blank space or the personal, nonprofit, or company spaces.

apps/web/core/utils/contracts/generate-ops-for-space-type.ts
```diff
@@ -60,10 +50,20 @@ export const generateOpsForSpaceType = async ({ type, spaceName, spaceAvatarUri,
...
 
   if (spaceAvatarUri) {
-    const imageOps = Image.make(spaceAvatarUri);
+    const { imageId, ops: imageOps } = Image.make(spaceAvatarUri);
```
baiirun, line 66: We now return the image id from `Image.make` so consumers can read the new image entity id easily, instead of having to parse the triples.

apps/web/partials/create-space/create-space-dialog.tsx
```diff
@@ -32,8 +32,7 @@ import { Animation } from '~/partials/onboarding/dialog';
 const spaceTypeAtom = atom<SpaceType | null>(null);
 const governanceTypeAtom = atom<SpaceGovernanceType | null>(null);
 const nameAtom = atom<string>('');
-const avatarAtom = atom<string>('');
-const coverAtom = atom<string>('');
+const imageAtom = atom<string>('');
```
baiirun, line 35: Unified `cover` and `avatar` into just `image`.

PR #883, opened 2024-12-03
fix: ts error in api build

proarunas:
This PR fixes a small typescript error in the API build process (TS2872: This kind of expression is always truthy).

### DIFF
--- a/packages/api/src/server.ts
+++ b/packages/api/src/server.ts
@@ -42,7 +42,7 @@ const server = app.listen(PORT, () => {
   const address = server.address();
 
   if (typeof address !== 'string') {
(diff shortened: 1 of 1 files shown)


proarunas:
It is also not exactly clear what is required for the .env file.

README.md is also a bit ambiguous. It can be interpreted to expect default credentials in the .env.example, but they are not there.
And there's no explanation of what services you are expected to run, in case that is intentional...

baiirun:
Hey @proarunas 👋🏼. Our codebases are still very unstable while we finalize the GRC-20 implementation and move towards launch. You won't find any docs on running the stack until we're more stable and feel good about the current implementation. For now we're still closed for external contributions until we get to good stability.

baiirun:
Closed for external contributions atm

PR #884, opened 2024-12-05
feat: extend relations type

sandren:
This PR extends the relations type to include a `space` like triples.

### DIFF
--- a/apps/web/core/events/edit-events.ts
+++ b/apps/web/core/events/edit-events.ts
@@ -50,7 +50,6 @@ export type EditEvent =
   | {
       type: 'UPSERT_RELATION';
       payload: {
-        spaceId: string;
         toEntityId: string;
         toEntityName: string | null;
         fromEntityId: string;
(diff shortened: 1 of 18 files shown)


apps/web/core/events/edit-events.ts
```diff
@@ -50,6 +50,7 @@ export type EditEvent =
   | {
       type: 'UPSERT_RELATION';
       payload: {
+        spaceId: string;
```
baiirun, line 53: This is already part of the event context, so you don't need to pass it in. Should be able to do 

```ts
const { spaceId } = event.context
```

apps/web/core/io/subgraph/fragments.ts
```diff
@@ -78,6 +78,9 @@ export const tripleFragment = `
  */
 export const relationFragment = `
   id
+  space {
```
baiirun, line 81: We should use `spaceId` instead since it's a bit faster to query as we don't have to traverse the relation in the database.

apps/web/partials/entity-page/editable-entity-table-cell.tsx
```diff
@@ -101,6 +101,7 @@ export const EditableEntityTableCell = memo(function EditableEntityTableCell({
                     send({
                       type: 'UPSERT_RELATION',
                       payload: {
+                        spaceId,
```
baiirun, line 104: Should be able to remove this if you use context in the event instead.

baiirun (approved):

PR #892, opened 2024-12-10
feat: search by entity id

sandren:
This PR extends the `useSearch` hook to allow searching by manually entered entity ID.

### DIFF
--- a/apps/web/core/utils/utils.ts
+++ b/apps/web/core/utils/utils.ts
@@ -1,7 +1,9 @@
+import { BASE58_ALLOWED_CHARS } from '@geogenesis/sdk';
 import { validate as uuidValidate, version as uuidVersion } from 'uuid';
 import { getAddress } from 'viem';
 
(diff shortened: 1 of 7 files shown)


apps/web/core/utils/utils.ts
```diff
@@ -314,3 +316,18 @@ export const uuidValidateV4 = (uuid: string) => {
 
   return uuidValidate(uuid) && uuidVersion(uuid) === 4;
 };
+
+export const validateEntityId = (maybeEntityId: EntityId | string | null | undefined) => {
+  if (typeof maybeEntityId !== 'string') return false;
+
+  if (maybeEntityId.length !== 22) return false;
```
baiirun, line 323: There are some edge cases where a valid base58 encoded string is less than 22 characters. I think the minimum in our implementation is 16. This is due to how repeating characters are compressed by the algo.

baiirun (approved):

PR #898, opened 2024-12-18
fix: add missing error logic

sandren:
This PR adds missing error logic to effects running in `useMutation`.

### DIFF
--- a/apps/web/core/hooks/use-add-editor.ts
+++ b/apps/web/core/hooks/use-add-editor.ts
@@ -71,7 +71,9 @@ export function useAddEditor(args: AddEditorArgs) {
         return hash;
       });
 
-      await Effect.runPromise(writeTxEffect);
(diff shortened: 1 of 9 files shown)


apps/web/core/hooks/use-add-editor.ts
```diff
@@ -71,7 +71,9 @@ export function useAddEditor(args: AddEditorArgs) {
         return hash;
       });
 
-      await Effect.runPromise(writeTxEffect);
+      await Effect.runPromise(writeTxEffect).catch(error => {
+        throw error;
+      });
```
baiirun, line 76: I'd expect `runPromise` should throw by default so we shouldn't need to rethrow. Have you checked by throwing an error inside of `writeTxEffect`? This is how Effect normally works

baiirun (approved):

PR #967, opened 2025-01-10
fix: voting polish

sandren:


apps/web/partials/active-proposal/accept-or-reject.tsx
```diff
@@ -64,14 +85,21 @@ export function AcceptOrReject({
...
+            <Pending isPending={isPendingRejection}>Reject</Pending>
+          </Button>
+          <span>or</span>
+          <Button onClick={onAccept} variant="success" disabled={voteStatus !== 'idle'}>
+            <Pending isPending={isPendingApproval}>Approve</Pending>
+          </Button>
+        </div>
+        {hasVoted && (
+          <div className="absolute inset-0 flex h-full w-full items-center justify-center">
+            <div className="text-smallButton">Vote registered</div>
```
baiirun, line 100: We should be able to use the existing `You accepted` or `You rejected` design higher up in this file to also show if the user has voted locally

baiirun (approved):

PR #990, opened 2025-01-11
feat: show arbitrary columns

sandren:
This PR introduces support for adding arbitrary columns to data blocks in table view.

### DIFF
--- a/apps/web/design-system/select-entity.tsx
+++ b/apps/web/design-system/select-entity.tsx
@@ -37,7 +37,9 @@ type SelectEntityProps = {
   allowedTypes?: RelationValueType[];
   placeholder?: string;
   containerClassName?: string;
+  searchClassName?: string;
   inputClassName?: string;
+  popoverClassName?: string;
   variant?: 'floating' | 'fixed';
   width?: 'clamped' | 'full';
   withSearchIcon?: boolean;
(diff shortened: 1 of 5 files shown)


apps/web/design-system/select-entity.tsx
```diff
@@ -182,12 +186,12 @@ export const SelectEntity = ({
...
             >
-              <div className={cx(variant === 'fixed' && 'pt-1', width === 'full' && 'w-full')}>
+              <div className={cx(variant === 'fixed' && 'pt-1', width === 'full' && 'w-full', popoverClassName)}>
                 <div
                   className={cx(
                     '-ml-px overflow-hidden rounded-md border border-divider bg-white',
-                    width === 'clamped' ? 'w-[400px]' : '-mr-px',
-                    withSearchIcon && 'rounded-t-none'
+                    width === 'clamped' ? 'w-[400px]' : '-mr-px'
+                    // withSearchIcon && 'rounded-t-none'
```
baiirun, line 194: Should we just remove this?

apps/web/partials/blocks/table/table-block-context-menu.tsx
```diff
@@ -401,6 +403,48 @@ const ToggleColumn = ({ column }: ToggleColumnProps) => {
...
+          const newRelation: StoreRelation = {
+            space: spaceId,
+            index: INITIAL_RELATION_INDEX_VALUE,
+            typeOf: {
+              id: EntityId(SYSTEM_IDS.SHOWN_COLUMNS),
+              name: 'Shown Columns',
+            },
+            fromEntity: {
+              id: EntityId(relationId),
+              name: '',
```
baiirun, line 425: Should use `null` instead of the empty string

PR #996, opened 2025-01-13
fix: revise clone entity

sandren:


apps/web/core/utils/contracts/clone-entity.ts
```diff
@@ -71,34 +68,34 @@ export const cloneEntity = async (options: Options): Promise<Array<Op>> => {
...
 
-  const blockOps = await cloneEntities(blocksToClone, newEntityId);
+  const blockOps = await cloneRelatedEntities(blocksToClone, newEntityId);
   newOps.push(...blockOps);
 
   return newOps;
 };
 
-const cloneEntities = async (entitiesToClone: Array<RelationType>, newEntityId: string) => {
+const cloneRelatedEntities = async (relatedEntitiesToClone: Array<RelationType>, newEntityId: string) => {
```
baiirun, line 80: We definitely need to check for cycles here. We could accidentally make a circular reference that causes us to clone infinitely.

Just need to check if we've seen `relation.id` before during the `cloneRelatedEntities` call.

baiirun (approved):

PR #1012, opened 2025-01-16
fix: query for quorum for proposals with single voter who is creator

baiirun:


apps/web/partials/governance/governance-proposals-list.tsx
```diff
@@ -422,13 +430,30 @@ async function fetchProposals({
...
+    // Votes should be >= 50%
+    .filter(p => {
       const votes = p.proposalVotes.nodes;
       const votesFor = votes.filter(v => v.vote === 'ACCEPT');
       const yesPercentage = Math.floor((votesFor.length / votes.length) * 100);
       return yesPercentage > 50;
-    }),
+    })
+    // Quorum
+    .filter(p => {
```
baiirun, line 447: Most relevant change is here. Rest of the changes relate to unifying how we query for the account id of the voters

PR #1105, opened 2025-02-26
feat: create version in space

sandren:


apps/web/app/space/(entity)/[id]/[entityId]/default-entity-page.tsx
```diff
@@ -97,15 +97,6 @@ const getData = async (spaceId: string, entityId: string) => {
...
   }
 
-  // Redirect from an invalid space to a valid one
-  if (entity && !spaces.includes(spaceId)) {
-    const newSpaceId = Spaces.getValidSpaceIdForEntity(entity);
-
-    console.log(`Redirecting from invalid space ${spaceId} to valid space ${spaceId}`);
-
-    return redirect(NavUtils.toEntity(newSpaceId, entityId));
-  }
```
Comment (baiirun) [inline apps/web/app/space/(entity)/[id]/[entityId]/default-entity-page.tsx:107]: What happens here if you navigate to an entity from a space where the entity doesn't exist?

apps/web/partials/entity-page/entity-page-context-menu.tsx
```diff
@@ -77,3 +134,67 @@ function EntityPageContextMenuItem({ children }: EntityPageContextMenuItemProps)
...
+
+  const [query, setQuery] = useState<string>('');
+
+  const fuseOptions = {
+    keys: ['spaceConfig.name', 'spaceConfig.description'],
+  };
+
+  const fuse = new Fuse(spaces, fuseOptions);
+
+  const renderedSpaces = query.length === 0 ? spaces : fuse.search(query).map(result => result.item);
```
baiirun, line 159: What does fuse get us? Usually fuse is used for large data sets but ours aren't very large.

apps/web/partials/entity-page/automatic-mode-toggle.tsx
```diff
@@ -0,0 +1,25 @@
...
+      if (!editable && shouldStartInEditMode) {
+        setEditable(true);
+      }
+
+      setShouldStartInEditMode(false);
+    }, 100);
+  });
+
+  return null;
+};
```
baiirun, line 25: Is there a reason we can't toggle to edit mode in the event handler for creating the new entity?

apps/web/partials/search/dialog.tsx
```diff
@@ -172,3 +212,88 @@ export const SearchDialog = ({ open, onDone }: Props) => {
...
+        <Input value={query} onChange={event => setQuery(event.target.value)} withSearchIcon />
+      </div>
+      <div className="max-h-[190px] space-y-1 overflow-auto p-1">
+        {renderedSpaces.map(space => {
+          return (
+            <Command.Item
+              key={space.id}
+              onSelect={() => {
+                // Ensure they arrive in edit mode
+                setShouldStartInEditMode(true);
```
baiirun, line 279: Would it be better to use a URL param instead of global state that we need to manually track?

PR #1108, opened 2025-03-03
feat: format date & time renderables

mikallojjus:
image
--
image
-- 
image

Designs: https://www.figma.com/design/iWKsYButqqKWd1bqeBrUUj/Geo---Web?node-id=36725-13622&p=f&t=UcEqP7T7cSo05YCU-0


apps/web/partials/diff/changed-entity.tsx
```diff
@@ -542,8 +557,8 @@ const ChangedAttribute = ({ changes, renderAttributeStagingComponent }: ChangedA
 
 type DateTimeProps = {
   mode: 'before' | 'after';
-  before: string | null;
-  after: string | null;
+  before: TripleChangeValue;
+  after: TripleChangeValue;
```
baiirun, line 561: What happens in a diff if the one of the entities in the diff doesn't have the triple being compared? e.g., you're adding a new time triple or deleting an existing time triple. 

Previously we'd use `| null` to denote that the triple may or may not exist.

apps/web/core/utils/change/change.ts
```diff
@@ -389,13 +390,17 @@ export function aggregateChanges({
     const tripleChanges: TripleChange[] = [];
     const relationChanges: RelationChange[] = [];
 
+    const processedAttributes = new Set<string>();
```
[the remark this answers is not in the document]
mikallojjus:
```
if (afterValue.value !== beforeValue.value || !equal(afterValue.options, beforeValue.options)) {
  // This is considered a change
}
```

It's now detecting changes in the options object (like format changes for dates) as actual changes, even if the underlying value hasn't changed.

The duplication happens because:
First loop: When processing `afterTriplesForEntity`, it finds a change in options and creates a change entry
Second loop: When processing beforeTriplesForEntity, it finds the same attribute again, detects the options change, and creates another change entry

By processing both "after to before" and "before to after" comparisons, it catches all these cases, but without tracking which attributes have been processed, it creates duplicate entries for updates.

apps/web/partials/diff/changed-entity.tsx
mikallojjus: Good catch, I forgot to test adding/removing triples.

baiirun (approved):

PR #1113, opened 2025-03-03
fix: academic field system id

sandren:


packages/sdk/src/core/ids/system.ts
```diff
@@ -142,7 +142,7 @@ export const VERIFIED_SOURCE_ATTRIBUTE = '5jodArZNFzucsYzQaDVFBL';
 export const SOURCE_SPACE_ATTRIBUTE = 'GzkEQP3yedWjXE8QPFKEwV';
 
 /** Core types */
-export const ACADEMIC_FIELD_TYPE = 'ExCjm3rzYVfpMRwDchdrE';
```
baiirun, line 145: We actually aren't using this package anymore in favor of https://github.com/graphprotocol/grc-20-ts. The academic field id should already be updated there.

baiirun:
Close in favor of using grc-20 lib

PR #1116, opened 2025-03-05
feat: format date values

mikallojjus:
Retrieve triple's options.format and use that to format dates


apps/web/design-system/editable-fields/date-field.tsx
```diff
@@ -18,6 +18,7 @@ interface DateFieldProps {
   variant?: 'body' | 'tableCell';
   value: string;
   isEditing?: boolean;
+  format?: string;
```
baiirun, line 21: Where is the format being set in the UI? I remember you having an unimplemented element in the UI. Is that in the other PR?

mikallojjus: Yes, this PR only contains code for displaying formatted values.

baiirun (approved):

PR #1120, opened 2025-03-07
fix: relation value types

sandren:


apps/web/core/hooks/use-properties.ts
```diff
@@ -35,27 +35,21 @@ export function useProperties(propertyIds: string[]): UsePropertyValueTypes {
...
       const schema = properties.map((s): PropertySchema => {
-        const relationValueType = relationValueTypes.find(t => t.attributeId === s.id) ?? null;
         return {
           id: s.id,
           name: s.name,
           valueType: (valueTypes.find(v => v.attributeId === s.id)?.valueTypeId ?? SystemIds.TEXT) as ValueTypeId,
-          relationValueTypeId: relationValueType?.relationValueTypeId,
-          relationValueTypeName: relationValueType?.relationValueTypeName,
           homeSpace: s.spaces[0],
+          relationValueTypes,
```
baiirun, line 52: I think this adds the relation types from every property to every other property. We need to only set the relation types for the _specific_ property.

baiirun (approved):

PR #1145, opened 2025-04-03
feat: add bulleted list view

b-d055:
This PR adds bulleted list view to data blocks.

### DIFF
--- a/apps/web/partials/blocks/table/table-block-bulleted-list-item.tsx
+++ b/apps/web/partials/blocks/table/table-block-bulleted-list-item.tsx
@@ -1,0 +1,147 @@
+import { SystemIds } from '@graphprotocol/grc-20';
+import Link from 'next/link';
+
+import { editEvent } from '~/core/events/edit-events';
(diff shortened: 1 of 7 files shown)


b-d055:
Ready for review, reusing the list placeholder icon for now

apps/web/partials/blocks/table/table-block-bulleted-list-item.tsx
```diff
@@ -0,0 +1,147 @@
...
+                    entityName: result.name,
+                    spaceId: currentSpaceId,
+                  },
+                  {
+                    type: 'FOC',
+                    data: result,
+                  }
+                );
+              }}
+              spaceId={currentSpaceId}
```
baiirun, line 83: This pattern has changed a little bit on master. The `FOC` event type doesn't exist anymore. Should be able to merge in master then copy the same usage in the gallery or list views.

b-d055: @baiirun updated, let me know if the update matches the new pattern (it seems to be working in testing)

PR #1146, opened 2025-04-03
feat: collection item actions

sandren:
This PR implements linking, verifying, and deleting actions for collection items.

### DIFF
--- a/apps/web/core/blocks/data/use-collection.ts
+++ b/apps/web/core/blocks/data/use-collection.ts
@@ -49,6 +49,7 @@ export function useCollection() {
   });
 
   return {
+    blockEntity,
     collectionItems,
     collectionRelations,
     isLoading: isCollectionItemsLoading || isCollectionRelationsLoading,
--- a/apps/web/core/blocks/data/use-data-block.tsx
+++ b/apps/web/core/blocks/data/use-data-block.tsx
@@ -48,6 +48,7 @@ export function useDataBlock() {
   const { shownColumnIds, mapping, isLoading: isViewLoading, isFetched: isViewFetched } = useView();
   const { source } = useSource();
   const {
+    blockEntity,
     collectionItems,
     collectionRelations,
     isFetched: isCollectionFetched,
@@ -188,6 +189,7 @@ export function useDataBlock() {
     spaceId,
     relationId,
 
+    blockEntity,
     rows: rows?.slice(0, PAGE_SIZE) ?? [],
     properties: propertiesSchema ? Object.values(propertiesSchema) : [],
     propertiesSchema,
--- a/apps/web/core/hooks/use-result.ts
+++ b/apps/web/core/hooks/use-result.ts
@@ -1,0 +1,53 @@
+'use client';
+
+import { useQuery } from '@tanstack/react-query';
+import * as Effect from 'effect/Effect';
+import * as Either from 'effect/Either';
+
+import { Subgraph } from '~/core/io';
+import { EntityId } from '~/core/io/schema';
+
+import { mergeSearchResult } from '../database/result';
+import { useSyncEngine } from '../sync/use-sync-engine';
+
+export const useResult = (id: EntityId) => {
+  const { store } = useSyncEngine();
+
+  const { data: result, isLoading } = useQuery({
+    queryKey: ['result', id],
+    queryFn: async () => {
+      const fetchResultEffect = Effect.either(
+        Effect.tryPromise({
+          try: async () =>
+            await mergeSearchResult({
+              id,
+              store,
+            }),
+          catch: error => {
+            console.error('error', error);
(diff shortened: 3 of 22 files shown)


apps/web/core/blocks/data/use-collection.ts
```diff
@@ -49,6 +49,7 @@ export function useCollection() {
   });
 
   return {
+    blockEntity,
```
baiirun, line 52: Do we still need this? Doesn't look like it's used anywhere

apps/web/core/blocks/data/use-data-block.tsx
```diff
@@ -48,6 +48,7 @@ export function useDataBlock() {
   const { shownColumnIds, mapping, isLoading: isViewLoading, isFetched: isViewFetched } = useView();
   const { source } = useSource();
   const {
+    blockEntity,
```
baiirun, line 51: This block entity is the same block entity already queried by the `useDataBlock` hook on line 41

apps/web/core/hooks/use-result.ts
```diff
@@ -0,0 +1,53 @@
+'use client';
```
baiirun, line 1: Is this different from the `useSearch` hook where we pass a single id as the query and get a result?

apps/web/design-system/editable-fields/editable-fields.tsx
```diff
@@ -80,17 +81,30 @@ export function TableStringField({ ...props }: TableStringFieldProps) {
...
+
+  useEffectOnce(() => {
+    if (autoFocus) {
+      setTimeout(() => {
+        ref.current.focus();
+        const length = ref.current.value.length;
+        ref.current.setSelectionRange(length, length);
+      }, 200);
+    }
+  });
```
baiirun, line 106: What happens in instances where there's multiple page string fields that render at once? For example in placeholder schema properties in the properties panel or in edit mode in data blocks?

apps/web/partials/blocks/table/table-block-gallery-item.tsx
```diff
@@ -303,14 +281,19 @@ export function TableBlockGalleryItem({
...
+            isEditing={false}
+            name={name}
+            href={href}
+            currentSpaceId={currentSpaceId}
+            entityId={rowEntityId}
+            spaceId={nameCell?.space}
+            relationId={relationId}
+            verified={verified}
+            onChangeEntry={() => null}
+            onLinkEntry={() => null}
```
baiirun, line 295: Why are these undefined?

apps/web/partials/blocks/table/table-block-table.tsx
```diff
@@ -259,45 +245,83 @@ export const TableBlockTable = ({
               const cells = row.getVisibleCells();
               const entityId = cells?.[0]?.getValue<Cell>()?.cellId;
 
+              if (row.original.placeholder) {
+                return (
+                  <TableCell key={entityId ?? index} width={784} isShown>
+                    <SelectEntity
```
baiirun, line 251: Why define this here instead of in the EditableCell like it currently exists?

apps/web/design-system/select-space.tsx
```diff
@@ -0,0 +1,235 @@
...
+    onQueryChange(EntityId(entityId));
+  });
+
+  const result = results?.[0];
+
+  const { data: spaceVersions, isLoading: isSpaceVersionsLoading } = useQuery({
+    enabled: !!query && !!result?.spaces && result.spaces.length > 0,
+    queryKey: ['space-versions', query, result?.spaces],
+    queryFn: async () => {
+      return await Promise.all(
```
baiirun, line 65: We can use `fetchEntityBatch` here so optimize a bit.

apps/web/partials/blocks/table/table-block.tsx
```diff
@@ -210,6 +207,29 @@ function useEntries(entries: Row[], properties: PropertySchema[], spaceId: strin
     }
   };
 
+  const onLinkEntry = (
```
baiirun, line 210: When do we use `onLinkEntry` vs using the `onChangeEntry`?

sandren: `onChangeEntry` is used to create new collection items in `<SelectEntity>` while the `onLinkEntry` is used in `<SelectSpace>` to set the source space and verification. There is some overlap, but they have to be separate functions because `onChangeEntry` creates a new relation ID while `onLinkEntry` expects you to pass in the existing relation ID.

apps/web/partials/blocks/table/table-block-table.tsx
baiirun: @sandren bumping this

baiirun (approved):

PR #1167, opened 2025-04-17
feat: collection metadata

sandren:


apps/web/partials/blocks/table/table-block-table-item.tsx
```diff
@@ -27,21 +36,119 @@ export const TableBlockTableItem = ({
   verified,
   onChangeEntry,
   onLinkEntry,
+  source,
 }: TableBlockTableItemProps) => {
+  if (isEditing) {
```
baiirun, line 41: We should be able to put all of this logic into `editable-entity-table-cell`. We're already managing rendering the name cell there and toggling between different UI depending on if its a collection, placeholder, etc. For edit mode anyway.

For the browse mode renderable we can update `entity-table-cell` which does the same work, but for browse mode.

sandren: The `isNameCell` condition of those components are no longer reached (can remove that dead code) due to us returning the new component in `<TableBlockTable>` here: https://github.com/geobrowser/geogenesis/blob/master/apps/web/partials/blocks/table/table-block-table.tsx#L296

baiirun: Yeah I'm suggesting moving the logic from `table-block-table-item` into these other files. That way all the logic for rendering content in a data block is in the same place. That's how it works for list and gallery views, whereas for tables some of the logic is in the link you sent, but some of the logic is in `EditableEntityTableCell` and `EntityTableCell`.

baiirun (approved):

PR #1168, opened 2025-04-18
Feat: geo location & mapbox

b-d055:
Add point type, geo location property, and mapbox


apps/web/design-system/editable-fields/editable-fields.tsx
```diff
@@ -334,3 +343,97 @@ export function TableImageField({ imageSrc, onImageChange, onImageRemove, varian
...
+    setLocalValue(props.value || '');
+    setLocalFormat(props.format || '');
+    console.log('props.value', props.value);
+    console.log('props.format', props.format);
+  }, [props.value, props.format]);
+
+  // Apply debounce effect
+  const debouncedCallback = debounce((value: string, format: string) => {
+    onChange(value, format);
+  }, 1000);
```
baiirun, line 386: There's a new hook for this pattern. `useOptimisticValueWithCallback`. Some of the other components in this file are using it a well.

apps/web/design-system/editable-fields/editable-fields.tsx
```diff
@@ -334,3 +343,97 @@ export function TableImageField({ imageSrc, onImageChange, onImageRemove, varian
...
+
+  // Handle map visibility toggle
+  const handleShowMapToggle = () => {
+    if (localFormat === 'EARTH COORDINATES') {
+      setLocalFormat('');
+    } else {
+      setLocalFormat('EARTH COORDINATES');
```
baiirun, line 393: Did we fully align with Yaniv that this is the correct name for this format? For these we should probably export them from the GRC-20 package so they're available for anybody building on the knowledge graph.

apps/web/design-system/editable-fields/editable-fields.tsx
```diff
@@ -334,3 +343,97 @@ export function TableImageField({ imageSrc, onImageChange, onImageRemove, varian
...
+          {/* Toggle */}
+          <div className="relative h-3 w-5 cursor-pointer rounded-lg bg-black" onClick={handleShowMapToggle}>
+            <div
+              className={`absolute top-[1px] h-[10px] w-[10px] rounded-full bg-white transition-all duration-300 ease-in-out ${localFormat === 'EARTH COORDINATES' ? 'right-[1px]' : 'right-[9px]'}`}
+            ></div>
+          </div>
+          <span className="text-[1rem] font-normal leading-5 text-grey-04">Show map in browse mode</span>
+        </div>
+      </div>
+      <Map
```
baiirun, line 432: Do we know the bundle size of this Map component? Our bundle size is already pretty large due to some of our other dependencies like tiptap. We should probably lazy load/dynamic import this component if we can.

b-d055: Yeah it's a chonker - it accounts for ~400mb on First Load JS. Added a lazy load for the mapbox-gl import itself. Used a basic skeleton and looks something like this on load. 

https://github.com/user-attachments/assets/a59ea335-5233-45c4-ac9d-b4f6a700018b

apps/web/design-system/editable-fields/editable-fields.tsx
baiirun: I think `format` makes sense, but just need to align on what the actual format value is

baiirun:
We should add diffs for the lat/long as well

apps/web/core/types.ts
```diff
@@ -11,10 +11,11 @@ export type ValueType =
   | 'TIME'
   | 'NUMBER'
   // | GEO_LOCATION
```
baiirun, line 13: We can remove this GEO_LOCATION placeholder

apps/web/core/utils/utils.ts
```diff
@@ -88,6 +88,42 @@ export class GeoNumber {
...
+      
+      const latitude = parseFloat(coordParts[0]);
+      const longitude = parseFloat(coordParts[1]);
+      
+      if (isNaN(latitude) || isNaN(longitude)) return undefined;
+      
+      return { latitude, longitude };
+    } catch (e) {
+      console.error(`Unable to parse coordinates: "${value}"`);
+      return undefined;
```
baiirun, line 112: Let's use `null` instead of undefined. Most places in the codebase we align on `null` to represent "empty" values

apps/web/design-system/editable-fields/geo-location-field.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const validNumberPattern = /^-?\d*\.?\d*$/;
+
+  const handlePointValueChange = (label: string, newValue: string) => {
+    if (newValue === '' || validNumberPattern.test(newValue)) {
+      const updatedPoints = {
+        ...pointValues,
+        [label]: newValue,
+      };
+      setPointsValues(updatedPoints);
+      setLocalValue(`${updatedPoints.latitude},${updatedPoints.longitude}`);
```
baiirun, line 59: Can't we use the new `GeoPoint.format(...)` function to create this string?

apps/web/design-system/editable-fields/geo-location-field.tsx
```diff
@@ -0,0 +1,133 @@
...
+    } else {
+      console.error('Invalid input');
```
baiirun, line 61: Could use a more useful error message here. We'd want to know where the error is happening and why

apps/web/design-system/map.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const markerRef = useRef<mapboxgl.Marker | null>(null);
+  const [isMapboxLoaded, setIsMapboxLoaded] = useState(false);
+
+  // Initialize map when component mounts
+  useEffect(() => {
+    if (!showMap || !mapContainerRef.current) return;
+
+    const loadMapbox = async () => {
+      // Dynamically import mapbox only when needed
+      const mapboxgl = (await import('mapbox-gl')).default;
```
baiirun, line 30: Can we skip importing if we already have the `mapRef`? Otherwise we might be importing every time lat/long changes

apps/web/partials/entity-page/editable-entity-page.tsx
```diff
@@ -119,7 +125,9 @@ export function EditableEntityPage({ id, spaceId, triples: serverTriples }: Prop
                   <TriplesGroup key={attributeId} triples={renderables as TripleRenderableProperty[]} />
                 )}
 
-                <div className="absolute right-0 top-6 flex items-center gap-1">
+                <div
+                  className={`absolute right-0 flex items-center gap-1 ${firstRenderable.attributeId === SystemIds.GEO_LOCATION_PROPERTY && renderableType === 'POINT' ? 'top-0' : 'top-6'}`}
+                >
```
baiirun, line 130: Might be worth adding a comment as to why we need separate styles for different value types

apps/web/partials/entity-page/editable-entity-page.tsx
```diff
@@ -551,25 +559,27 @@ function TriplesGroup({ triples }: TriplesGroupProps) {
...
-                value={renderable.value}
-                onChange={value => {
-                  send({
-                    type: 'UPSERT_RENDERABLE_TRIPLE_VALUE',
-                    payload: {
-                      renderable,
-                      value: {
-                        type: 'TEXT',
-                        value: value,
+              <>
```
baiirun, line 562: Do we still need the fragment?

apps/web/partials/entity-page/get-renderable-type-options.tsx
```diff
@@ -134,7 +137,7 @@ export const getRenderableTypeSelectorOptions = (
           <IconWrapper>
             <Date />
           </IconWrapper>
-          <p>Date & time</p>
+          <p>Date & time </p>
```
baiirun, line 140: Do we need the space here?

apps/web/partials/entity-page/readable-entity-page.tsx
```diff
@@ -1,17 +1,21 @@
+import { cons } from 'effect/List';
```
baiirun, line 1: Do we need this import? not sure why the linter CI step isn't catching it

apps/web/design-system/editable-fields/editable-fields.tsx
```diff
@@ -4,7 +4,7 @@ import Zoom from 'react-medium-image-zoom';
 import Textarea from 'react-textarea-autosize';
 
 import * as React from 'react';
-import { ChangeEvent, useRef } from 'react';
+import { ChangeEvent, useRef, useEffect } from 'react';
```
baiirun, line 7: Should be able to remove this `useEffect`

apps/web/design-system/editable-fields/editable-fields.tsx
```diff
@@ -66,6 +66,14 @@ type PageStringFieldProps = {
...
 };
 
+interface PageGeoLocationFieldProps {
+  onChange: (value: string, isBrowseMode: string) => void;
+  placeholder?: string;
+  variant?: 'mainPage' | 'body' | 'smallTitle';
+  value?: string;
+  format?: string;
+}
+
```
baiirun, line 76: Do we need this interface defined here?

apps/web/design-system/editable-fields/geo-location-field.tsx
```diff
@@ -0,0 +1,141 @@
+import * as React from 'react';
+import { useEffect } from 'react';
+import Textarea from 'react-textarea-autosize';
+
+import { useOptimisticValueWithSideEffect } from '~/core/hooks/use-debounced-value';
+import { Map } from '../map';
+import { cva } from 'class-variance-authority';
+import { GeoPoint } from '~/core/utils/utils';
+
+const DISPLAY_MAP_FORMAT_OPTION = 'EARTH COORDINATES';
```
baiirun, line 10: Did we align on which earth coordinates spec we use for this?

apps/web/design-system/editable-fields/editable-fields.tsx
b-d055: nope, this was moved, thanks

apps/web/design-system/editable-fields/geo-location-field.tsx
b-d055: per Yaniv, removing the toggle all together for now - all geo location point values will render as map

baiirun (approved):

PR #1173, opened 2025-04-22
feat: currency units

mikallojjus:
Currency units implemented

---

image

---

image

---

image

---

image

---

image

---

image

### DIFF
--- a/apps/web/design-system/editable-fields/number-field.tsx
+++ b/apps/web/design-system/editable-fields/number-field.tsx
@@ -1,3 +1,4 @@
+import { SystemIds } from '@graphprotocol/grc-20';
 import { cva } from 'class-variance-authority';
 
 import * as React from 'react';
@@ -46,7 +47,7 @@ export function NumberField({
   const { entity } = useQueryEntity({ id: unitId });
 
   const currencySign = React.useMemo(
-    () => unitId && entity?.triples.find(t => t.attributeId === 'Tt2mYqE1kJTRLt2iLQjATb')?.value?.value,
+    () => unitId && entity?.triples.find(t => t.attributeId === SystemIds.CURRENCY_SIGN_ATTRIBUTE)?.value?.value,
     [unitId, entity]
   );
 
--- a/apps/web/design-system/editable-fields/number-field.tsx
+++ b/apps/web/design-system/editable-fields/number-field.tsx
@@ -2,6 +2,7 @@ import { cva } from 'class-variance-authority';
 
 import * as React from 'react';
 
+import { useQueryEntity } from '~/core/sync/use-store';
 import { GeoNumber } from '~/core/utils/utils';
 
 type Props = {
@@ -9,6 +10,7 @@ type Props = {
   placeholder?: string;
   value?: string;
   format?: string;
+  unitId?: string;
   isEditing?: boolean;
   variant?: 'body' | 'tableCell';
 };
@@ -29,6 +31,7 @@ export function NumberField({
   onChange,
   value,
   format,
+  unitId,
   placeholder = 'Add value...',
   isEditing = false,
   variant,
@@ -40,8 +43,15 @@ export function NumberField({
     setLocalValue(value || '');
   }, [value]);
 
+  const { entity } = useQueryEntity({ id: unitId });
+
+  const currencySign = React.useMemo(
+    () => entity?.triples.find(t => t.attributeId === 'Tt2mYqE1kJTRLt2iLQjATb')?.value?.value,
+    [entity]
+  );
+
   if (!isEditing) {
-    return <span className={numberFieldStyles({ variant })}>{GeoNumber.format(value, format)}</span>;
+    return <span className={numberFieldStyles({ variant })}>{GeoNumber.format(value, format, currencySign)}</span>;
   }
 
   if (isEditing && !onChange) {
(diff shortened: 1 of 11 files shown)


apps/web/partials/diff/changed-entity.tsx
```diff
@@ -600,8 +601,26 @@ interface NumberDiffProps {
...
+  const { entity: afterUnitEntity } = useQueryEntity({ id: after?.options?.unit });
+
+  const [currencySignBefore, currencySignAfter] = React.useMemo(
+    () => [
+      before?.options?.unit &&
+        beforeUnitEntity?.triples.find(t => t.attributeId === 'Tt2mYqE1kJTRLt2iLQjATb')?.value?.value,
+      after?.options?.unit &&
+        afterUnitEntity?.triples.find(t => t.attributeId === 'Tt2mYqE1kJTRLt2iLQjATb')?.value?.value,
+    ],
+    [before, beforeUnitEntity, after, afterUnitEntity]
```
mikallojjus, line 615: Checking for `[before | after]?.options?.unit` is a workaround, which I'm not a fan of. It is because `useQueryEntity` skips API request if `id` is `undefined`. 

https://github.com/geobrowser/geogenesis/blob/feat/currency-units/apps/web/core/sync/use-store.tsx#L49-L51

I could change it to:
```
if (!enabled) {
  return;
}
```
But I cannot guarantee it won't break anything.

baiirun: I'm not seeing `[before | after]?.options?.unit` anywhere in the PR. Am I missing something?

apps/web/partials/entity-page/number-options-dropdown.tsx
```diff
@@ -143,6 +134,109 @@ const NumberFormatView = ({ formatOptions, value }: { formatOptions?: NumberForm
...
+    { name: string | null; symbol?: string; sign?: string; id: string }[]
+  >([]);
+
+  const {
+    query,
+    onQueryChange,
+    isLoading: isSearchLoading,
+    results: searchResults,
+  } = useSearch({
+    filterByTypes: ['K6kJ6TTwwCcc3Dfdtbradv'], // Attribute Id for type currency
```
baiirun, line 149: We should add this id to @graphprotocol/grc-20 and import it from the package instead of hardcoding it.

apps/web/partials/diff/changed-entity.tsx
```diff
@@ -600,8 +601,26 @@ interface NumberDiffProps {
...
-  const formattedNumberBefore = before?.value ? GeoNumber.format(before.value, before?.options?.format) : null;
-  const formattedNumberAfter = after?.value ? GeoNumber.format(after.value, after?.options?.format) : null;
+
+  const { entity: beforeUnitEntity } = useQueryEntity({ id: before?.options?.unit });
+  const { entity: afterUnitEntity } = useQueryEntity({ id: after?.options?.unit });
+
+  const [currencySignBefore, currencySignAfter] = React.useMemo(
+    () => [
+      before?.options?.unit &&
+        beforeUnitEntity?.triples.find(t => t.attributeId === 'Tt2mYqE1kJTRLt2iLQjATb')?.value?.value,
```
baiirun, line 611: We should add this id to @graphprotocol/grc-20 and import it from the package instead of hardcoding it.

apps/web/design-system/editable-fields/number-field.tsx
```diff
@@ -1,33 +1,42 @@
...
 
   React.useEffect(() => {
     // Update local value if value prop changes from outside the component
     setLocalValue(value || '');
   }, [value]);
 
+  const { entity } = useQueryEntity({ id: unitId });
+
+  const currencySign = React.useMemo(
+    () => unitId && entity?.triples.find(t => t.attributeId === 'Tt2mYqE1kJTRLt2iLQjATb')?.value?.value,
```
baiirun, line 25: We should add this id to @graphprotocol/grc-20 and import it from the package instead of hardcoding it.

apps/web/partials/entity-page/number-options-dropdown.tsx
```diff
@@ -143,6 +134,109 @@ const NumberFormatView = ({ formatOptions, value }: { formatOptions?: NumberForm
...
+  React.useEffect(() => {
+    setFilteredEntities(
+      entities.map(entity => ({
+        name: entity.name,
+        symbol: entity.triples.find(t => t.attributeId === 'NMCZXJNatQS59U31sZKmMn')?.value?.value, // Attribute Id for symbol
+        sign: entity.triples.find(t => t.attributeId === 'Tt2mYqE1kJTRLt2iLQjATb')?.value?.value, // Attribute Id for sign
+        id: entity.id,
+      }))
+    );
+  }, [searchResults, entities]);
```
baiirun, line 165: Do we need an effect for this? You should be able to just map it during render

apps/web/partials/entity-page/number-options-dropdown.tsx
```diff
@@ -143,6 +134,109 @@ const NumberFormatView = ({ formatOptions, value }: { formatOptions?: NumberForm
...
+
+  React.useEffect(() => {
+    setInputValue(query);
+  }, [query]);
```
baiirun, line 169: Can't you just use the `query` value directly instead of setting it in separate state?

apps/web/partials/entity-page/number-options-dropdown.tsx
```diff
@@ -143,6 +134,109 @@ const NumberFormatView = ({ formatOptions, value }: { formatOptions?: NumberForm
...
+      },
+      {
+        name: 'United States Dollar',
+        symbol: 'USD',
+        id: '2eGL8drmSYAqLoetcx3yR1',
+      },
+      {
+        name: 'Euro',
+        symbol: 'EUR',
+        id: 'EWCAJP9TQoZ3EhcwyRg7mk',
```
baiirun, line 195: These IDs should go in @graphprotocol/grc-20

baiirun (approved):

PR #1178, opened 2025-04-25
chore: switch to noble md5

nikgraf:
Just realized this is only run in Node. Initially I thought the crypto package is polyfilled. So we could also discard the PR. An alternative would be to move the functionality to the GRC-20 TS lib since I need it in Hypergraph as well.

When I run `pnpm test` in substream the tests fail and the new id.test.ts does not run. No idea why. If I run `pnpm vitest ./sink/utils/id.test.ts` they run fine.


baiirun:
Closing this since we'll be migrating to a new data service in the near future

PR #1184, opened 2025-04-30
feat: implement date ranges

mikallojjus:
Implement date ranges.

Value for range is two ISO strings separated by `/`

image

---

image

---

image

### DIFF
--- a/apps/web/core/utils/utils.ts
+++ b/apps/web/core/utils/utils.ts
@@ -90,6 +90,7 @@ export class GeoNumber {
 
 export class GeoDate {
   static defaultFormat = 'MMM d, yyyy - h:mmaaa';
+  static intervalDelimiter = '/';
 
   /**
    * We return blocktime from the subgraph for createdAt and updatedAt fields.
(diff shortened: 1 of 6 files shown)


apps/web/core/utils/utils.ts
```diff
@@ -207,13 +208,46 @@ export class GeoDate {
...
+
+        const startDate = parseISO(startDateString);
+        const endDate = parseISO(endDateString);
+
+        const formattedStartDate = formatInTimeZone(startDate, 'UTC', validatedFormat);
+        const formattedEndDate = formatInTimeZone(endDate, 'UTC', validatedFormat);
+
+        return `${formattedStartDate} — ${formattedEndDate}`;
+      }
+
```
baiirun, line 246: Can we update the tests to test for this?

apps/web/core/utils/utils.ts
```diff
@@ -207,13 +208,46 @@ export class GeoDate {
...
+    return `${dateIsoString}${this.intervalDelimiter}${dateIsoString}`;
+  };
+
+  static isDateInterval = (dateString?: string): boolean => {
+    if (!dateString) {
+      return false;
+    }
+
+    return dateString.includes(this.intervalDelimiter);
+  };
```
baiirun, line 229: Can we add a test for this?

apps/web/core/utils/utils.ts
mikallojjus: ✅

apps/web/core/utils/utils.ts
mikallojjus: ✅

PR #1198, opened 2025-05-19
fix: prevent collection loading flicker when adding new item

b-d055:


apps/web/core/blocks/data/use-collection.ts
```diff
@@ -63,10 +63,14 @@ export function useCollection() {
     })
     .filter(item => item !== undefined);
 
+  // Don't show loading if we already have data and are just refetching
+  const hasExistingData = orderedCollectionItems.length > 0 || collectionRelations.length > 0;
```
baiirun, line 67: Does this work if the collection is empty to start but then an item is added?

baiirun: We may also be able to use `placeholderData: keepPreviousData` in React Query within `useQueryEntities` and `useQueryEntity`.

baiirun: If `placeholderData` doesn't work another option is to use `isFetched` or one of the other "state" flags from React Query and return it from `useQueryEntities`. This signals that the initial fetch has succeeded and to not go into loading state for future loads.

b-d055: @baiirun we tried that, but the issue is that useQuery returns `isFetched` as true regardless of how placeholderData is set.  It goes to `false` on every fetch (even with `placeholderData`). 

However, having just looked at the docs again it appears the `status` of useQuery can help us determine if the data is cached, so we can try using that instead.

baiirun (approved):

PR #1226, opened 2025-06-19
Add browse property types/renderable types

b-d055:


apps/web/partials/entity-page/data-type-pill.tsx
```diff
@@ -0,0 +1,77 @@
...
+  RELATION: Relation,
+};
+
+export function DataTypePill({ 
+  dataType, 
+  dataTypeId,
+  renderableType, 
+  spaceId 
+}: DataTypePillProps) {
+  // Use dataTypeId if available, otherwise fall back to hard-coded mapping
```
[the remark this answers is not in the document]
b-d055: @baiirun The requested feature is to go to the entity page for the type when you click on the pill. I put this here as a placeholder - how should we be getting the UUID for eachdata type?

apps/web/partials/entity-page/entity-page-metadata-header.tsx
```diff
@@ -41,8 +43,47 @@ export function EntityPageMetadataHeader({ spaceId }: EntityPageMetadataHeaderPr
 
   const typesRenderableObj = typesRenderable.find(r => r?.find(re => re.propertyId === SystemIds.TYPES_PROPERTY));
 
+  // Check if this entity is a Property
+  const isPropertyEntity = types.some(type => type.id === SystemIds.PROPERTY);
```
baiirun, line 47: An entity having types: Property doesn't actually guarantee that it's a property with a data type.

In the new data model we're separating the data type from the property entity. This is to add some property data consistency guarantees to the system. The only way to check if it's a property is to actually query the properties from the db (or the local store if a property was created locally).

Instead we should always query for the propertyData and _that_ is what determines if an entity is a property or not.

So basically just remove `isPropertyEntity` and the enabled flag.

apps/web/partials/entity-page/entity-page-metadata-header.tsx
```diff
@@ -41,8 +43,47 @@ export function EntityPageMetadataHeader({ spaceId }: EntityPageMetadataHeaderPr
...
+    enabled: isPropertyEntity 
+  });
+
+  const propertyDataType = React.useMemo(() => {
+    if (!isPropertyEntity || !propertyData) return null;
+
+
+    // Find the renderable type from entity relations
+    const renderableTypeRelation = relations.find(relation => {
+      return relation.type.name === 'Renderable type'
```
baiirun, line 61: We should use the renderable type id instead of the name. There's a RENDERABLE_TYPE_PROPERTY exports from constants. I'll work on getting it added to SystemIds

apps/web/partials/entity-page/entity-page-metadata-header.tsx
```diff
@@ -41,8 +43,47 @@ export function EntityPageMetadataHeader({ spaceId }: EntityPageMetadataHeaderPr
...
+    });
+
+    return {
+      id: propertyData.id,
+      dataType: propertyData.dataType,
+      dataTypeId: propertyData.dataTypeId,
```
baiirun, line 67: We shouldn't need the data type id for anything since data types are just hardcoded into the system now

apps/web/partials/entity-page/entity-page-metadata-header.tsx
```diff
@@ -41,8 +43,47 @@ export function EntityPageMetadataHeader({ spaceId }: EntityPageMetadataHeaderPr
...
+      renderableType: renderableTypeRelation ? {
+        id: renderableTypeRelation.toEntity.id,
+        name: renderableTypeRelation.toEntity.name as RenderableType,
```
baiirun, line 70: There's no guarantee that the renderable type entity names and how we represent them in the codebase will stay in sync, so this cast seems dangerous. We should probably map the id if the renderable type entity to the string union representation in the app.

Note that this is already broken since `RenderableType` represents strings as uppercase and the entity names aren't.

apps/web/partials/entity-page/readable-entity-page.tsx
```diff
@@ -87,18 +93,32 @@ function ValuesGroup({ entityId, values }: { entityId: string; values: ValueRend
         }
         return (
           <div key={`${entityId}-${t.propertyId}-${index}`} className="break-words">
-            <Text as="p" variant="bodySemibold">
-              {values[0].propertyName || t.propertyId}
-            </Text>
+            <Link href={NavUtils.toEntity(spaceId, attributeId)}>
+              <Text as="p" variant="bodySemibold">
+                {values[0].propertyName || t.propertyId}
```
baiirun, line 98: Can probably extract the propertyName above the mapping like you do for space and id

apps/web/core/hooks/use-property.ts
```diff
@@ -0,0 +1,83 @@
...
+  // Subscribe to sync events for real-time updates
+  useEffect(() => {
+    if (!enabled || !id) return;
+
+    const unsubscribe = stream.on('entity:updated', event => {
+      // Handle entity updates that might affect this property
+      if (event.entity.id === id) {
+        cache.invalidateQueries({ queryKey: ['property', id] });
+      }
+    });
```
baiirun, line 65: This likely won't work since we aren't really using the entity:updated event anywhere when updating data. We probably want to listen for the entities synced event instead. I'll be working on a properties store that handles all this anyway so we can probably just remove this effect for now.

apps/web/core/hooks/use-property.ts
```diff
@@ -0,0 +1,83 @@
...
+      console.log('useProperty queryFn called with id:', id);
+      
+      // Properties are entities, so first check if we have it in the entity store
+      const localEntity = store.getEntity(id);
+      console.log('localEntity found:', !!localEntity);
+      
+      // Always fetch from the server for now to debug
+      try {
+        const result = await Effect.runPromise(getProperty(id, signal));
+        console.log('getProperty result:', result);
```
baiirun, line 38: Should be able to remove the logs in this hook

apps/web/core/hooks/use-property.ts
b-d055: removed pattern

apps/web/partials/entity-page/entity-page-metadata-header.tsx
b-d055: @baiirun for now I've removed the type case and updated logic in the data pill rendering

apps/web/core/hooks/use-property.ts
```diff
@@ -0,0 +1,58 @@
+import { useQuery } from '@tanstack/react-query';
```
baiirun, line 1: Do you still need this hook now that we have the `useQueryProperty` hook from `use-store`?

apps/web/partials/entity-page/entity-page-metadata-header.tsx
```diff
@@ -37,9 +40,43 @@ export function EntityPageMetadataHeader({ spaceId }: EntityPageMetadataHeaderPr
...
+
+    return {
+      id: propertyData.id,
+      dataType: propertyData.dataType,
+      renderableType: renderableTypeRelation ? {
+        id: renderableTypeRelation.toEntity.id,
+        name: renderableTypeRelation.toEntity.name
+      } : null,
+    };
+  }, [propertyData, relations]);
```
baiirun, line 67: The `useQueryProperty` hook should already return the renderable type

b-d055: Ah.. yeah well that simplifies things a lot - updated to use new `useQueryProperty`

apps/web/core/blocks/data/data-selectors.ts
```diff
@@ -172,7 +172,7 @@ export function generateSelector(
     }
   }
 
-  if (property.renderableType === 'RELATION' || property.renderableType === 'IMAGE') {
+  if (property.renderableType === SystemIds.RELATION_TYPE || property.renderableType === SystemIds.IMAGE_TYPE) {
```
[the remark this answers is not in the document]
b-d055: I actually think this specifically just shouldn't have been changed. It's named `renderableType` here but if we look upstream, `property`'s type actually comes from `RenderableProperty['type']` which is what we are now calling dataType, not renderableType - but let's connect to discuss your larger point which is still valid.

baiirun (approved):

PR #1235, opened 2025-06-26
feat: backlinks

sandren:


baiirun (commented):
Couple of things
1. When a page doesn't have any references we shouldn't show the header. That's the behavior in prod atm
2. In prod we fetch the backlinks in a separate query on purpose. The query can be pretty slow, and we don't want to block rendering the main data. We should put the backlinks in its own query and suspend it like we do in prod.

apps/web/app/space/(entity)/[id]/[entityId]/default-entity-page.tsx
```diff
@@ -86,15 +84,9 @@ export default async function DefaultEntityPage({
...
-          <ErrorBoundary fallback={<EmptyErrorComponent />}>
-            {/*
-              Some SEO parsers fail to parse meta tags if there's no fallback in a suspense boundary. We don't want to
-              show any referenced by loading states but do want to stream it in
-            */}
-            {/* <React.Suspense fallback={<div />}>
-              <EntityReferencedByServerContainer entityId={props.id} name={props.name} spaceId={params.id} />
-            </React.Suspense> */}
-          </ErrorBoundary>
+          <React.Suspense fallback={null}>
```
Comment (baiirun) [inline apps/web/app/space/(entity)/[id]/[entityId]/default-entity-page.tsx:87]: We need to double check that SEO properly populates. There were weird issues in the past outlined in the old comment

apps/web/app/space/[id]/page.tsx
```diff
@@ -88,15 +87,6 @@ export default async function SpacePage(props0: Props) {
       </React.Suspense>
       <ToggleEntityPage id={props.id} spaceId={spaceId} values={props.values} />
       <Spacer height={40} />
-      <ErrorBoundary fallback={<EmptyErrorComponent />}>
```
Comment (baiirun) [inline apps/web/app/space/[id]/page.tsx:91]: We don't want Referenced by on space pages?

PR #1251, opened 2025-07-11
fix: backlink types

sandren:


apps/web/partials/entity-page/backlinks.tsx
```diff
@@ -54,8 +55,9 @@ type BacklinkProps = {
 
 const Backlink = ({ backlink }: BacklinkProps) => {
   const { space } = useSpace(backlink.spaceIds[0] ?? '');
+  const entity = useEntity({ id: backlink.id, spaceId: backlink.spaceIds[0] ?? '' });
```
baiirun, line 58: We should just query this in the backlinks query instead of doing a second query. This could kick off dozens of queries on the client if the backlink list is large. Better to do it in one query on the server

baiirun: Should do the same thing for the space as well on the server tbh

baiirun (approved):

PR #1255, opened 2025-07-16
feat: space deployment

sandren:


baiirun:
Is this still needed? Have you checked using the existing logic and just updating `@graphprotocol/grc-20`?

PR #1260, opened 2025-07-17
fix: join project card

sandren:


apps/web/design-system/select-entity.tsx
```diff
@@ -46,7 +46,7 @@ import { showingIdsAtom } from '~/atoms';
 
 type SelectEntityProps = {
   onDone?: (
-    result: { id: string; name: string | null; space?: EntityId; verified?: boolean },
+    result: { id: string; name: string | null; space?: EntityId; primarySpace?: EntityId; verified?: boolean },
```
baiirun, line 49: Should be safe to remove `EntityId` and replace with `string`. We aren't really using `EntityId` consistently anymore and I've been removing its usage.

PR #1267, opened 2025-07-29
feat: edit relation page

sandren:


apps/web/partials/entity-page/entity-page-relations.tsx
```diff
@@ -1,32 +1,59 @@
...
+export const EntityPageRelations = ({ spaceId, entityId, relations: serverRelations }: EntityPageRelationsProps) => {
   const [isOpen, setIsOpen] = useState<boolean>(true);
 
+  const { hydrate } = useSyncEngine();
+
+  const clientRelations = useRelations({
+    selector: r => r.entityId === entityId,
+  });
+
+  const relations = clientRelations.length >= serverRelations.length ? clientRelations : serverRelations;
```
baiirun, line 46: This logic will break if we ever delete a relation locally. It will revert to showing server relations even though it should always be showing the client relations.

baiirun: We should probably always just show client relations after we've hydrated them locally. Or we can just always use the client as the source of truth.

PR #1268, opened 2025-07-30
fix: personal space deployment

sandren:


apps/web/app/api/space/deploy/deploy.ts
```diff
@@ -40,7 +40,7 @@ export function deploySpace(args: DeployArgs) {
       try: () =>
         Graph.createSpace({
           name: args.spaceName,
-          spaceEntityId: args.entityId,
+          spaceEntityId: args.entityId || spaceEntityId,
```
baiirun, line 43: Can we just use `spaceEntityId` since we use `args.entityId` internally in `generateOpsForSpace`?

PR #1271, opened 2025-08-04
Add: edit block space & image upload

b-d055:


apps/web/core/blocks/data/filters.ts
```diff
@@ -268,6 +270,9 @@ async function getResolvedFilter(filter: PropertyFilter): Promise<Filter> {
     columnName: maybePropertyEntity?.name ?? null,
     value: filter.is,
     valueName: maybeValueEntity?.name ?? null,
-    valueType: 'RELATION',
+    valueType:
+      filter.property === SystemIds.NAME_PROPERTY || filter.property === SystemIds.DESCRIPTION_PROPERTY
+        ? 'TEXT'
+        : 'RELATION',
```
baiirun, line 276: I don't think this is correct since we should be able to filter by any data type depending on the property.

apps/web/core/hooks/use-spaces-query.ts
```diff
@@ -1,94 +1,97 @@
...
+                  return {
+                    id: {
+                      equals: t,
+                    },
+                  };
+                }),
+              },
+              first: 10,
+              skip: 0,
+            }),
```
baiirun, line 46: This doesn't filter specifically by space. Fuzzy searches for _any_ entity. If we set the types -> Space then we'll filter by space as well.

apps/web/core/sync/experimental_query-layer.ts
```diff
@@ -284,16 +284,16 @@ export class EntityQuery {
...
-      // if (condition === undefined) {
-      //   return true;
-      // }
-
-      // if (Array.isArray(condition)) {
-      //   const clause = condition as StringCondition[];
-      //   return clause.some(space => space.equals && entity.spaces.includes(space.equals));
-      // }
-      // return false;
+        //return true;
```
baiirun, line 287: Should be able to remove this

apps/web/design-system/editable-fields/editable-fields.tsx
```diff
@@ -63,14 +63,30 @@ type PageStringFieldProps = {
...
 
 export function PageStringField({ ...props }: PageStringFieldProps) {
+  const [localValue, setLocalValue] = React.useState(props.value ?? '');
+  const debouncedValue = useDebouncedValue(localValue, 1500);
+
+  React.useEffect(() => {
+    if (props.shouldDebounce && localValue.length > 0) {
+      props.onChange(localValue);
+    }
+  }, [debouncedValue]);
```
baiirun, line 77: What was the motivation for adding this back?

apps/web/partials/blocks/table/change-entry.ts
```diff
@@ -1,9 +1,27 @@
...
+    };
+    value: { type: RenderableEntityType; value: string };
+  };
+};
 
 type ChangeEntryParams =
   | {
       type: 'EVENT';
-      data: any; // @TODO(migration): Correct type
+      data: EventPayload; // @TODO(migration): Correct type
```
baiirun, line 24: Do we still need the TODO?

apps/web/core/hooks/use-spaces-query.ts
b-d055: @baiirun  as far as I can tell this is filtering down for spaces only. `filterByTypes` only contains the UUID for space type and in the results I'm only seeing spaces, even if I specifically try to find regular entities.

apps/web/design-system/editable-fields/editable-fields.tsx
b-d055: Were having a strange textbox blur issue and thought this might be related - but can remove if you'd like. I just tested without it and it still seems to be working.

apps/web/core/blocks/data/filters.ts
```diff
@@ -246,28 +247,43 @@ async function getResolvedEntityFilter(entityId: string, typeId: string): Promis
...
+    valueType === 'RELATION'
+      ? E.findOne({ store, cache: queryClient, id: filter.is })
+      : undefined,
+  ]);
+  
+  console.log('Resolved filter debug', { 
+    valueName: maybeValueEntity?.name ?? null,
+    property: property,
+    is: filter.is,
+  });
```
baiirun, line 280: Should be able to remove this

PR #1277, opened 2025-08-11
Fix: hide activity tab temporarily

b-d055:


apps/web/app/space/(entity)/[id]/[entityId]/layout.tsx
```diff
@@ -79,6 +80,12 @@ export default async function ProfileLayout(props: Props) {
 
   const profile = await getProfilePage(entityId);
 
+  // Check if entity has an accounts relation (indicating a real Geo user)
+  const hasAccountsRelation = result?.entity?.relations.some(
+    relation => relation.type.id === SystemIds.ACCOUNTS_PROPERTY
+  );
+  const isGeoUser = hasAccountsRelation;
```
Comment (baiirun) [inline apps/web/app/space/(entity)/[id]/[entityId]/layout.tsx:87]: Does this actually indicate a real user? Couldn't anybody just add an account relation?

Comment (b-d055) [reply apps/web/app/space/(entity)/[id]/[entityId]/layout.tsx:87]: @baiirun  the comment here is presumptuous, but the actual linear ticket states 

> This should be changed so that the activity tab only shows when a person entity has an account linked to it that can actually make edits in geo. 

How else do we determine if the linked account can make edits?

baiirun (approved):

PR #1278, opened 2025-08-13
feat: property format suggestions

b-d055:


apps/web/core/constants.ts
```diff
@@ -13,3 +13,48 @@ export const ROOT_SPACE = '2df11968-9d1c-489f-91b7-bdc88b472161';
...
 export const FORMAT_PROPERTY = '396f8c72-dfd0-4b57-91ea-09c1b9321b2f';
+export const UNIT_PROPERTY = '11b06581-20d3-41ea-b570-2ef4ee0a4ffd';
+export const DEFAULT_TIME_FORMAT = 'MMMM d, yyy';
+export const DEFAULT_NUMBER_FORMAT = ',?';
+
+export const UNICODE_LINK = 'https://unicode-org.github.io/icu/userguide/format_parse/numbers/skeletons.html';
+export const GRC_20_SPECIFICATION_LINK =
+  'https://github.com/graphprotocol/graph-improvement-proposals/blob/main/grcs/0020-knowledge-graph.md#52-number';
+
+export const SUGGESTED_NUMBER_FORMATS = [
```
b-d055, line 24: @baiirun do you think these are things that could live in geo root as opposed to being hardcoded or nah?

baiirun: Yeah this is more of an application-specific thing so I think it's fine to hardcore

PR #1298, opened 2025-09-30
Fix: broken links when http protocol not specified

b-d055:


apps/web/design-system/editable-fields/web-url-field.tsx
```diff
@@ -56,6 +56,8 @@ export function WebUrlField({
     return <GraphUrlField currentSpaceId={spaceId} value={value as `graph://${string}`} />;
   }
 
+  const normalizedUrl = value && !value.match(/^[a-zA-Z][a-zA-Z\d+\-.]*:/) ? `https://${value}` : value;
```
sandren, line 59: Let's add a comment about that regex (and do we really want to support non-http protocols here?).

sandren (changes requested):

PR #1331, opened 2025-11-25
Feat: pinata images

b-d055:


apps/web/core/utils/utils.ts
```diff
@@ -400,6 +400,17 @@ export const getImagePath = (value: string) => {
   }
 };
 
+// Get the fallback image URL (Lighthouse gateway) for when Pinata fails
+export const getImagePathFallback = (value: string) => {
```
baiirun: Should we just have a single `getImagePath` function that automatically calls this?

Could just put this at the bottom of the existing `getImagePath` function. Then we don't need the fallback state in `GeoImage` at all AFAICT.

baiirun (approved):

PR #1555, opened 2026-03-10
fix: remove the re-order funcationality from collection data blocks

b-d055:


ohohoreilly:

Opened **2026-03-10** and now conflicting with master. At this distance a rebase is usually more work than redoing the change against current code, and the surrounding code has moved a long way underneath *"fix: remove the re-order funcationality from collection data blocks"*.

@b-d055 — if the idea still stands but the branch does not, a fresh PR or a ticket is probably a better route than reviving this one.

PR #1639, opened 2026-04-03
Feat/update filter functionality

webmagic123:


apps/web/core/blocks/data/use-data-block.tsx
```diff
@@ -131,6 +135,48 @@ export function useDataBlock(options?: UseDataBlockOptions) {
     sort: serverSort,
   });
 
+  const collectionSuggestionIdSlice = React.useMemo(() => {
+    if (source.type !== 'COLLECTION' || !collectionFilterSuggestionEntityIds?.length) return undefined;
+    return collectionFilterSuggestionEntityIds.slice(0, FILTER_SUGGESTION_ENTITY_CAP);
+  }, [source.type, collectionFilterSuggestionEntityIds]);
+
+  useQueryEntities({
```
b-d055: @webmagic123 these two queries seem expensive (they fetch up to 5000 entities), can we disable them until the filter popover is actually open? That way we aren't sending these requests unnecessarily. I think we already track the popover ref, table-block, we can use something like `const [filterOpen, setFilterOpen] = useState(false)`

PR #1664, opened 2026-04-10
fix: convert commenting functionality to use the SDK in the Web app

webmagic123:
Feature:
 Convert the web app's commenting (create + edit) to use the Geo SDK for generating graph operations

Working Flow:
  Before: creating a comment meant manually assembling every piece.
       Publish.prepareLocalDataForPublishing to get ops.
  After: that whole block is replaced with two SDK calls:
  - `Graph.createComment({ id, content, replyTo, resolved })` -> returns the ops directly.
  - `Graph.updateComment({ id, content })` -> same, for edits.

Fixed:
  - Kept master's comments-section.tsx and its ancestorComments threading(mapped that to the SDK's single replyTo (the immediate parent) inside the hook).
  - Took master's geo-sdk 0.20.3 (which provides the Graph.createComment/updateComment APIs) and its lockfile.
  - Switched to GEO_NETWORK.id once the SDK supports mainnet.


apps/web/core/hooks/use-create-comment.tsx
```diff
@@ -229,112 +218,40 @@ export function useCreateComment(targetEntityId: string) {
...
+        let ops: Op[];
+        try {
+          const result = await Graph.createComment({
+            id: commentEntityId,
+            content: text,
+            replyTo: replyToTarget,
+            resolved: false,
+            // The SDK's `Network` type only permits 'TESTNET' as of geo-sdk 0.20.3; switch to
+            // GEO_NETWORK.id once the SDK widens it to include mainnet.
+            network: 'TESTNET',
```
b-d055: `network: 'TESTNET'` routes this request to the wrong host on any non-testnet build.

`Graph.createComment` is not a pure op builder. It calls `resolveGeoNetwork(network)` and issues a GraphQL request to that network's `apiOrigin` (`fetchReplyToRelations` in `dist/src/client/comments.js:5`, called at `:70`). `'TESTNET'` resolves to the built-in `https://api-testnet.geobrowser.io` (`dist/src/networks.js:2`), ignoring `GEO_NETWORK.apiOrigin`, which is derived from `NEXT_PUBLIC_API_ENDPOINT` (`core/sdk/geo-network.ts:50`).

`core/sdk/geo-network.ts:14-15` states the rule:

> Nothing outside this module may hardcode a chain id, a contract address, or a `network: 'TESTNET'` literal.

Effect on a build pointed at the v2 API or mainnet: the parent comment does not exist on the host being queried, `fetchReplyToRelations` returns `[]`, and the reply is written with no reply-to relation to the target entity. `fetchCommentEntitiesForTarget` (`core/hooks/use-comments.ts:154-156`) keeps only entities that have one, so nested replies do not appear in the thread. The relations are on-chain, so re-indexing does not recover them. `core/sdk/geo-client.ts:29-34` records the same issue previously affecting image uploads.

On the code comment above: the type claim is correct. `CreateCommentParams.network` is `Network` (`= 'TESTNET'`), not `Networkish`, so `network: GEO_NETWORK` does not compile against `Graph.createComment`. An SDK change is not required to fix this. Two options:

1. `Ops.comments.create`, which also addresses the comment on line 224. It builds the same ops with no network call.
2. `geo.comments.create`. This file already imports a client bound to the correct network on line 14 (`import { geo } from '~/core/sdk/geo-client'`, which is `createGeoClient({ network: GEO_NETWORK })`). It takes its origin from the client context:

```ts
const result = await geo.comments.create({
  id: commentEntityId,
  content: text,
  replyTo: replyToTarget,
  resolved: false,
});
```

`Graph.createComment` is marked `@deprecated` and its note points at both of these.

apps/web/core/hooks/use-create-comment.tsx
```diff
@@ -229,112 +218,40 @@ export function useCreateComment(targetEntityId: string) {
...
-              toEntity: { id: ancestor.id, name: null, value: ancestor.id },
-              toSpaceId: ancestor.spaceId,
-              isLocal: true,
-              hasBeenPublished: false,
-            });
-          }
+        // The SDK builds the reply chain from the immediate parent; default to the entity
+        // being commented on for a top-level comment. `ancestorComments` is ordered
+        // [immediate parent, ..., root], so index 0 is the parent to reply to.
+        const immediateParent = ancestorComments?.[0];
```
b-d055, line 224: Keeping only `ancestorComments[0]` drops the reply-to chain when the parent is not yet indexed.

The caller passes the full chain: `comments-section.tsx:991` builds `[{ id: comment.id, spaceId: comment.spaceId }, ...ancestors]`. The replaced code emitted a reply-to relation for the target entity plus one per ancestor. Now only the immediate parent is used and the remainder is re-derived by `Graph.createComment` from the indexer.

This races the hook's own optimistic row. The row is inserted immediately and reconciles only once the poller sees the comment (`FIRST_POLL_MS` plus up to 45 x 2s, so roughly 91s worst case). Nothing in `comments-section.tsx` disables Reply while a row is pending; `isPublishing` only drives the "Publishing" label at `:1070` and `:1255`.

Repro: post comment A, then reply to A while A still reads "Publishing". `fetchReplyToRelations(A)` returns `[]`, so the reply gets a single reply-to to A and no relation to the target entity, and `fetchCommentEntitiesForTarget` drops it once the optimistic row reconciles.

The chain is already available here. `Ops.comments.create` accepts it directly (`dist/src/ops/comments.d.ts`):

```ts
import { IdUtils, Ops, type Op } from '@geoprotocol/geo-sdk/lite';

const ancestors = ancestorComments ?? [];
const immediateParent = ancestors[0];
const replyToTarget = immediateParent
  ? { entityId: immediateParent.id, spaceId: immediateParent.spaceId }
  : { entityId: targetEntityId, spaceId: targetSpaceId };

const { ops } = Ops.comments.create({
  id: commentEntityId,
  content: text,
  replyTo: replyToTarget,
  resolved: false,
  // The remaining ancestors, then the entity being commented on. Empty for a
  // top-level comment, whose replyTo is already the target entity.
  replyToRelations: immediateParent
    ? [...ancestors.slice(1), { id: targetEntityId, spaceId: targetSpaceId }].map(a => ({
        entityId: a.id,
        spaceId: a.spaceId,
        position: null,
      }))
    : [],
});
```

Two notes on that:

- The target-entity relation has to be in the list. It is the only relation `fetchCommentEntitiesForTarget` filters on.
- `position: null` is safe. `use-comments.ts:70-77` selects the parent by depth rather than by relation position, so only the set of relations affects rendering. The SDK generates ascending positions for the ops regardless (`ops/comments.js`).

This also makes op construction synchronous, which removes the `try`/`catch` at `:242` and its optimistic-row rollback.

apps/web/core/hooks/use-create-comment.tsx
```diff
@@ -229,112 +218,40 @@ export function useCreateComment(targetEntityId: string) {
...
+        const replyToTarget = immediateParent
+          ? { entityId: immediateParent.id, spaceId: immediateParent.spaceId }
+          : { entityId: targetEntityId, spaceId: targetSpaceId };
+
+        // Generate the comment ops through the SDK instead of hand-building values/relations.
+        let ops: Op[];
+        try {
+          const result = await Graph.createComment({
```
b-d055: Comment creation now depends on a network round trip that has no timeout. Mostly resolved by moving to `Ops.comments.create`; noting it so the tradeoff is explicit.

`Graph.createComment` awaits `graphqlData`, which throws `GeoApiError` on a fetch failure, a non-2xx response, a GraphQL `errors` array, or an unparseable body (`dist/src/client/api.js:60-105`). The `catch` at `:242` then deletes the optimistic row and returns `null`, so a transient failure or an offline moment blocks commenting. Op construction was previously local.

It also passes `globalThis.fetch` with no `AbortSignal` (`dist/src/graph/create-comment.js:9`), so the request has no timeout and is not cancelled on unmount. Every other fetch in this hook threads `signal` through.

For a top-level comment the request returns nothing useful, since the page entity has no reply-to relations.

apps/web/core/hooks/use-create-comment.tsx
```diff
@@ -531,41 +448,15 @@ export function useCreateComment(targetEntityId: string) {
 
       try {
         const newName = getCommentName(newText);
```
b-d055, line 442: `getCommentName` and the SDK's `deriveCommentName` both run now, and they produce different results.

`Graph.createComment` and `Graph.updateComment` derive the published Name themselves (`ops/comments.js`, via `deriveCommentName`). The local `getCommentName` at `:30` now only feeds the optimistic row and the edit title (`:159`, `:450`). The rules differ:

| | local `getCommentName` | SDK `deriveCommentName` |
|---|---|---|
| strips | character class `` [#*_~`>\[\]()!] `` | markdown constructs (`**bold**`, `[t](u)`, `---`, list markers) |
| over 20 chars | `slice(0, 20).trimEnd() + '...'` | `slice(0, 20)`, no ellipsis |
| empty result | `'Comment'` | `''` |

Two consequences:

- A comment over 20 characters renders optimistically with a trailing `...` that the published Name does not have.
- For input that the SDK strips to empty but the local rule does not (`---`, `# `, `> `, `a`), `createEntity` skips the Name value entirely (`dist/src/graph/create-entity.js:123`, `if (name)`). The comment publishes with no Name while the optimistic row showed the raw text.

Deriving the optimistic name the same way the ops do would keep these in sync.

b-d055 (changes requested):
@webmagic123  Two blocking issues in use-create-comment.tsx, both within the same 20 lines, plus one low-severity consistency note.

Both blockers are resolved by the same change: build the ops locally with Ops.comments.create instead of Graph.createComment. The reply chain the SDK re-fetches from the indexer is already passed into this hook by the caller, so the round trip adds a network dependency and two failure modes for data that is already in hand.

Details inline.

b-d055 (approved):

PR #1748, opened 2026-05-01
feat(data-blocks): per-property AND/OR mode + in-operator collapse

jwalkingjew:
## Summary
- Replaces the global filter `mode` with per-property modes (one AND/OR per column).
- Collapses multi-value selections into single GraphQL `in` clauses: RELATION → `toEntity.id.in`, TEXT → `text.inInsensitive`, numeric/date filters route to the right scalar via a new `dataType` discriminator on `ValueCondition`.
- Adds `dataType`-aware scalar routing in the converter so future numeric / date / datetime UI lands without converter changes.
- Migrates old persisted filters in-memory on read (legacy `mode: 'OR'` fans out to every property group); next save writes the new shape only — no compat shim long-term.

## Why
- The global `'AND' | 'OR'` toggle was the wrong granularity. Users want "type is Person OR Author, AND status is Active" — per-property OR with implicit AND between columns. The global mode couldn't express that.
- N values for the same property used to expand to N nested `OR` clauses on the wire (each its own `relations.some {...}` / `values.some {...}`). The schema has `in` everywhere relevant, so a multi-value selection collapses to one clause — smaller payload, simpler filter, less work on the API.

## Architecture notes
- New `buildGroupWhere(filters, groupMode)` in `use-data-block.tsx` replaces the prior `buildSingleFilterWhere` / `buildOrWhere` / `buildAndWhere` trio.
- `WhereCondition` (DSL) gains `ValueCondition.dataType` and `NumberCondition.in`. The local `EntityQuery` engine doesn't need updates for the dataType hint — the converter is the only consumer that uses it.
- Persisted format adds `modes: Record<columnId, 'AND' | 'OR'>` and stops emitting the legacy top-level `mode`. Schema keeps `mode` optional for read-time compat.
- The per-group default mode is `'OR'` (collapses cleanly to `in`); the persisted `modes` map only stores entries that diverge from default — keeps wire noise down.

## Out of scope (separate PRs)
- **Backlinks in the filter creation UI.** The engine handles them already (`isBacklink` round-trips through persistence + WHERE-builder + converter), but the picker has no path to add one. Tracked as PR B.
- **Multi-value chips for numeric / date / boolean.** Converter routes by dataType when given multi-value, but the UI doesn't surface those pickers today. Future PR adds the pickers; the engine is ready.
- **Per-group `AND` for spaces.** Kept OR-only (matches the top-level `spaceIds.in` API shape).

## Test plan
- [ ] **Single relation column, multi-value, OR mode (default).** Network tab: filter payload uses `relations.some.toEntity.id.in: [...]` (one clause), not nested OR.
- [ ] **Same column, click toggle to AND.** Payload becomes `and: [{...}, {...}]` (one per chip). Results narrow to entities matching all values.
- [ ] **Two different columns, both multi-value.** Each column collapses to its own `in` clause; the two clauses are AND'd between groups regardless of per-group toggle.
- [ ] **TEXT column, multi-value.** No mode toggle visible. Payload uses `text.inInsensitive: [...]`.
- [ ] **Open a data block in legacy format** (persisted with `mode: 'OR'`, multiple property groups). Renders correctly. Edit any chip → save. Reopen the entity values raw and confirm the persisted filter string now uses `modes` and the legacy `mode` key is gone.
- [ ] **Add → remove a relation chip in a group, then add again.** No stale `modes[columnId]` entries persist when the group goes empty.
- [ ] **Power tools panel + non-edit (temporary) filters.** Per-group toggle still works in the temporary state.
- [ ] **Chat tool flow:** issue a `setDataBlockFilters` write with `modesByColumn: { typeId: 'AND' }` → verify the persisted filter has the right `modes`.

### DIFF
--- a/apps/web/app/api/chat/tools/write/data-block.test.ts
+++ b/apps/web/app/api/chat/tools/write/data-block.test.ts
@@ -176,7 +176,7 @@ describe('setDataBlockFilters', () => {
         kind: 'setDataBlockFilters',
         blockId: BLOCK,
         spaceId: SPACE,
-        mode: 'AND',
+        modesByColumn: {},
         filters: [
           expect.objectContaining({
             columnId: SystemIds.TYPES_PROPERTY,
--- a/apps/web/app/api/chat/tools/write/data-block.ts
+++ b/apps/web/app/api/chat/tools/write/data-block.ts
@@ -51,7 +51,15 @@ type SetDataBlockFiltersInput = {
   parentEntityId: string;
   spaceId: string;
   filters: FilterInput[];
-  mode?: FilterMode;
+  /**
+   * Per-property mode keyed by columnId. Default for any missing column is
+   * `'OR'` — multiple values for the same property collapse into one
+   * `in`/`inInsensitive` clause. Pass `'AND'` for a column when the user
+   * wants intersection semantics (e.g. "tagged with both Sports and
+   * Politics"). Only relation-typed properties really need AND; values OR
+   * by default.
+   */
+  modesByColumn?: Record<string, FilterMode>;
 };
 
 export function buildSetDataBlockFiltersTool(context: WriteContext) {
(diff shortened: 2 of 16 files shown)


ohohoreilly:

Opened **2026-05-01** and now conflicting with master. At this distance a rebase is usually more work than redoing the change against current code, and the surrounding code has moved a long way underneath *"feat(data-blocks): per-property AND/OR mode + in-operator collapse"*.

@jwalkingjew — if the idea still stands but the branch does not, a fresh PR or a ticket is probably a better route than reviving this one.

PR #1751, opened 2026-05-01
fix(space-members): seed popover/dialog count from SSR to avoid 0-flash

jwalkingjew:
## Summary

Follow-up to #1749. The members/editors popovers (and the Manage… dialogs) briefly rendered \"0 members\" / \"0 editors\" on open before the first paginated API page resolved, even though the chip already had the correct number from SSR.

The chip reads `space.totalMembers` / `space.totalEditors` server-side (added in #1733). This change threads those same numbers into the popover/dialog as `initialTotalMembers` / `initialTotalEditors`, and the shared `useSpaceParticipantsInfinite` hook uses them as the fallback until the live API count arrives.

```ts
const totalCount = query.data?.pages[0]?.totalCount ?? initialTotalCount ?? 0;
```

So the count is correct on first paint and only ever \"refreshes\" if the SSR value and the live API value disagree (which they shouldn't — both come from the same GraphQL `totalCount`).

## Test plan

- [ ] Open the AI space members popover: footer reads `231 members` immediately, no flash to 0.
- [ ] Open \"Manage members\" on the AI space: heading reads `231 members` immediately.
- [ ] Same for editors popover and \"Manage editors\".
- [ ] Smaller space (≤100 members): no regressions.

### DIFF
--- a/apps/web/core/space-members/use-space-participants-infinite.ts
+++ b/apps/web/core/space-members/use-space-participants-infinite.ts
@@ -45,6 +45,12 @@ type UseSpaceParticipantsInfiniteArgs = {
   kind: ParticipantKind;
   enabled?: boolean;
   pageSize?: number;
+  /**
+   * Server-known total to display before the first page resolves. Avoids the
+   * "0 members → 231 members" flash on popover open. The chip already has
+   * this number from the SSR `space.totalMembers` / `space.totalEditors`.
+   */
+  initialTotalCount?: number;
 };
 
 export function useSpaceParticipantsInfinite({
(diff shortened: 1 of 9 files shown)


ohohoreilly:

Opened **2026-05-01** and now conflicting with master. At this distance a rebase is usually more work than redoing the change against current code, and the surrounding code has moved a long way underneath *"fix(space-members): seed popover/dialog count from SSR to avoid 0-flash"*.

@jwalkingjew — if the idea still stands but the branch does not, a fresh PR or a ticket is probably a better route than reviving this one.

PR #1762, opened 2026-05-07
[codex] Include member and editor spaces in search

jwalkingjew:
## Summary

Adds accepted member/editor spaces from the browse sidebar query cache to the global REST search additional_space_ids list.

This keeps search scoped to root/current/personal spaces plus spaces the user is already a member or editor of, without issuing a separate lookup from the search hook. Pending membership/editorship rows are intentionally excluded.

## Validation

- bun run test core/io/queries.test.ts
- bunx eslint core/hooks/use-global-search-space-ids.ts


jwalkingjew:
Parking this for now. The current API only supports a maximum of 10 additional_space_ids, so we’ll come back to the member/editor-space expansion after the API limitation is resolved.

PR #1764, opened 2026-05-07
[codex] Show initial search results with pagination

jwalkingjew:
## Summary

- Switch the shared entity and space search hooks to paginated infinite queries that can intentionally run with an empty query when a search window is open.
- Open search popovers/dialog results on focus so users see initial options before typing, then refine the same result set as they type.
- Add scroll-to-load-more behavior across the global search dialog, entity/property/relation selection, compact search, find entity, autocomplete, and advanced type/space filters.

## Validation

- ../../node_modules/.bin/eslint core/hooks/use-search.ts core/hooks/use-spaces-query.ts design-system/select-entity.tsx design-system/select-entity-compact.tsx design-system/find-entity.tsx design-system/autocomplete/entity-search-autocomplete.tsx partials/search/dialog.tsx partials/blocks/table/table-block-filter-creation-prompt.tsx
- bun run test core/io/queries.test.ts
- git diff --check

## Notes

I also attempted bun run build locally, but this worktree's available dependency install is missing several existing project packages such as ai, @ai-sdk/*, @upstash/*, and @tiptap/extension-underline, so the build could not get to this change.


jwalkingjew:
Replaced by the split search PR stack:

1. #1767 Dedupe search result type tags
2. #1768 Wait for relation target search types
3. #1769 Derive real entity spaces
4. #1770 Use entity-wide spaces in search results
5. #1771 Use top search space name
6. #1772 Show initial search results
7. #1773 Add search dropdown infinite scroll
8. #1774 Filter final search results

PR #1773, opened 2026-05-08
[codex] Add search dropdown infinite scroll

jwalkingjew:
## Summary
- Convert search and space search hooks to paginated infinite queries.
- Expose next-page state and fetch functions to dropdown surfaces.
- Add scroll-near-bottom loading behavior across search dropdowns.

## Validation
- eslint touched infinite-scroll search files


b-d055 (commented):
@jwalkingjew  I think all of co-pilots suggestions here are valid and make this code better overall, if they can get fixed/tested quickly I'd recommend implementing themall

PR #1777, opened 2026-05-08
fix(data-blocks): promote spaceIds/typeIds buried in empty-name and-wrap

jwalkingjew:
## Summary

A data block scoped to one space (no other filters, no type) was taking forever to load. Two compounding causes:

1. `convertWhereConditionToEntityFilter` AND-wraps the produced filter with the empty-name exclusion (`{ name: { isNull: false, isNot: '' } }`), burying any `spaceIds` clause one level deep.
2. The extract helpers in `getAllEntities` (`extractSingleSpaceIdFromFilter`, `extractSpaceIdsFromFilter`, and the `typeIds` equivalents) only inspected the top of the `EntityFilter`. So `spaceIds` was never lifted to the top-level `entitiesConnection` arg, which:
   - dropped the indexed top-level scan path on `entitiesConnection`
   - left `$spaceId` null in the variables, so the per-row `valuesList(filter: { spaceId: { is: $spaceId } })` and `relationsList(...)` filters were unconstrained — pulling values/relations from every space each row participated in

That second effect amplified the recent `valuesList`/`relationsList` `first: 1000` bump (#1758) into the "FOREVER to load" symptom on space-only data blocks.

### Fix

Teach the space-filter and type-filter extract/remove helpers to peek one level into a top-level `and` array (only `and` — `or` semantics differ), and to collapse trivially-empty shells (`and: []`, `and: [singleton]`) when stripping. Covers the wrap shape produced by the converter and `filterStateToWhere`'s `mergeWhereConditions` output.

### Before vs after (variables for a one-space data block)

Before:
```json
{
  "spaceId": null,
  "filter": {
    "and": [
      { "spaceIds": { "in": ["bf7a..."] } },
      { "name": { "isNull": false, "isNot": "" } }
    ]
  }
}
```

After:
```json
{
  "spaceId": "bf7a...",
  "filter": { "name": { "isNull": false, "isNot": "" } }
}
```

## Test plan

- [ ] CI: new unit tests in `apps/web/core/io/space-filter.test.ts` and `apps/web/core/io/type-filter.test.ts` cover the and-wrap extraction + collapsing-on-removal cases.
- [ ] Smoke: open a data block scoped to a single space with no other filters; confirm load time matches pre-#1758 expectations.
- [ ] Smoke: data block with space + type filter still scopes correctly (both lifted to top level).
- [ ] Smoke: multi-space data block (multi-element `in`) still works.
- [ ] Regression check: data block that filters by `name` (where the and-wrap might have a name collision) still resolves correctly without losing the user filter — covered by the collision branch in `removeSpaceIdsFromFilter` that keeps the and-wrap when keys would clash.

## Followups (not in this PR)

The `valuesList`/`relationsList` `first: 1000` per-row over-fetch is still wasteful for data blocks that only render a small set of `shownColumnIds`. Worth a separate PR: either parameterize the page sizes per consumer, introduce a slim data-block-only document, or drop the nested `toEntity.valuesList` from the data-block path.

### DIFF
--- a/apps/web/core/io/space-filter.test.ts
+++ b/apps/web/core/io/space-filter.test.ts
@@ -183,4 +183,35 @@ describe('removeSpaceIdsFromFilter', () => {
       and: [{ name: { isNull: false, isNot: '' } }, { id: { is: 'entity-123' } }],
     });
   });
+
+  it('strips only the first and-child spaceIds when multiple are present', () => {
+    // Mirrors extractor's first-wins semantics: only the clause that gets
+    // promoted is removed. Later spaceIds clauses represent independent
+    // constraints and must survive on the residual filter so the server
+    // still enforces them. A previous version stripped every match,
+    // silently broadening results.
+    const filter: EntityFilter = {
+      and: [
+        { spaceIds: { in: ['space-abc'] } },
+        { spaceIds: { in: ['space-def'] } },
+        { name: { isNull: false, isNot: '' } },
+      ],
+    };
+    expect(removeSpaceIdsFromFilter(filter)).toEqual({
+      and: [{ spaceIds: { in: ['space-def'] } }, { name: { isNull: false, isNot: '' } }],
+    });
+  });
+
+  it('leaves and-children intact when extraction took the top-level spaceIds', () => {
+    // Top-level wins during extraction, so the and-array is left untouched
+    // — any nested spaceIds is an independent constraint, not a duplicate
+    // of the promoted clause.
+    const filter: EntityFilter = {
+      spaceIds: { in: ['space-abc'] },
+      and: [{ spaceIds: { in: ['space-def'] } }, { name: { isNull: false, isNot: '' } }],
+    };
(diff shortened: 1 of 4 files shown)


jwalkingjew:
Thanks for moving this to ready for review. The focused filter tests pass locally, and the intended empty-name `and` wrapper case looks good to me.

One remaining concern before merge: the new extraction helpers only promote the first `spaceIds` / `typeIds` clause they find in a top-level `and`, but the corresponding remove helpers strip every child in that `and` that has `spaceIds` / `typeIds`. If a caller ever passes a shape like:

```ts
{ and: [{ spaceIds: A }, { spaceIds: B }, { name: ... }] }
```

then only `A` gets promoted to the top-level query arg, while both `A` and `B` are removed from the residual filter, which would broaden results. Same issue applies to `typeIds`.

I think this is fixable either by making extraction/removal handle all matching top-level `and` children consistently, or by only stripping the exact clause that was promoted and leaving additional matching clauses in the filter. A regression test for multiple matching children would make me comfortable merging.

PR #1781, opened 2026-05-08
feat: implement slide panel of entity

webmagic123:


apps/web/partials/blocks/table/table-block.tsx
```diff
@@ -265,6 +276,11 @@ function useEntries(
     setEditable(true);
     setHasPlaceholderRow(true);
     setPlaceholderFocusKey(k => k + 1);
+
+    if (source.type !== 'COLLECTION') {
+      createEntityWithTypes({ filters: filterState, advanceNextId: false });
```
b-d055, line 280: @webmagic123  when I test this in the browser, if I click on the "+" for a non-collection I keep seeing the same entity there, it wont let me create a new one, this is a bug I believe caused by the false here

apps/web/partials/entity-page/backlinks-client-container.tsx
```diff
@@ -0,0 +1,21 @@
...
+  entityId: string;
+};
+
+export function BacklinksClientContainer({ entityId }: BacklinksClientContainerProps) {
+  const { data } = useQuery({
+    queryKey: ['entity-backlinks', entityId],
+    queryFn: () => fetchEntityBacklinksPayload(entityId),
+  });
+
+  return <Backlinks backlinks={data ?? []} />;
```
b-d055: @webmagic123  this looks like a other bug, it's always returning "Referenced by" in the side panel, even if there are no backlinks

apps/web/partials/entity-page/entity-side-panel.tsx
```diff
@@ -0,0 +1,515 @@
...
+
+function EntitySidePanelModeToggle() {
+  const panelCtx = React.useContext(EntitySidePanelEditContext);
+  const controls = useAnimation();
+  const { smartAccount } = useSmartAccount();
+  const isLoggedIn = Boolean(smartAccount?.account.address);
+
+  const { canEdit: canEditSpace } = useAccessControl(panelCtx?.spaceId ?? '');
+
+  if (!panelCtx || !isLoggedIn) {
```
b-d055: @webmagic123  I haven't been able to replicate, but `EntitySidePanelModeToggle` returns early for `!panelCtx || !isLoggedIn` before calling `React.useCallback`. If `smartAccount` changes while the side panel is mounted, this component can render a different hook count. Consider moving the callback hook above this early return or make `onToggle` a plain function.

apps/web/partials/blocks/table/table-block.tsx
b-d055, line 280: @webmagic123  additionally, we previously only created in memory placeholders, now this will actually create new entities that may never be used if they don't get a name/tye/etc

b-d055 (changes requested):

apps/web/partials/blocks/table/table-block.tsx
webmagic123: @b-d055 
The issue has been fixed in https://github.com/EE-Solutions/geo-project-issues/issues/139

apps/web/partials/entity-page/backlinks-client-container.tsx
webmagic123: @b-d055 
Fixed.

apps/web/partials/entity-page/entity-side-panel.tsx
webmagic123: @b-d055 
Fixed.

apps/web/partials/blocks/table/table-block.tsx
webmagic123: @b-d055 
That issue has been fixed in https://github.com/EE-Solutions/geo-project-issues/issues/139

b-d055 (commented):
## Blocking: Cmd/Ctrl/Shift/middle-click broken on relation chips while the side panel is open

`apps/web/design-system/chip.tsx` — both `handleEntityLinkClick` and `handleRelationEntityLinkClick` unconditionally call `event.preventDefault()` and `event.stopPropagation()` whenever the panel is open:

```tsx
const handleEntityLinkClick = React.useCallback(
  (event: React.MouseEvent<HTMLAnchorElement>) => {
    if (!isSidePanelOpen) return;
    event.preventDefault();
    event.stopPropagation();
    openInSidePanel(entityId, targetSpaceId);
  },
  [entityId, isSidePanelOpen, openInSidePanel, targetSpaceId]
);
```

These are attached to real `next/link` anchors via `PrefetchLink`, so modifier-clicks the browser would normally turn into "open in new tab" / "open in background tab" / "open in new window" are swallowed and instead just swap the panel's entity. Regression vs. the current main-page behavior.

The repo already has the correct guard in `apps/web/partials/power-tools/power-tools-table.tsx` — please mirror it at the top of both handlers:

```ts
if (event.metaKey || event.ctrlKey || event.shiftKey || event.altKey || event.button === 1) return;
```

### Repro
1. Open a data block on any entity, click the "Open in side panel" affordance on a row so the panel opens.
2. In the panel (or on the page underneath), find a relation chip.
3. Cmd-click (Mac) / Ctrl-click (Linux/Win) the chip's main label — expected: new tab; actual: panel swaps entity.
4. Middle-click the chip — same regression.
5. Shift-click the chip — same regression (should open in new window).
6. Repeat on the relation-entity sub-link (the dots area) to confirm `handleRelationEntityLinkClick` has the same bug.

b-d055:
@webmagic123  Two housekeeping items before this is ready to merge:

1. **Conflicts**: GitHub is reporting this branch as `CONFLICTING` against `master`. Please rebase / merge `master` in and resolve.
2. **Format**: please run `bun run format` before pushing. Spot-checked a few files in the diff that look unformatted

b-d055 (approved):

PR #1797, opened 2026-05-13
Fix/deterministic space for entity creation

webmagic123:


apps/web/core/sync/orm.ts
```diff
@@ -112,9 +116,18 @@ export function mergeRelations(localRelations: Relation[], remoteRelations: Rela
   const deletedRelationIds = new Set(locallyDeleted.map(r => r.id));
   const deletedRelationKeys = new Set(locallyDeleted.map(relationKey));
 
+  const localViewFromEntityIds = new Set(
+    localRelations
+      .filter(r => r.type.id === SystemIds.VIEW_PROPERTY && !r.isDeleted)
+      .map(r => r.fromEntity.id)
```
b-d055: The dedup-by-column-property rule is great, but it ends up enforced in five places: `relationKey()` in `core/sync/store.ts`, this near-identical `relationKey()` plus the `localViewFromEntityIds` filter that drops remote `VIEW_PROPERTY` relations, the new `dedupeRelationsByColumnProperty` / `relationsMatchingColumnProperty` helpers in `shown-column-relations.ts`, the read-time `selectViewRelation` + `dedupeRelationsByColumnProperty` calls in `use-view.ts`, and a render-time `useEffect` in `editable-entity-page.tsx` (RelationsGroup, ~L469-508) that issues `storage.relations.delete` ops from a presentation component.

Could we collapse the merge-time half into one site, the `hydrateReactiveState` merge in `store.ts` (~L339)? After building `mergedIncoming`, group `[...unchangedRelations, ...mergedIncoming]` by `relationKey` and keep the newest entry, preferring the one with `isLocal`. That single change subsumes both this `localViewFromEntityIds` filter and the RelationsGroup `useEffect`, and removes the duplicate `relationKey` here.

The write-time helpers should stay. `relationsMatchingColumnProperty` is "find rows to tombstone before insert," not dedup, and `selectViewRelation` is read-time precedence among intentionally-transient overlapping rows (e.g. `setView` writes-new-then-deletes-old).

apps/web/partials/blocks/table/table-block.tsx
```diff
@@ -67,13 +73,52 @@ interface Props {
   onConsumedInitialFiltersOpen?: () => void;
 }
 
-function makePlaceholderRow(entityId: string, properties: { id: string; name: string | null }[]) {
+type PendingEntity = { entityId: string; spaceId: string };
+
+const pendingEntitiesByBlockId = new Map<string, PendingEntity[]>();
```
b-d055: Optimistic-row state is currently tracked in three layers that all carry the same data:

- this module-level `pendingEntitiesByBlockId: Map`
- sessionStorage via `readPendingFromSession` / `writePendingToSession` (~L86-103)
- React state `pendingEntities` (~L205)

Three layers means three cleanup paths. The codebase already has a clean pattern for "ephemeral state mirrored to storage" in `apps/web/core/state/chat-store.ts:54-55` using `atomWithStorage`. Folding the three layers into one atom removes a class of synchronization bugs, and also removes this module-level `Map` (which is an SSR/Next.js anti-pattern when modules cross the RSC boundary).

While the state is being consolidated, `useEntries` (now ~337 lines) is a natural place to split. The seam shows up at `:316-322` and again at `:397-401`, where the same "what space owns this entity?" question is answered with two near-identical four-level `??` chains over `entityTargetSpaceRef` / `pendingMeta` / `placeholderCreateSpaceRef` / `placeholderTargetSpaceId`. Two readers, two precedence orders, one underlying fact.

A `useOptimisticRows(blockEntityId, entries)` hook owning the atom plus `entityTargetSpaceRef` and `committedPlaceholderIdsRef` can expose:

```ts
{ pendingNotInQuery, getTargetSpace, markPending, markCommitted, isCommitted }
```

Both `??` chains collapse to `getTargetSpace(entityId) ?? actionSpaceId`. Active-placeholder state (`hasPlaceholderRow`, `placeholderFocusKey`, focus refs) stays in `useEntries`.

Also: `placeholderTargetSpaceId` is in the return object at L515 but no caller destructures it. Safe to drop from the public surface.

apps/web/partials/blocks/table/query-setup-types-select-entity-popover.tsx
```diff
@@ -0,0 +1,589 @@
...
+
+export function QuerySetupTypesSelectEntityPopover({
+  trigger,
+  spaceId: _spaceId,
+  selectedTypes,
+  onChangeSelectedTypes,
+  allowedTargetTypes,
+  disabled,
+}: QuerySetupTypesSelectEntityPopoverProps) {
+  void _spaceId;
```
b-d055: `spaceId` is accepted as a required prop, destructured as `_spaceId`, then explicitly `void`'d on this line. Either drop the prop from the type/signature or use it. The `void` keeps a lint rule quiet but leaves a confusing API contract for callers.

apps/web/partials/blocks/table/table-block.tsx
```diff
@@ -261,10 +439,68 @@ function useEntries(
...
+        { entityId: idToCreate, spaceId: targetSpaceId },
+        ...prev.filter(p => p.entityId !== idToCreate),
+      ]);
+
+      placeholderCreateSpaceRef.current = targetSpaceId;
+      setPlaceholderTargetSpaceId(targetSpaceId);
+      setHasPlaceholderRow(false);
+      focusRowEntityIdRef.current = idToCreate;
+      setPlaceholderFocusKey(k => k + 1);
+      window.setTimeout(() => {
```
b-d055: `window.setTimeout(..., 300)` to clear `focusRowEntityIdRef` after focus is assumed delivered. The magic 300ms is a guess at React/DOM timing. A `useEffect` keyed on the new row appearing in `renderedEntries` (or on `placeholderFocusKey`) would be deterministic and self-documenting.

apps/web/partials/blocks/table/table-block.tsx
```diff
@@ -95,12 +140,45 @@ function makePlaceholderRow(entityId: string, properties: { id: string; name: st
   };
 }
 
+/** Committed locally but not yet returned by the query */
+function makeOptimisticRow(
+  entityId: string,
```
b-d055: `makePlaceholderRow` (around L100-118) and this `makeOptimisticRow` are near-duplicates. They differ only by `placeholder: true | false` and whether `targetSpaceId` is optional or required.

Could be one function with an options bag:

```ts
function makeRow(
  entityId: string,
  properties: { id: string; name: string | null }[],
  options: { placeholder: boolean; targetSpaceId?: string },
): Row
```

Saves ~30 lines and makes the relationship between the two row types explicit.

apps/web/core/hooks/use-create-entity-with-filters.ts
```diff
@@ -62,13 +79,17 @@ export function useCreateEntityWithFilters(spaceId: string) {
...
+      nextEntityIdRef.current = newId;
+      setNextEntityId(newId);
+      return entityId;
     },
-    [nextEntityId, spaceId, storage]
+    [defaultSpaceId, storage]
   );
 
   return {
     onClick,
```
b-d055, line 92: This hook returns both `nextEntityId` (React state) and a fresh id from `onClick`. The internal ref is updated synchronously inside `onClick`, but the state value lags by one render. Callers that read `nextEntityId` after calling `onClick` will see the previous id, not the new one.

Could you drop `nextEntityId` from the public return and have callers always use the value returned from `onClick`? Removes the ambiguity.

apps/web/partials/blocks/table/table-block-filter-creation-prompt.tsx
```diff
@@ -1094,7 +1321,28 @@ export const TableBlockFilterPrompt = React.forwardRef<TableBlockFilterPromptHan
...
+                        >
+                          Clear all
+                        </button>
+                      ) : null}
+                      {showPopoverDone ? (
+                        <button
+                          type="button"
+                          className="text-smallButton font-medium text-ctaPrimary transition-colors hover:text-ctaHover"
+                          onClick={onEntitiesDone}
+                        >
```
b-d055: This PR replaces an existing `<TextButton color="ctaPrimary">` "Done" with a hand-rolled `<button className="...">` here. The same hand-rolled pattern is now repeated at:

- `apps/web/partials/blocks/table/data-block-scope-dropdown.tsx` L357-378 (`isSetup` branch)
- `apps/web/partials/blocks/table/query-setup-types-select-entity-popover.tsx` L205-237 (`headerRow`)
- `apps/web/partials/blocks/table/table-block-filter-creation-prompt.tsx` L1376-1422 (`FilterValueDropdownFooter`)

The existing `<TextButton color="grey-04" | "ctaPrimary">` primitive in `apps/web/design-system/text-button.tsx` already encodes both color states the new sites need. Could you swap the hand-rolled buttons for `<TextButton>` at these sites? Stops the styles from drifting between sites ("Clear all" already uses `text-text hover:text-ctaPrimary` in two places and `text-grey-04 hover:text-text` in the other two) and keeps the design system as the single source of truth.

apps/web/core/sync/use-store.tsx
```diff
@@ -55,6 +55,65 @@ const reactive = createAtom(
...
+  for (const r of entity.relations ?? []) {
+    if (r.timestamp && r.timestamp > latest) latest = r.timestamp;
+  }
+  return latest;
+}
+
+/**
+ * Prepend unpublished local entities that match the query filter but are not yet in the server page
+ */
+function mergeUnpublishedLocalEntities(
```
b-d055, line 87: `mergeUnpublishedLocalEntities` unconditionally prepends unpublished local entities to the first page of any `useQueryEntities` result. Today only the data-block (`use-data-block.tsx`) and power-tools (`use-power-tools-data.ts`) call sites paginate over an open-ended `where` and benefit from this. Every other caller passes `{ id: { in: [...] } }` and is unaffected, but the behavior is implicit on the general hook.

Could you add an opt-in flag (`includeUnpublishedLocal?: boolean`, defaulting to `false`) and set it `true` only at the two known call sites? Makes the behavior visible to future contributors who might wire up other paginated queries.

apps/web/core/hooks/use-creatable-space-ids.ts
```diff
@@ -0,0 +1,82 @@
...
+  );
+
+  return map;
+}
+
+/**
+ * Per-space member/editor check for the "+" create-entity dropdown.
+ * Uses the same access path as `useAccessControl` (including personal-space ownership).
+ */
+export function useCreatableSpaceIds(spaceIds: string[], enabled: boolean) {
```
b-d055, line 48: For the GEO source path the visible row list can be large (scope sections + remote search), and each row triggers its own `getSpace` + `getSpaceAccess` round-trip. The 60s `staleTime` softens repeats but the initial fan-out is still N requests per dropdown open.

Worth a `// TODO` comment so the next person knows this is a known scaling cliff, or a follow-up if a batch access primitive lands.

apps/web/core/hooks/use-search.ts
```diff
@@ -111,104 +108,55 @@ export function useSearch({
...
+          signal,
+          additionalSpaceIds,
+        });
+
+        const rows = !filterByTypes?.length
+          ? page.results
+          : page.results.filter(r => resultMatchesFilterTypes(r, filterByTypes));
+
+        return { rows, offset: pageParam, rawCount: page.rawCount, total: page.total };
+      } catch (error) {
```
b-d055, line 158: This PR reverts the `queryFn` from `Effect.tryPromise` + `Either.isLeft` to bare `try { ... } catch { return emptyPage }`. Runtime behavior is equivalent (the previous `default: throw error` branch was unreachable because the catch coerced every error into `AbortError`), but this hook is now a stylistic island in a file tree that uses Effect heavily for IO.

Either restore the Effect shape for consistency, or add a one-line comment explaining why this hook deliberately uses bare try/catch. While you're in the file, consider restoring a `console.error(error)` inside the catch so failures are at least visible in the console.

b-d055 (changes requested):
@webmagic123  please see comments and suggestions

b-d055 (approved):

PR #1799, opened 2026-05-14
feat(navbar): hide create dropdown when signed out

jwalkingjew:
## Summary
- The '+' create button in the navbar was always rendered. Gate `CreateEntityDropdown` on the smart account address so it only appears once the user is signed in.
- Uses the same auth signal (`useSmartAccount().smartAccount?.account.address`) that `NavbarActions` already uses to swap between the connect button and the user menu.

## Test plan
- [ ] Signed out: '+' button is not visible in navbar
- [ ] After sign-in: '+' button appears and dropdown works (New entity / New property / Import data in a space context)
- [ ] After sign-out: '+' button disappears


jwalkingjew:
Replaced by a fresh PR based on master.

PR #1812, opened 2026-05-19
feat: add onboarding popups for Geo Browser

webmagic123:


apps/web/partials/hints/edit-mode-toggle-tip.tsx
```diff
@@ -0,0 +1,219 @@
...
+        height,
+        borderRadius,
+        boxShadow: '0 0 0 9999px color-mix(in srgb, var(--color-text) 20%, transparent)',
+      }}
+      {...backdropMotionProps}
+    />
+  );
+}
+
+export function EditModeToggleTip({ open, dismiss, anchorRef }: EditModeToggleTipProps) {
```
b-d055, line 36: @webmagic123 — **Refactor opportunity: shared `<SpotlightTip>` primitive**

The three new tip files are ~80% duplicated:
- `computeAnchorRect` / `computeCutoutRect`
- viewport-clamped `computeTipPosition` (only differs in above/below + arrow math)
- `useLayoutEffect` with scroll/resize/ResizeObserver
- `createPortal` + `AnimatePresence` + backdrop + arrow + dismiss button shell
- the `TIP_GAP_PX` / `TIP_WIDTH_PX` / `VIEWPORT_PADDING_PX` / `SCROLLBAR_GUTTER_PX` constants

Proposed shape:
```tsx
<SpotlightTip
  anchorRef={ref}
  placement="below"
  width={180}
  open={open}
  onDismiss={dismiss}
>
  Describe your edits before publishing
</SpotlightTip>
```

Each tip file collapses to ~20 lines of "what triggers this tip." Could land as a follow-up.

apps/web/partials/onboarding/review-edits-tip.tsx
```diff
@@ -0,0 +1,261 @@
...
+}
+
+/**
+ * Sum the translation of every ancestor transform so we can subtract it from
+ * the element's bounding rect. The flow bar mounts with a small entry slide
+ * (`y: 4 → 0`); without this, getBoundingClientRect captures the in-flight
+ * transform and the cutout would follow the animation instead of snapping to
+ * the final pill position.
+ */
+function getAncestorTransformOffset(element: HTMLElement): { x: number; y: number } {
```
b-d055: @webmagic123 — `FLOWBAR_ENTRY_DELAY_MS = 250` above already waits for framer-motion's entry slide to finish; after settle, ancestor transforms are `none`, so this math is a no-op in steady state.

It also compensates indiscriminately — any future hover/active transform on an ancestor (e.g. a scale-on-hover added to the flow-bar wrapper) will offset the cutout. Either trust the settle delay and drop this helper, or listen for `transitionend` from the flow-bar wrapper directly.

apps/web/partials/onboarding/edit-mode-toggle-tip.tsx
```diff
@@ -0,0 +1,219 @@
...
+const TIP_WIDTH_PX = 168;
+const TIP_NUDGE_RIGHT_PX = 12;
+const VIEWPORT_PADDING_PX = 16;
+
+const SCROLLBAR_GUTTER_PX = 20;
+
+const TIP_Z_BACKDROP = 1000;
+const TIP_Z = 1001;
+
+const SPOTLIGHT_PADDING_PX = 0;
```
b-d055: @webmagic123 — `SPOTLIGHT_PADDING_PX = 0` is added/subtracted in `computeAnchorRect` / `computeCutoutRect`, but the constant is `0`, so the math is a no-op. Either remove the constant + math, or set a non-zero default if the intent was to give the cutout breathing room. (Same applies in `proposal-name-tip.tsx`.)

apps/web/partials/onboarding/proposal-name-tip.tsx
```diff
@@ -0,0 +1,251 @@
...
+    if (!open) {
+      setPosition(null);
+      setCutoutRect(null);
+      return;
+    }
+
+    let cancelled = false;
+    let rafId = 0;
+    let retryCount = 0;
+    const maxRetries = 90;
```
b-d055: @webmagic123 — this effect runs three measurement strategies in parallel: (1) a 90-frame rAF retry loop (~1.5s), (2) a 600ms `setTimeout`, and (3) the standard `useLayoutEffect` measurement. If the 600ms timeout is sufficient, the rAF loop is redundant. If there's a race the loop is guarding against, a comment naming it ("anchor mounts after SlideUp's first paint because X") would make it clear which strategy is load-bearing.

b-d055 (commented):
@webmagic123 — did a pass over the Copilot comments. Quick triage:

**Worth acting on:**
- `review-changes.tsx:411` — gate `useProposalNameTip` on `statusBarState.reviewState === 'idle'`. The tip can otherwise render over the publish-progress pill or error modal. One-line fix since `statusBarState` is already in scope on that component.
- `proposal-name-tip.tsx:149` — the early-return on a hidden anchor doesn't clear `position`/`cutoutRect`, so the tip freezes at its last spot instead of exiting. The other two tips don't check visibility at all.

**Real issue, wrong fix — z-index (`proposal-name-tip.tsx:21` + `review-edits-tip.tsx:21`):**
The tips at z-10050/10051 do outrank the status bar (z-10001), but Copilot's suggested fix won't work — `SlideUp` is `z-[10000]` and status-bar is `z-10001`, so there's no integer room between them. Once `useProposalNameTip` is gated on review state (above), the tip never co-renders with the status bar and z-order stops mattering for the bug. A shared `z-layers.ts` module would let us renumber the stack coherently if we want belt-and-suspenders later.

**Partially right — accessibility (×3 comments):**
- `role="dialog"` is the wrong role for these.
- Adding a focus trap / `aria-modal` conflicts with the `pointer-events-none` backdrop, which exists so the user can keep interacting with the underlying control. Switching the role to `tooltip` or `status` and adding Escape-to-dismiss covers the intent without modal semantics.

Leaving a few separate comments for things outside Copilot's review.

apps/web/atoms/product-onboarding.ts
```diff
@@ -0,0 +1,13 @@
+import { atomWithStorage } from 'jotai/utils';
+
+/** Persisted after the user dismisses the edit/browse mode toggle tip on their personal space. */
+export const editModeToggleTipDismissedAtom = atomWithStorage<boolean>(
```
b-d055: Three new versioned boolean keys (`geoEditModeToggleTipDismissedV1`, etc.) will land in every user's localStorage permanently. The codebase already has the pattern for this in `atoms/index.ts:13`:

```ts
export const dismissedNoticesAtom = atomWithStorage<Array<string>>('dismissedNotices', []);
```

Used today by `space-notices.tsx` and `personal-home-dashboard.tsx`. Suggest one `Array<string>` atom keyed by hint id instead:
- One key in localStorage instead of three.
- Adding a hint = one new id string. `useResetOnboarding` resets all hints with a single setter.

Worth doing pre-merge specifically because storage keys are sticky; changing them later requires a migration shim or leaves orphan keys.

apps/web/partials/hints/edit-mode-toggle-tip.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { useHydrated } from '~/core/hooks/use-hydrated';
+import { useOnboarding } from '~/core/hooks/use-onboarding';
+import { usePersonalSpaceId } from '~/core/hooks/use-personal-space-id';
+import { useSpaceId } from '~/core/hooks/use-space-id';
+
+import { SpotlightTip } from './spotlight-tip';
+
+const TIP_NUDGE_RIGHT_PX = 12;
+
+export function useEditModeToggleTip() {
```
b-d055, line 16: This hook, `useReviewEditsTip` (`review-edits-tip.tsx:18`), and `useProposalNameTip` (`proposal-name-tip.tsx:15`) share the same shape: `hydrated && gate && !dismissed && !isOnboardingVisible` then `useState(false)` then `useEffect` with optional settle timeout then `{ open: shouldOffer && open, dismiss }`. The only differences are the gate condition, the settle delay (0 / 250 / 600), and which atom is read.

Worth collapsing into one helper, e.g. `useDismissibleHint(id, { gate, settleMs = 0 })`. Pairs naturally with the storage change in comment 1 (one atom keyed by id).

apps/web/partials/onboarding/proposal-name-tip.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+import { useAtom } from 'jotai';
+
+import { proposalNameTipDismissedAtom } from '~/atoms/product-onboarding';
+import { useHydrated } from '~/core/hooks/use-hydrated';
+import { useOnboarding } from '~/core/hooks/use-onboarding';
+
+import { SpotlightTip } from './spotlight-tip';
+
+const SLIDE_UP_SETTLE_MS = 600;
```
b-d055: `SLIDE_UP_SETTLE_MS = 600` here and `FLOWBAR_ENTRY_DELAY_MS = 250` in `review-edits-tip.tsx:16` duplicate animation durations that live in `slide-up.tsx` (`{ type: 'spring', duration: 0.5 }`) and `flow-bar.tsx` (`duration: 0.15, delay: 0.15`). If either animation is retuned, the tip timing drifts silently.

Two options:
- Export the durations as constants from `slide-up.tsx` / `flow-bar.tsx` and import them here.
- Use framer-motion's `onAnimationComplete` on the panel/bar to drive a "settled" callback the tip subscribes to. Removes the timer entirely.

apps/web/partials/hints/spotlight-tip.tsx
```diff
@@ -0,0 +1,273 @@
+'use client';
```
b-d055, line 1: `partials/onboarding/` currently means the wallet/profile wizard (`dialog.tsx` + `useOnboarding`). The tip components added in this PR are product hints that fire after that flow. They gate against `!isOnboardingVisible`, but they're a separate concept.

Suggest moving the new tip files to a generic location like `partials/hints/` or `partials/coachmarks/`, and renaming `atoms/product-onboarding.ts` to match. As more hints get added, tying them to the "onboarding" name will be misleading.

b-d055 (commented):
A few maintainability and perf comments. None blocking, but the storage-key one is cheaper to address pre-merge than after.

b-d055 (approved):

PR #1824, opened 2026-05-20
fix: navigate to the posts tab and show a new post entity as the side…

webmagic123:
… panel when create a post


apps/web/core/state/editor/editor-provider.tsx
```diff
@@ -118,21 +121,26 @@ function EditorBlocksProvider({ children }: { children: React.ReactNode }) {
     initialBlocks,
     initialTabs,
     initialCollectionItems: allCollectionItems,
+    ignoreRouteTabId,
   } = useEditorInstance();
 
-  const tabId = useTabIdFromSearchParams();
-  const activeEntityId = tabId ?? entityId;
-  const isTab = React.useMemo(() => tabId && !!initialTabs && Object.hasOwn(initialTabs, tabId), [initialTabs, tabId]);
+  const routeTabId = ignoreRouteTabId ? null : useTabIdFromSearchParams();
```
b-d055: This calls `useTabIdFromSearchParams()` only when `ignoreRouteTabId` is false, which violates `react-hooks/rules-of-hooks` and will be flagged by lint. It happens to work today because the prop is stable per mount, but flipping the prop on a mounted instance would crash.

Minimal fix — hoist the call:

```ts
const searchParamTabId = useTabIdFromSearchParams();
const routeTabId = ignoreRouteTabId ? null : searchParamTabId;
```

The cleaner long-term shape is to drop `ignoreRouteTabId` entirely and accept an explicit `activeTabId` prop, so panel and route both pass their tab source in deliberately rather than the provider reading `useSearchParams()` internally. Happy to take that as a follow-up — the hoist above is enough for now.

apps/web/partials/entity-page/personal-profile-create-post-side-panel-sync.tsx
```diff
@@ -0,0 +1,68 @@
...
+      void router.push(postsTabUrl, { scroll: false });
+      return;
+    }
+
+    if (!tabMatches) {
+      void router.replace(postsTabUrl, { scroll: false });
+      return;
+    }
+
+    setEditable(true);
```
b-d055: This flips the global `editable` atom to open the panel in edit mode. Nothing resets it on panel close, so the underlying profile page is silently left in edit mode after the user dismisses the panel.

`openSidePanel(postEntityId, spaceId, true)` on the next line already passes `openedWithMainViewEditing=true` into the panel's own edit context (`EntitySidePanelEditModeProvider`), so I think this `setEditable(true)` call can just be dropped — the panel will still open in edit mode without leaking into the route.

Worth reproducing locally: trigger the create-post flow, dismiss the panel, confirm the profile page is no longer in edit mode.

apps/web/partials/entity-page/entity-side-panel.tsx
```diff
@@ -0,0 +1,512 @@
...
+              </React.Suspense>
+              <Spacer height={40} />
+              <Editor spaceId={entitySpaceId} shouldHandleOwnSpacing />
+              <ToggleEntityPage id={entityId} spaceId={entitySpaceId} />
+              <CommentSection entityId={entityId} spaceId={entitySpaceId} />
+            </div>
+          </EntityPageContentContainer>
+        </div>
+      </EditorProvider>
+    </EntityStoreProvider>
```
b-d055, line 248: This block and `apps/web/app/space/(entity)/[id]/[entityId]/default-entity-page.tsx` lines 65–107 render the same cover / container / heading / inline description / metadata header / tabs / editor / toggle / comments sequence inside `EditorProvider`. The panel will silently drift from the canonical entity-page layout the moment one side changes.

Suggest extracting a shared `EntityPageBody` (or `EntityPageContent`) that takes `entityId`, `spaceId`, `tabRelations`, `tabEntities`, and a `variant` for the small chrome differences. Each caller keeps its own data loading and providers, so this doesn't pull in server fetching or require an `EditorProvider` API change — it's purely lifting the duplicated JSX.

A full `EntityPageSurface` that also unifies data loading and provider setup is probably too much here — route pages are server-prefetched and the panel is client-loaded, so unifying those is a separate refactor. The data-loading hook and explicit `activeTabId` API are good follow-ups once the body extraction lands.

b-d055 (changes requested):

b-d055:
@webmagic123  Two things I'd want addressed before merge, one structural change that's small enough to fit here, and a few follow-ups to file separately.

1. Conditional hook call in `EditorProvider` (see inline): rules-of-hooks violation that will be flagged by lint.
2. The create-post flow flips the global `editable` atom and never resets it (see inline): leaks edit mode into the underlying profile page after the panel closes.

Beyond those: extract a shared `EntityPageBody` so the side panel can't drift from the canonical entity-page layout. A full `EntityPageSurface` (unifying data loading + `EditorProvider` API change) is the right end state but probably too much for this PR, route pages are server-prefetched while the panel is client-loaded, so unifying those is a separate refactor.

Follow-ups:
- Promote `ignoreRouteTabId` to an explicit `activeTabId` prop on `EditorProvider`. Cleanest after the body extraction lands.
- Lift the side-panel data loading into a `useEntityPageSurfaceData` hook so `EntitySidePanelBody` isn't doing scope resolution + four entity queries + render orchestration in one component. `useSidePanelEntityScope` in particular is dense and untested.
- Encode the create-post flow as a typed state machine. The path-change auto-close in `EntitySidePanel` has to special-case `pendingCreatePostSidePanel` because the navigate → set-edit → open sequence races with it — that bypass is the symptom.
- No tests cover the new control flow. Even one assertion that "open via create-post, close panel, profile not in edit mode" would catch the leak above.

apps/web/core/state/editor/editor-provider.tsx
webmagic123: @b-d055 
Updated

apps/web/partials/entity-page/personal-profile-create-post-side-panel-sync.tsx
webmagic123: @b-d055 
Updated

apps/web/partials/entity-page/entity-side-panel.tsx
webmagic123: @b-d055 
Updated

b-d055 (approved):

PR #1829, opened 2026-05-23
feat: update onboarding flow

Ivanwell:
- Feature:
  A reworked new-user onboarding + personal-space creation flow. 
  A new account can pick interest spaces and a role, and get an optimistic, instantly-usable account — the UI reflects registration and membership requests immediately, while the on-chain registration and indexer catch up in the background.

- Done:
  1. Onboarding UI redesign
  2. Space creation + role relations
  3. Membership requests routed through one path
  4. Optimistic "Membership pending" bridge
  5. Logout cleanup
  
- Solution:
Gap: interest picks still waited on the full create chain — registration plus the 30s–3min indexer wait - before showing anything.

1.  onRegistered callback on createPersonalSpaceOnChain / useCreatePersonalSpace, fired the moment the space is on-chain (registered + content landed) but before the indexer wait - so the account resolves and membership requests fire in seconds. Guarded so a callback throw can't fail an already-succeeded creation.
2.  Wallet-address scoping: during onboarding the personal space id doesn't exist yet, so optimistic pending rows are keyed by the wallet address; readers (activeRequestedSpacesForOwner / reconcileRequestedSpaces) match either id, so the row survives once the real id resolves.
3. Flip -> enrich -> roll back: picks flip to pending immediately, display data fills in from a background fetch, and everything rolls back if creation fails.
4. Sidebar dedupe by id, since a pick can briefly hold both an address-scoped and a personal-space-scoped entry.
5. Plus this session's cleanups: extracted the duplicated fire/roll-back logic into settleMembership(), removed a guaranteed-no-op safety net


apps/web/partials/onboarding/dialog.tsx
```diff
@@ -129,6 +158,20 @@ export const OnboardingDialog = () => {
...
+  // featured-space traversal the Browse sidebar uses, so there's no local
+  // fallback list to fall back to — an empty step just shows no topics.
+  useEffect(() => {
+    if (step !== 'interested-in') return;
+
+    fetchBrowseSidebarData(undefined)
+      .then(data => setFeaturedSpaces(data.featured))
+      .catch(error => {
+        console.error('[onboarding] failed to load featured spaces', error);
+        setFeaturedSpaces([]);
```
b-d055, line 172: **Onboarding dead-ends here when the featured fetch fails.**

Reproduced on the preview build: block the GraphQL endpoint while on the "Welcome to Geo" step, then hit Continue. `findFuzzy` fails, gets caught at L452, and calls `onProfileContinue([])` — so you skip "Is this you?" and land straight on `interested-in`. The featured fetch then fails too, and `runQuery` collapses any error into `null` (`fetch-featured-spaces.ts:139-146`), so `fetchFeaturedSpaces` returns `[]` — the same value it returns when there genuinely are no featured spaces. This `.catch` produces `[]` for the same reason.

End state: an empty card, `Create profile` disabled (L772), and Esc / outside-click both prevented (L247-249). The only exit is Log out. A reload does recover it — `onboardingStep` is persisted, so the effect refires — but nothing tells the user that.

Not asking to drop the required pick. It just needs the failed and empty cases separated so the step can show something:

- fetch failed → error state + Retry
- fetched and genuinely empty → allow continue

That needs `fetchBrowseSidebarData` to surface which one happened; right now it can't distinguish them.

apps/web/partials/onboarding/pending-personal-space-runner.tsx
```diff
@@ -1,21 +1,63 @@
...
 import { useSyncEngine } from '~/core/sync/use-sync-engine';
 import { readRegisteredSpaceId } from '~/core/utils/contracts/create-personal-space-on-chain';
 import { devLog } from '~/core/utils/dev-log';
 import { describeError } from '~/core/utils/error-diagnostics';
 
-import { avatarAtom, nameAtom, spaceIdAtom } from './dialog';
+import { avatarAtom, nameAtom, selectedRoleIdsAtom, selectedTopicIdsAtom, spaceIdAtom } from './dialog';
+
+/** Links the new personal space entity to each Geo role the user picked during onboarding. */
+function createGeoRoleRelation(spaceId: string, fromEntityId: string, roleEntityId: string) {
```
b-d055: **Context for the Copilot comments on this path — the roles step was dropped on request, so this is dead code rather than a bug.**

Nothing writes `selectedRoleIdsAtom` any more, so `roleIds` is always `[]` and neither this function nor the settlement block at L285-297 can run. That means three of Copilot's comments here don't need fixing as filed:

- the `entityId` collision on this function
- "picks are cleared even when `getSpace` returns null" (L300)
- "restore role selection or update the stated scope" (`dialog.tsx:64`)

Rather than fix them, I'd delete the path:

- `createGeoRoleRelation` and the `GEO_ROLES_PROPERTY` import here
- the `roleIds` block at L285-297 and the `selectedRoleIds` plumbing (L90-95, L299)
- `selectedRoleIdsAtom` in `dialog.tsx:66` and its reads in `use-geo-logout.ts`
- `GEO_ROLE_TYPE`, `DOMAIN_EXPERT_ROLE`, `DEVELOPER_ROLE`, `CURIOUS_EXPLORER_ROLE` in `constants.ts:131-134` — unreferenced (`GEO_ROLES_PROPERTY` is used only by the dead code above)

Worth recording in case it's ever revived: Copilot is right that `entityId: spaceId` is wrong. `Relation.entityId` is the relation entity's own id (`core/types.ts:242`), so every role relation would alias onto the space entity. Compare `personal-post-flow.ts:89-91`, which generates a fresh id for it.

One thing that does need a change: the PR description still lists "pick interest spaces **and a role**" and "Space creation + role relations" under Done. That's what sent both Copilot and me down this path — worth updating.

*Separate nit, different file:* if skipping the interests step for a claimed profile is intentional, `DOT_CONFIGS['existing-entity-match']` still renders two dots on that path and only delivers one step.

b-d055 (commented):
@webmagic123  can you make these changes?

apps/web/partials/onboarding/pending-personal-space-runner.tsx
webmagic123: @b-d055 
Updated.

apps/web/partials/onboarding/dialog.tsx
webmagic123: @b-d055 
Updated.

b-d055 (approved):

PR #1849, opened 2026-06-05
feat: add publish rankings feature

webmagic123:


apps/web/core/hooks/use-request-to-be-member.ts
```diff
@@ -12,16 +12,21 @@ import { usePersonalSpaceId } from '~/core/hooks/use-personal-space-id';
...
 
 interface UseRequestToBeMemberArgs {
-  /** The space ID (bytes16 hex without 0x, e.g., UUID format) of the space to join */
   spaceId: string | null;
 }
 
+/**
+ * Guards against duplicate membership requests across hook instances
+ */
+const inFlightOrSubmittedRequests = new Set<string>();
```
b-d055: **Suggestion: revert this file and scope the dedupe to the ranking access hook — this change regresses behavior that was deliberately fixed in #1472.**

This session-scoped Set adds the key on entry but only deletes it on error, so after a successful submission (or the `alreadyRequested` early-return) the key stays for the lifetime of the tab. Since this hook also powers the explore/space-page join buttons, that changes their behavior app-wide: a user whose membership proposal is **rejected** and who clicks "Request to join" again gets a silent no-op — the early return resolves the mutation as a success, so the button shows "Requested" even though nothing was sent. Re-request-after-rejection was a deliberate fix in #1472 (`b3bf784a5`), and this re-breaks it.

The race this Set guards is real, but it only exists in the **ranking auto-fire path**: `ensureMembership` in `use-ranking-compose-access.ts:78-80` fires `requestToBeMember()` from an effect, and it's mounted by both the compose screen and the embedded block — two instances with separate `membershipRequestedRef`s can each submit before either tx lands. That can be solved entirely inside the ranking-owned file, leaving this shared hook byte-identical to master:

```ts
// use-ranking-compose-access.ts (ranking-only file)
const autoRequestedMemberships = new Set<string>();

// inside ensureMembership, replacing the per-instance ref check:
const requestKey = `${spaceId}:${personalSpaceId}`;
if (!autoRequestedMemberships.has(requestKey) && membershipRequestStatus !== 'pending') {
  autoRequestedMemberships.add(requestKey);
  const alreadyPending = await hasActiveMemberProposal(spaceId, personalSpaceId).catch(() => false);
  if (!alreadyPending) {
    requestToBeMember();
  }
}
```

This keeps everything the Set was added for, scoped to where the risk lives:
- **Cross-mount race**: the module-level Set in the ranking hook dedupes across instances synchronously, same as here.
- **Already-pending dedupe**: `hasActiveMemberProposal` (existing master helper, untouched by this PR) is checked before auto-firing.
- **"Once per session" becomes harmless**: it only suppresses the *automatic* background request — users can always still join manually via the buttons, which would go through the unmodified master hook.

The only thing the revert gives up is the `hasActiveMemberProposal` pre-check on manual button clicks — which master never had either, so that's the status quo, not a regression. If that check is wanted for the buttons too, it deserves its own PR where the retry semantics from #1472 can be preserved explicitly (e.g. release the key in a `finally` instead of the `catch`).

apps/web/core/hooks/use-ranking-compose-access.ts
```diff
@@ -0,0 +1,96 @@
...
+  });
+
+  const isLoading = isLoadingPersonalSpace || isLoadingSpace || isLoadingAccess;
+
+  const status: RankingComposeAccessStatus = (() => {
+    if (isLoadingSmartAccount) return 'loading';
+    if (!smartAccount) return 'needs-login';
+    if (isLoading || !isFetched) return 'loading';
+    if (!isRegistered || !personalSpaceId) return 'needs-onboarding';
+    if (space?.type === 'DAO' && !canEdit) return 'needs-membership';
```
b-d055: **Access check silently passes for a nonexistent space.**

When `useSpace(spaceId)` finishes loading and returns `space = null` (any well-formed but nonexistent space ID in the URL — `getSpace` explicitly returns `null` for unknown spaces), the status block falls through to `'ready'`:

```ts
if (space?.type === 'DAO' && !canEdit) return 'needs-membership'; // space=null → false → falls through
return 'ready';
```

The route only validates that the params are non-empty (`hasValidParams`), not that the space exists, so this is reachable by typing a URL. With status `'ready'`, the compose UI renders fully and publish proceeds (the rank ops go to the user's personal space, so the write even succeeds) — instead of showing a not-found/error state. And because the status never becomes `'needs-membership'` for a null space, `ensureAccess()` is never invoked on this path, so nothing downstream catches it either.

Suggest an explicit branch once loading completes, e.g. a `'not-found'` status that the compose screen and embedded block render as an error (or redirect away):

```ts
if (!space) return 'not-found';
if (space.type === 'DAO' && !canEdit) return 'needs-membership';
```

(Mapping `!space` to `'needs-membership'` instead would be wrong — it would show a misleading "request membership" state for a space that doesn't exist.)

apps/web/core/blocks/ranking/use-ranking-submissions.ts
```diff
@@ -0,0 +1,244 @@
...
+import type { RankingSubmissionSlot } from './ranking-submission-types';
+import { linearVoteWeight } from './ranking-vote-weights';
+import { useMyRanking } from './use-my-ranking';
+
+/** One ballot per author — avoids double-counting duplicate ballots per author. */
+export function dedupeSubmissionsByAuthor(submissions: RankingSubmissionRecord[]): RankingSubmissionRecord[] {
+  const byAuthor = new Map<string, RankingSubmissionRecord>();
+  for (const submission of submissions) {
+    const existing = byAuthor.get(submission.authorSpaceId);
+    if (!existing || submission.createdAt >= existing.createdAt) {
```
b-d055: **`dedupeSubmissionsByAuthor` can keep the stale ballot when timestamp formats differ.**

(Note: this is latent rather than live today — the function is currently only referenced by its own test, since `useRankingSubmissions` still returns an empty `submissions` array pending the global-rankings backend. Flagging now while it's cheap, because it gets wired up the moment that lands.)

`createdAt` is built as `String(myRankEntity.updatedAt ?? '')`, and `updatedAt` is typed `string | number` — `core/io/schema.ts` documents exactly why: *"The indexer has historically returned these as stringified unix seconds, but pg_graphql's Datetime scalar can also serialize as a full ISO string."*

That makes the lexicographic compare here format-dependent:

```ts
if (!existing || submission.createdAt >= existing.createdAt) {
```

`'1704067200'` (unix seconds) vs `'2027-06-10T14:22:30Z'` (ISO) compares by first character, so whenever one author's submissions arrive in mixed formats, the *older* ballot can win and the user's latest ranking is silently dropped from the accumulation. Because the failure is silent and data-shaped, it would be very hard to diagnose once this is live.

Suggest normalizing before comparing:

```ts
const toEpochMillis = (v: string) => (/^\d+$/.test(v) ? Number(v) * 1000 : Date.parse(v) || 0);

if (!existing || toEpochMillis(submission.createdAt) >= toEpochMillis(existing.createdAt)) {
```

(`dedupe-submissions.test.ts` currently only exercises same-format ISO inputs — worth adding a mixed-format case so this stays covered.)

apps/web/core/hooks/use-onboarding.tsx
```diff
@@ -10,6 +10,7 @@ import { useSearchParams } from 'next/navigation';
 import { usePersonalSpaceId } from './use-personal-space-id';
 
 const isOnboardingVisibleAtom = atom(false);
+const onboardingDismissedAtom = atom(false);
```
b-d055: **Dismissal looks good — but let's not let a dismissed user browse the whole session without a personal space.**

Making onboarding dismissible is a nice unblock for the ranking flow (and the `create-space`/`completed` steps being non-dismissible is exactly right). The gap is what happens *after* dismissal: since `shouldOnboard` now stays false for the rest of the tab session, a signed-in-but-unregistered user can browse the entire app in a state that used to be impossible. Everywhere that needs a `personalSpaceId` (joining spaces, proposals, edits) they'll hit error toasts instead of the onboarding dialog, and only the ranking compose flow knows how to recover (it's the only caller of `showOnboarding`).

Suggestion: treat dismissal as "not right now" rather than "not this session" — re-arm the prompt on the next navigation:

```ts
// use-onboarding.tsx — store where it was dismissed, not just that it was
const onboardingDismissedAtPathAtom = atom<string | null>(null);

export function useOnboarding() {
  const pathname = usePathname();
  const [dismissedAtPath, setDismissedAtPath] = useAtom(onboardingDismissedAtPathAtom);
  const onboardingDismissed = dismissedAtPath === pathname;

  // hideOnboarding: setDismissedAtPath(pathname)
  // showOnboarding / onDisconnect: setDismissedAtPath(null)
  // shouldOnboard: unchanged — already checks !onboardingDismissed
}
```

That keeps the behavior this branch needs — the user can close the dialog and finish what they're looking at, and the back-arrow dismissal from the ranking profile step still works — but the moment they navigate anywhere else, the prompt returns. No one ends up wandering the app unregistered.

(Alternative if per-navigation feels too aggressive: keep the session flag but have the *generic* "needs a personal space" error paths call `showOnboarding()` instead of toasting, the way the compose screen does. Either works; the current "dismiss once, suppressed all session" is the only variant that leaves users stranded.)

b-d055 (commented):
@webmagic123  see comments

apps/web/core/hooks/use-onboarding.tsx
webmagic123: @b-d055 
Updated.

apps/web/core/blocks/ranking/use-ranking-submissions.ts
webmagic123: @b-d055 
Updated.

apps/web/core/hooks/use-ranking-compose-access.ts
webmagic123: @b-d055 
Updated.

apps/web/core/hooks/use-request-to-be-member.ts
webmagic123: @b-d055 
updated.

PR #1867, opened 2026-06-11
fix: bury property groups in property "+" menu

b-d055:
Property groups are an advanced feature, so showing a prominent "Property groups" section while setting up a new type entity added noise to the common case. Remove the standalone header and only render the groups editor once a group exists. Group creation now lives in the property "+" menu (New property / New property group) via a shared useCreatePropertyGroup hook, with a focus atom coordinating auto-focus across page sections.

### DIFF
--- a/apps/web/atoms/index.ts
+++ b/apps/web/atoms/index.ts
@@ -5,6 +5,11 @@ export const showingIdsAtom = atomWithStorage<boolean>('showingIds', false);
 
 export const editingPropertiesAtom = atom<boolean>(false);
 
+// Holds the entity id of a newly-created property group whose name input should
+// auto-focus. Set when a group is created (from the "+" menu) and consumed by the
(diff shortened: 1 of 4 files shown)


ohohoreilly:
Closing as part of a sweep of the open-PR queue — **not a judgement on the work**, and please reopen if you still want it.

Opened 2026-06-11, and it now conflicts with master. At this distance a rebase is usually more work than redoing the change against current code, and the property menu has almost certainly moved underneath it.

@b-d055 — if this is still wanted, reopen it and I will help get it current. If the idea still stands but the branch does not, it is probably worth a fresh PR or a ticket rather than reviving this one.

Context: there are 60 open PRs, 10 older than two months. The intent is to make the queue mean something, so that genuinely ready work is visible instead of buried — #2449 sat ready for three days this week partly for that reason. Nothing here is being discarded: the branch and its history stay, and reopening costs a click.

PR #1870, opened 2026-06-12
Rankings social preview image

yanivtal:


apps/web/app/api/ranking-og/route-utils.ts
```diff
@@ -0,0 +1,47 @@
...
+import { type RankingOgVariant, isRankingOgVariant } from '~/core/blocks/ranking/ranking-og-storage';
+
+export function jsonResponse(status: number, body: unknown) {
+  return new Response(JSON.stringify(body), {
+    status,
+    headers: { 'Content-Type': 'application/json' },
+  });
+}
+
+export function isSameOrigin(req: Request): boolean {
```
b-d055, line 12: This one was flagged by calude and it's worth considering:

**these routes are effectively unauthenticated and abusable.**

`generate` only checks `isSameOrigin()` here, which compares the request's own `Origin` header to its `Host` — any scripted client sets both, and a missing `Origin` returns `true` outside production (line 15). `preview/route.tsx` has no auth at all. And the R2 object key includes the caller-supplied `ogVersion` (only shape-checked by `isSafeOgVersion`), so one real rank can spawn unlimited stored objects — `rankingOgObjectExists` only dedupes identical keys.

**Why a shared secret isn't the fix here:** the browser calls `generate` directly in the publish flow (`ranking-compose-screen.tsx:71`). A shared secret can't authenticate a browser — anything the client knows, an attacker reads from the bundle/Network tab. The existing `hasRankingOgAdminSecret` (`x-internal-api-secret`) is correct for the **server-to-server** caller (the backfill script at `scripts/generate-ranking-og.ts:55`), but not for the browser path. So fix per caller:

**Recommended — authenticate the user, not the origin.** We already have this pattern: `/api/chat/authorize-write` reads the `WALLET_ADDRESS` cookie via `cookies()` + `parseWalletCookie` (`core/cookie.ts`). Since `generate` is server-side it can do the same: verify the wallet cookie *and* that the wallet owns the `authorSpaceId` being generated for, keeping the admin secret as the OR-branch for the script. Then add a per-wallet rate limit. This ties every call to a logged-in author acting on their own rank, which also defuses the `ogVersion` storage-blowup (only the author can write keys under their own rank).

```ts
// generate/route.tsx
import { cookies } from 'next/headers';
import { WALLET_ADDRESS } from '~/core/cookie';

const wallet = parseWalletCookie((await cookies()).get(WALLET_ADDRESS)?.value);
const isAuthor = wallet && (await walletOwnsSpace(wallet, body.authorSpaceId));
if (!isAuthor && !hasRankingOgAdminSecret(req)) {
  return jsonResponse(403, { ok: false, error: 'forbidden' });
}
```

**Deeper alternative:** drop the browser's eager call and generate lazily on the read path — when `file` gets a cache miss, generate-and-store there behind the internal secret (in-process). The publish flow then POSTs nothing and there's no public mutation endpoint to abuse. More refactoring, but removes the whole class of problem.

Either way, `preview` still needs auth + a rate limit (it renders a real image per unauthenticated GET).

**Reproduce:**
```bash
# preview: zero auth, renders a real image per hit
curl -s "https://<host>/api/ranking-og/preview?rankEntityId=<rank>&authorSpaceId=<sp>&blockEntityId=<bl>&blockEntitySpaceId=<sp>" -o out.png

# generate: spoof origin, balloon storage by varying ogVersion
for i in $(seq 1 500); do
  curl -s -X POST "https://<host>/api/ranking-og/generate" \
    -H "Origin: https://<host>" -H "Content-Type: application/json" \
    -d "{\"rankEntityId\":\"<rank>\",\"authorSpaceId\":\"<sp>\",\"blockEntityId\":\"<bl>\",\"blockEntitySpaceId\":\"<sp>\",\"ogVersion\":\"v$i\",\"variants\":[\"landscape\",\"story\"]}"
done
# -> 1000 new R2 objects + renders from an unauthenticated box
```

apps/web/core/blocks/ranking/ranking-og-data.ts
```diff
@@ -0,0 +1,129 @@
...
+  if (!profile.avatarUrl || profile.avatarUrl === PLACEHOLDER_SPACE_IMAGE) return null;
+  return profile.avatarUrl;
+}
+
+function entityDisplay(entity: Entity | undefined, entityId: string): RankingOgEntryData {
+  return {
+    entityId,
+    name: entity?.name?.trim() || 'Untitled',
+    description: entity?.description?.trim() || Entities.description(entity?.values ?? [])?.trim() || null,
+    image: Entities.cover(entity?.relations) ?? Entities.avatar(entity?.relations) ?? null,
```
b-d055, line 85: This was another one flagged by claude, it's a security risk and likely worth fixing:

This line takes `Entities.cover()/avatar()` (raw `.toEntity.value`) and line 124 takes `profileAvatar()` (only filters the placeholder); both land in `` in `ranking-og-image.tsx`, which Satori fetches **server-side** at render time. No host/scheme allowlist. Combined with the unauthenticated `preview` route, anyone can make our server fetch an arbitrary URL.

Add a host/scheme allowlist (CDN + IPFS gateway) before the value reaches the renderer.

**Reproduce:**
1. Set your profile avatar (or a ranked entity's cover) to `http://169.254.169.254/latest/meta-data/` or a host you control (e.g. RequestBin).
2. `curl "https://<host>/api/ranking-og/preview?rankEntityId=<rank>&authorSpaceId=<sp>&blockEntityId=<bl>&blockEntitySpaceId=<sp>"`
3. Observe the server-originated fetch (and its source IP) hit your endpoint.

apps/web/partials/blocks/table/ranking-compose-screen.tsx
```diff
@@ -0,0 +1,438 @@
...
+    const ogVersion = buildRankingOgVersion({
+      rankEntityId: published.rankEntityId,
+      orderedEntityIds: published.orderedEntityIds,
+      rankingName: displayName,
+      rankingStartDate,
+      rankingEndDate,
+      authorName: published.authorName,
+      authorAvatarUrl: published.authorAvatarUrl,
+    });
+    const generated = await generateRankingOgImages({
```
b-d055, line 299: It looks like this `await generateRankingOgImages(...)` gates the redirect to the share view (line 310), but the fetch has no timeout/abort. The isSaving clears when the chain publish finishes  before this call, so on a slow R2 the user sits on a dead screen. I'm not sure how likely this is, but on failure the `catch` only `console.error`s and silently drops the share params. I would make this fire-and-forget or wrap it in `AbortSignal.timeout(...)` and navigate regardless.

patches/@geoprotocol%2Fgeo-sdk@0.19.3.patch
```diff
@@ -0,0 +1,27 @@
+diff --git a/dist/src/ranks/vote-ops.js b/dist/src/ranks/vote-ops.js
```
b-d055, line 1: This should go away. This rewrites the SDK's compiled `dist/src/ranks/vote-ops.js`. I'm assuming you had this in place to test something temporarily?  IF there really is an issue with the SDK we should be fixing that

docs/images/logo-branding.png
b-d055, line 1: I don't think these docs images are needed, but they're also not hurting anything living here I guess

b-d055 (commented):
@yanivtal  some comments added for consideration

b-d055:
closing in favor of #1880

PR #1877, opened 2026-06-15
feat(diff): use v2 versioned endpoint for entity diffs; drop client-side postProcessDiffs

0xNeoDev:
## Summary

Convert the entity-diff path to the gaia **`/v2/versioned/entities/:id/diff`** endpoint and drop the client-side `postProcessDiffs` pipeline for it. v2 resolves names, inlines media URLs at the from/to version, folds the rich block shape, and merges data-block config server-side, so the decoded response maps straight to an `EntityDiff`.

## Changes (3 files, +62 / −100)
- **`core/io/rest/schemas/diff-shared.ts`** — add optional v2 response fields: value `propertyName`; relation `typeName` + endpoint `toEntityName`/`imageUrl`/`videoUrl`; block `blockName`/`values`/`relations`.
- **`core/utils/diff/diff.ts` (`mapApiEntityDiff`)** — read those fields; normalize GRC-20 value-type names (`BOOL`/`INT64`/`FLOAT64` → `BOOLEAN`/`INTEGER`/`FLOAT`); fold the rich `dataBlock` shape.
- **`core/io/subgraph/fetch-entity-diff.ts`** — `fetchEntityDiff` now hits `/v2/versioned` and maps the response directly; `fromEditId` optional → snapshot mode. Removes `postProcessDiffs`, block discovery, per-block fetches, and the snapshot-to-diff path.

## Scope / notes
- `postProcessDiffs` is **retained** for the proposal-diff path (`fetch-proposal-diffs.ts`) — there's no v2 *proposals* endpoint yet. The mapper changes are null-safe and additive, so the proposal path is unchanged.
- Requires the gaia v2 endpoint deployed (defi-wonderland/gaia#215 / geobrowser/gaia#750).
- Verified against staging: existing `/versioned` (v1) output is byte-identical pre/post, and the v2 path renders names / blocks / media correctly with no console errors.

## Follow-up

This PR converts only the **single-entity** diff path. Follow-up work will apply the
same approach to the other diff endpoints — notably the **proposal diff**
(`fetch-proposal-diffs.ts`), which still runs the client-side `postProcessDiffs`
pipeline — once the corresponding gaia v2 endpoints exist
(`/v2/versioned/proposals/:id/diff`, `/proposal-groups/diff`). At that point
`postProcessDiffs` can be removed entirely.


ohohoreilly:
Closing this, and the reason is our side rather than yours.

This has been open since 15 June with no review and no response. It was also never able to build a preview — the Vercel check has been sitting on "a member of the Team first needs to authorize it" the whole time, and nobody did. Three months without an answer is not a decision you should have to keep waiting on, so I would rather close it honestly than leave it open as though a review were still coming.

For the record, the work was not wrong. Build, Test and Lint are all green, and the branch **still applies cleanly to master today**, three months on.

Where things stand:

- **The endpoint migration landed anyway**, independently. `fetchRawEntityDiff` now calls `/versioned/entities/:id/diff` (`fetch-entity-diff.ts:105`).
- **Your other point did not, and is still true.** Master runs `Diff.postProcessDiffs` over the v2 response in both branches of `fetchEntityDiff` (lines 41 and 71) — the client-side pass you argued v2 makes redundant.

I have written that second half up as **GEO-2982**, crediting this PR, including the schema additions in `diff-shared.ts` as the reusable part and the three things a reviewer would need to confirm before deleting the client pass (`postProcessDiffs` is shared with the proposal path, which is not on v2, so it is not a straight deletion).

So the analysis is kept and attributed. Thanks for it, and sorry it sat this long — if you want to pick up GEO-2982, say so and I will make sure a reviewer and a Vercel authorization are lined up first this time.

ohohoreilly:
Correction to my comment above, having now tested the claim rather than taken it at face value.

I said the second half of this PR "did not [land], and is still true", and pointed GEO-2982 at it. That was wrong, and I should not have said it without checking. **The premise does not hold, and merging this would have been a regression.**

I pulled 14 real diffs from the live v2 endpoint, across several entities with genuine multi-edit history:

- diffs carrying **any** resolved name (`propertyName`, `typeName`, `toEntityName`): **0 of 14**
- diffs with a non-empty `blocks[]`: **0 of 14**

A representative body with real content:

```
"values": [{"propertyId": "94e43fe8…", "spaceId": "41e85161…", "type": "DATETIME", "before": null, "after": "2026-07-25T04:40:00Z"}]
"relations": [{"relationId": "60ebdfc3…", "typeId": "8f151ba4…", "spaceId": "41e85161…", "changeType": "ADD",
               "before": null, "after": {"toEntityId": "a2a5ed0c…", "toSpaceId": null, "position": null}}]
```

Raw ids throughout — no names, no inlined media URLs, no folded blocks. Resolving those ids is exactly what `postProcessDiffs` does (`diff.ts:236-258`), so dropping it would have rendered diffs as raw UUIDs — which is GEO-1891, "Diffs showing entity ID instead of name", a bug that has already been fixed once.

The endpoint migration in this PR was right and is on master. The `postProcessDiffs` removal was not. GEO-2982 is closed as won't-do with the measurements recorded.

The substantive point stands unchanged, though: this sat for three months without anyone testing the claim either way, and that is the part that was our failure rather than yours. Sorry for the earlier comment overstating it in the other direction — I would rather correct it than leave a wrong note on your PR.

PR #1913, opened 2026-06-18
fix: open a tab to view someone's ranking

webmagic123:
Feature: 
A "Ranked by" dropdown showing who voted on a ranking, Clicking a voter opens their ranking tab,

Fixed:
1. Clicking someone in the dropdown swaps to their ranking.
2. Swap rankings in place - switching between voters doesn't tear the UI down and rebuild it.
3. Speed up the dropdown's avatars
 3.1 only fetch a voter's space when their profile has no avatar
 3.2 don't fetch voter spaces at all until profiles have loaded
 3.3 share the rank-entity space resolution instead of resolving it per voter
4. Chunk the space-id requests — voter space lookups were sent as one unbounded query; they're now batched at a fixed limit.
5. Resolve pending entities on my own ranking tab.


apps/web/core/blocks/ranking/use-ranking-voters.ts
```diff
@@ -0,0 +1,130 @@
...
+  });
+
+  const { data: spacesById = EMPTY_SPACE_MAP } = useQuery({
+    queryKey: ['ranking-voter-spaces', spaceIds],
+    enabled: spaceIds.length > 0,
+    staleTime: 60_000,
+    queryFn: async ({ signal }) => {
+      const batches = chunkSpaceIds(spaceIds);
+      const results = await Promise.all(
+        batches.map(batch => Effect.runPromise(getSpaces({ spaceIds: batch }, signal)))
```
b-d055, line 91: **Blocker (performance).** This fetches a full `FullSpace` per voter to read exactly one field — `?.entity.image` on line 113.

`spacesQuery` → `FullSpace` (`core/io/query-fragments.tsx:426-457`) embeds `topic { ...FullEntity }` *and* `page { ...FullEntity }`, and each `FullEntity` is `valuesList(first: 1000)` + `relationsList(first: 1000)` with nested `toEntity.valuesList` (`query-fragments.tsx:71-94`), plus `membersList`/`editorsList`. All of it is then decoded through `SpaceEntityDto` so that `Entities.avatar(entity.relations)` can yield one URL (`core/io/dto/spaces.ts:58`).

The pre-existing "Ranked by" avatar group paid this cost for at most 3 spaces (`ranking-period-metadata.tsx:90,100`); this pays it for every voter, uncapped.

Mostly avoidable: `/profile/batch` already returns `avatarUrl` (`core/io/subgraph/fetch-profile.ts:189`), and the space image is only consulted when the profile avatar is missing (lines 112-114). Suggest deriving this query's input from the subset of `spaceIds` whose resolved profile came back without a usable `avatarUrl` — normally empty, so the query drops out entirely in the common case.

Mitigating context: this fires per popover-open rather than per page-load (`RankingVotersList` only mounts while open, `ranking-voters-pill.tsx:58`), and the spaces query doesn't gate the spinner, so avatars fill in late rather than blocking.

apps/web/core/blocks/ranking/use-ranking-voters.ts
```diff
@@ -0,0 +1,130 @@
...
+function chunkSpaceIds(spaceIds: string[]): string[][] {
+  const batches: string[][] = [];
+  for (let start = 0; start < spaceIds.length; start += SPACE_ID_REQUEST_LIMIT) {
+    batches.push(spaceIds.slice(start, start + SPACE_ID_REQUEST_LIMIT));
+  }
+  return batches;
+}
+
+export function useRankingVoters(refs: AggregatedRankingSubmitterRef[]) {
+  const rankEntityIdsNeedingSpace = React.useMemo(
```
b-d055: **Refactor (duplication), non-blocking.** Lines 41-59 are a byte-identical copy of `use-ranking-submitter-space-ids.ts:20-40` — same `rankEntityIdsNeedingSpace` memo, same `useQueryEntities` call, same `rankEntitySpaceById` build — diverging only afterward in output shape (`string[]` vs `{rankEntityId, spaceId}[]`).

Worth stating that this costs no extra request: `GeoStore.queryKeys` runs `where` through `stableStringify` (`core/sync/store.ts:114-133,305-307`), and both callers pass identical `where`/`first` derived from the same `refs` array, so React Query dedupes. Purely a maintenance concern — the space-resolution rule ("`to_space`, else the rank entity's home space") now lives in two places and will drift.

Suggest extracting a shared `useRankEntitySpaceById(refs): Map<string, string>` and building both outputs on top of it.

One caution for whoever picks this up: please **don't** swap the chunked spaces query below for `useSpacesByIds`. That hook doesn't chunk (`core/hooks/use-spaces-by-ids.ts:24`) and the API pages at 100, so it would silently truncate past 100 voters. The `SPACE_ID_REQUEST_LIMIT = 100` chunking here is a correctness win over that hook — if anything the chunking should move *into* `useSpacesByIds` so its other call sites benefit.

apps/web/partials/blocks/table/use-ranking-block-state.ts
```diff
@@ -546,6 +566,55 @@ export function useRankingBlockState({
...
 
+  const showViewerOwnTab = isSharedRankingView && hasMySubmission;
+  const viewerOwnDisplayEntityIds = React.useMemo(
+    () => (showViewerOwnTab ? (mySubmission?.orderedEntityIds ?? EMPTY_ENTITY_IDS) : EMPTY_ENTITY_IDS),
+    [showViewerOwnTab, mySubmission]
+  );
+
+  const { entries: viewerOwnEntries } = useRankingEntryEntities(spaceId, viewerOwnDisplayEntityIds);
+
+  const viewerOwnEntryByEntityId = React.useMemo(() => {
```
b-d055: **Refactor (duplication) + a behavioral gap it hides, non-blocking.**

Both loops here are byte-identical to `globalRankingEntryByEntityIdBase` (378-396) and `myRankingEntryByEntityIdBase` (410-428) — same `row.placeholder` guard, same `fromRow` description/image merge — differing only in which entries array is overlaid and in having no seed. A `buildRankingEntryMap(seedEntries, rows, entries)` helper collapses all three losslessly (this one passes `EMPTY_RANKING_ENTRIES`).

More consequential: unlike the other two, this map never receives the pending-proposal backfill applied at 545-561, and `viewerOwnDisplayEntityIds` can't reach `pendingCandidateEntityIds` — `ownRankingEntityIds` (515-518) returns empty whenever `isSharedRankingView` is true, which is precisely when `showViewerOwnTab` is true (569).

Net effect: a governance-pending entity in the viewer's own ranking renders "Untitled" on the `viewer` tab while the same entity resolves correctly on the `my` and `global` tabs. Since `useRankingEntryEntities` drops ids it can't resolve, an entity missing from both `rows` and `viewerOwnEntries` renders a skeleton row indefinitely.

Suggest renaming this to `viewerOwnEntryByEntityIdBase`, adding `viewerOwnDisplayEntityIds` to `pendingCandidateEntityIds`, and deriving the final map through the same `pendingEntriesByEntityId` overlay as the other two.

b-d055 (changes requested):
@webmagic123  see comments, I'd only consider the first blocking, the other two are nice to have improvements

webmagic123:
@b-d055 
- Feature: a "Ranked by" dropdown showing who voted on a ranking, Clicking a voter opens their ranking tab, 
- What I've implemented.
1. Clicking someone in the dropdown swaps to their ranking.
2. Swap rankings in place -  switching between voters doesn't tear the UI down and rebuild it.
3. Speed up the dropdown's avatars
  3.1 only fetch a voter's space when their profile has no avatar
  3.2 don't fetch voter spaces at all until profiles have loaded
  3.3 share the rank-entity space resolution instead of resolving it per voter
4. Chunk the space-id requests — voter space lookups were sent as one unbounded query; they're now batched at a fixed limit.
5. Resolve pending entities on my own ranking tab.

apps/web/partials/blocks/table/use-ranking-block-state.ts
```diff
@@ -163,16 +211,24 @@ export function useRankingBlockState({
     submittedAtMs,
   } = useRankingSubmissions(entityId, spaceId, displayName);
 
+  const [sharedRankOverride, setSharedRankOverride] = React.useState<{
+    rankEntityId: string;
+    authorSpaceId: string;
+    authorName?: string | null;
+  } | null>(null);
+  const activeSharedRankEntityId = sharedRankOverride?.rankEntityId || sharedRankEntityId;
```
b-d055, line 219: When a voter is selected here and their ranking resolves to null (`getEntity` fails, the indexer lags, or `isRankSubmittedToBlock` returns false), `myDisplayEntityIds` falls back to `initialSharedOrderedIds` (line 357). Those are the server-resolved ids for the ranking in the URL, so the tab shows the original ranking's items under the selected voter's name. `myRankingEntryByEntityIdBase` also seeds its entries from `initialSharedEntries` (line 443).

Repro: open `/r/{A}`, pick voter B from Ranked by, and have B's `shared-ranking` query return null. The tab reads "B's ranking" and lists A's items.

Suggest ignoring the server seed once an override is set:
```ts
const initialSharedOrderedIds = sharedRankOverride ? EMPTY_ENTITY_IDS : (initialSharedRanking?.orderedEntityIds ?? EMPTY_ENTITY_IDS);
const initialSharedEntries = sharedRankOverride ? EMPTY_RANKING_ENTRIES : (initialSharedRanking?.entries ?? EMPTY_RANKING_ENTRIES);
```

apps/web/core/blocks/ranking/use-ranking-voters.ts
```diff
@@ -0,0 +1,122 @@
...
+          name: profile?.name ?? null,
+          avatarUrl: profileAvatarUrl ?? spaceAvatarUrl,
+          fallbackSeed: profile?.address ?? ref.spaceId,
+          address: profile?.address ?? null,
+        };
+      }),
+    [resolvedRefs, profilesBySpaceId, spacesById]
+  );
+
+  const isLoading = spaceIds.length > 0 && isLoadingProfiles;
```
b-d055, line 119: `isLoading` only covers the profiles query. When refs have no `spaceId`, `spaceIds` stays empty until `useRankEntitySpaceById` resolves each rank entity's home space. During that time `isLoading` is false, so the popover renders no rows and no skeleton, only the "N voters" footer. If that lookup fails, the list stays empty.

This is the common path on testnet right now: all 390 Aggregated rankings relations there have a null `toSpaceId`.

Suggest returning the entities query's loading state from `useRankEntitySpaceById` and including it here, for example:
```ts
const isLoading = isResolvingSpaces || (spaceIds.length > 0 && isLoadingProfiles);
```

b-d055 (commented):
@webmagic123  see comments

apps/web/partials/blocks/table/use-ranking-block-state.ts
webmagic123: @b-d055 
Updated.

apps/web/core/blocks/ranking/use-ranking-voters.ts
webmagic123: @b-d055 
Updated.

PR #1951, opened 2026-06-30
fix: use datetime properties for ranking block start/end times

webmagic123:


b-d055 (dismissed):
see update below

apps/web/core/ranking-block-ids.ts
```diff
@@ -5,8 +5,8 @@ export const RANKING_BLOCK_TYPE_ID = '150db6defe2344f0805afa57502e2c32';
 
 export const RANKING_BLOCK_TYPE_NAME = 'Ranking Block';
 
-export const RANKING_START_DATE_PROPERTY_ID = 'eed03a040acd4a9e81e08272ed70a817';
-export const RANKING_END_DATE_PROPERTY_ID = 'b08b8f63dc1e41568b0819946f2b011c';
+export const RANKING_START_TIME_PROPERTY_ID = '2d696bf0510f403e985b8cd1e73feb9b';
+export const RANKING_END_TIME_PROPERTY_ID = 'c3445f6be2c04f25b73a5eb876c4f50c';
```
b-d055, line 9: **Blocking: no fallback or migration for the old property IDs.**

The old IDs (`eed03a040acd4a9e81e08272ed70a817` and `b08b8f63dc1e41568b0819946f2b011c`) shipped in #1849, and after this change nothing reads them. `useRankingBlockDates`, `use-editor.tsx`, and `resolve-ranking-share.ts` all query only the new IDs.

So every ranking block already published under the old IDs reads back empty:

- `getRankingPeriodState('', '')` returns `'no-date'`
- `rankingSubmissionsOpen('no-date')` returns `true`

Existing rankings lose their period label and their OG card dates, and any ranking that had already ended silently reopens for votes. This needs a read fallback (try the new ID, then the old one) or a data migration before it ships.

apps/web/core/blocks/ranking/ranking-period.ts
```diff
@@ -38,17 +77,14 @@ export function formatRankingPeriodLabel(
...
     case 'in-progress': {
       const end = parseRankingDate(endDate);
       if (!end) return null;
-      const days = differenceInCalendarDays(end, now);
-      if (days > 0) return days === 1 ? 'Ends in 1 day' : `Ends in ${days} days`;
-      if (days === 0) return 'Ends today';
-      return null;
+
+      const label = formatRelativeCountdown('Ends', end, now);
+      return label ?? 'Ended';
```
b-d055, line 107: The label path is missing the date-only handling that `getRankingPeriodState` got, so the two disagree on the final day.

With `endDate = '2026-06-04'` at `now = 2026-06-04T12:00Z`, `hasRankingEnded` compares calendar days and returns `in-progress`, so submissions stay open. But `parseRankingDate` resolves the end to midnight UTC, `formatRelativeCountdown` sees `msRemaining <= 0` and returns `null`, and this fallback renders **Ended** next to an open submission form for the whole day. The old code returned `'Ends today'` here.

The day before is wrong in the other direction: the label counts down to midnight ("Ends in 30 min") while the period actually runs through all of the end day.

This path is dead code today, since nothing reads the old date-only IDs. It goes live the moment the read fallback above is added, so the two are worth fixing together.

apps/web/design-system/editable-fields/date-field.tsx
```diff
@@ -749,13 +751,14 @@ function DateTimeInput({ variant, initialDate, onDateChange, label }: DateInputP
       timeFormState.isValid;
 
     if (isValid) {
-      const isoString = GeoDate.toISOStringUTC({
+      const parts = {
         day: newDay,
         month: newMonth,
         year: newYear,
         minute: newMinute,
         hour: newMeridiem === 'am' ? newHour : (Number(newHour) + 12).toString(),
```
b-d055: Empty hour plus a "pm" toggle stores noon instead of midnight.

`onToggleMeridiem` calls `updateDate` unconditionally, and an empty hour passes both `hourValidator` and the `isValidHourCheck` above, so `isValid` stays true. `Number('') + 12` is `12`, so a user who picks a date and taps "pm" without typing an hour gets 12:00 local. The "am" path serializes `''` to `'00'` and gets midnight.

This is pre-existing on master via the UTC path, but this PR puts it on the ranking setup flow.

b-d055 (changes requested):
This works as expected for a newly created ranking block. The issues below are the ones a fresh manual test can't surface, because they only show up on pre-existing data or after time passes. Details are in the inline comments, plus one blocker that lives outside this diff.

## Blocking: the countdown never ticks, and the publish gate is client-only

`apps/web/partials/blocks/table/use-ranking-block-state.ts:199` and `apps/web/partials/rankings/ranking-compose-screen.tsx:117`

`periodState` and `periodLabel` are computed in a `useMemo` keyed on `[startDate, endDate]`, with `now` captured at evaluation time and no interval anywhere. Open the page five minutes before close and "Ends in 5 min" stays on screen indefinitely.

`rankingSubmissionsOpen` gates `canPublish` client-side (`ranking-compose-screen.tsx:462`) with no server-side check, so a user sitting on the compose screen can still publish after the ranking has actually closed. Day-granularity labels made this tolerable. Minute-precision labels don't. A ticking `now` fixes the display, but the publish-after-close hole probably wants a server-side check regardless.

apps/web/core/io/subgraph/fetch-featured-rankings.ts
```diff
@@ -156,8 +156,8 @@ export async function fetchFeaturedRankings(): Promise<FeaturedRanking[]> {
       const entity = page.entity;
       const relations = page.relations.length > 0 ? page.relations : entity.relations;
 
-      const rankingStartDate = readDateValue(entity, RANKING_START_DATE_PROPERTY_ID, spaceId);
-      const rankingEndDate = readDateValue(entity, RANKING_END_DATE_PROPERTY_ID, spaceId);
+      const rankingStartDate = readDateValue(entity, LEGACY_RANKING_END_DATE_PROPERTY_ID, spaceId);
+      const rankingEndDate = readDateValue(entity, LEGACY_RANKING_START_DATE_PROPERTY_ID, spaceId);
```
b-d055: Looks like these two are swapped, start is reading the legacy end property and vice versa. Also seems like this only reads the legacy IDs, so blocks created after this PR would come back empty here and never leave the featured list. Maybe this could go through the same fallback resolve-ranking-share uses?

apps/web/core/blocks/ranking/ranking-period.ts
```diff
@@ -1,27 +1,85 @@
...
+    if (isDateOnlyValue(trimmed)) return parseISO(trimmed);
+    return parseISO(GeoDate.toFullISOString(trimmed));
   } catch {
     return null;
   }
 }
 
+/** Date-only values (legacy DATE properties) are compared by calendar day; datetimes use timestamps. */
+function isDateOnlyValue(value: string): boolean {
+  return /^\d{4}-\d{2}-\d{2}$/.test(value.trim());
```
b-d055: I might be wrong here, but I don't think this regex matches what's actually stored. The old date-only input seems to have saved full ISO strings like `2026-06-30T00:00:00.000Z`, so existing rankings would take the datetime path and end at midnight UTC at the start of their end day, earlier than before.

apps/web/design-system/editable-fields/date-field.tsx
```diff
@@ -779,13 +781,14 @@ function DateTimeInput({ variant, initialDate, onDateChange, label }: DateInputP
...
```
b-d055: Looks like the `+ 12` mishandles 12 o'clock: 12:30 PM would become hour 24 and roll into the next day, and 12 AM would store noon. The empty hour + pm case from my last review also still seems to land here.

apps/web/core/blocks/ranking/use-ranking-period.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export function useRankingPeriod(startDate: string, endDate: string) {
+  const [now, setNow] = React.useState(() => new Date());
+
+  React.useEffect(() => {
+    let timer: number | undefined;
+
+    const scheduleFrom = (from: Date) => {
+      const delay = msUntilRankingPeriodChange(startDate, endDate, from);
+      if (delay === null) return;
```
b-d055, line 35: If this returns null, it looks like `now` never gets refreshed, so if the dates arrive later from sync the memos below compute against the mount-time clock and an ended ranking could still show as open until a remount.

b-d055 (changes requested):
One more that I can't comment inline: fetch-featured-rankings.test.ts still imports RANKING_START_DATE_PROPERTY_ID / RANKING_END_DATE_PROPERTY_ID, which got renamed to LEGACY_*, so tsc/vitest fail on this tip. Fixing those imports would also have caught the swapped reads below.

b-d055 (approved):

PR #1961, opened 2026-07-06
feat: add Score as default block property and gate vote buttons on it

webmagic123:


apps/web/core/blocks/data/is-score-visible-on-browse-view.ts
```diff
@@ -0,0 +1,18 @@
...
+export function isScoreVisibleOnBrowseView(shownColumnIds: readonly string[], blockRelationId: string | undefined): boolean {
+  if (isScorePropertyShown(shownColumnIds)) return true;
+  if (!blockRelationId) return true;
+
+  const scoreRelations = store.getResolvedRelations(blockRelationId, true).filter(
+    r => isShownColumnRelationType(r.type.id) && ID.equals(columnPropertyIdFromRelation(r), SCORE_SYSTEM_PROPERTY)
+  );
+
+  if (scoreRelations.some(r => r.isDeleted)) return false;
+  return true;
```
b-d055, line 17: This helper defaults to returning true, so vote buttons show even when Score was never added to Properties (the only false path requires a soft-deleted Score relation). The issue says votes shouldn't show if Score isn't in the properties list, so I'd expect this to work like isScorePropertyShown instead.

Right now the views are split: list, gallery, and ranking-entry-row use this lenient check, while table, bulleted list, and ranking-list-view use the strict one. So the same block can show votes in one view and hide them in another. Is the lenient behavior intentional for browse views? If so it should probably apply to table/bulleted too, and if not, collapsing everything onto isScorePropertyShown would fix the inconsistency and also drop the store read this does during render.

One caveat on impact: editors on list/gallery get Score auto-seeded by the effect in table-block.tsx, and new blocks seed it via DEFAULT_BLOCK_SHOWN_PROPERTIES, so in practice this mostly affects non-editors viewing older blocks that never had Score.

apps/web/partials/blocks/table/table-block-list-item.tsx
```diff
@@ -62,8 +67,12 @@ export function TableBlockListItem({
   autoFocus = false,
   focusRequestKey,
   collectionTypeFilters,
+  listRank,
 }: Props) {
   const { storage } = useMutate();
+  const { relationId: blockRelationId } = useDataBlockInstance();
+  const { shownColumnIds } = useView();
+  const showVoteButtons = isScoreVisibleOnBrowseView(shownColumnIds, blockRelationId);
```
b-d055: Same question as on is-score-visible-on-browse-view.ts, this shows votes even when Score isn't in Properties. Should this be isScorePropertyShown(shownColumnIds) to match the table view?

apps/web/partials/blocks/table/table-block-gallery-item.tsx
```diff
@@ -64,6 +67,9 @@ export function TableBlockGalleryItem({
   collectionTypeFilters,
 }: Props) {
   const { storage } = useMutate();
+  const { relationId: blockRelationId } = useDataBlockInstance();
+  const { shownColumnIds } = useView();
+  const showVoteButtons = isScoreVisibleOnBrowseView(shownColumnIds, blockRelationId);
```
b-d055: Same as the list item, gallery uses the lenient check here. If the strict behavior is what we want, this needs isScorePropertyShown too.

apps/web/partials/blocks/table/ranking-entry-row.tsx
```diff
@@ -58,6 +63,10 @@ export function RankingEntryRow({
   linkToEntity = true,
   rankStyle = 'avatar-badge',
 }: Props) {
+  const { relationId: blockRelationId } = useDataBlockInstance();
+  const { shownColumnIds } = useView();
+  const showVoteButtons = isScoreVisibleOnBrowseView(shownColumnIds, blockRelationId);
```
b-d055: This uses the lenient check while ranking-list-view.tsx:89 uses the strict isScorePropertyShown. They're alternate ranking sub-views so they never render together, but given the same block config (Score not shown, no deleted Score relation) this row shows votes and the list view hides them. Whichever behavior is right, these two should probably agree.

apps/web/partials/blocks/table/table-block.tsx
```diff
@@ -690,16 +696,36 @@ const ConfiguredTableBlock = ({
...
+  );
+
+  // List/gallery: persist Score when missing. Soft-deleted Score (user hid it) is not re-added.
+  const hasScoreShown = isScorePropertyShown(shownColumnIds);
+  const scoreBrowseDefault = isScoreVisibleOnBrowseView(shownColumnIds, relationId);
+  React.useEffect(() => {
+    if (!canEdit) return;
+    if (view !== 'LIST' && view !== 'GALLERY') return;
+    if (!relationId || hasScoreShown) return;
+    ensureScoreShownColumn({
```
b-d055, line 726: I don't think removing Score survives a publish plus reload. deleteRelation only writes a local soft-delete tombstone, which is what the guard in ensureScoreShownColumn checks. Once the deletion is published the tombstone is gone on a fresh load, getResolvedRelations returns no Score relations, and this effect re-adds Score for editors. So a user who hides Score on a list/gallery block will see it come back after publishing. Is there a plan for persisting the removal, or am I missing a durable signal somewhere?

apps/web/partials/blocks/table/table-block.tsx
```diff
@@ -690,16 +696,36 @@ const ConfiguredTableBlock = ({
...
+      spaceId,
+      relationsIncludingDeleted: store.getResolvedRelations(relationId, true),
+    });
+  }, [canEdit, view, relationId, hasScoreShown, storage, spaceId]);
+
+  React.useEffect(() => {
+    if (view !== 'LIST' && view !== 'GALLERY') return;
+    if (source.type === 'COLLECTION') return;
+    if (!shouldApplyDefaultScoreSort(sortState, scoreBrowseDefault)) return;
+    setSortState(DEFAULT_SCORE_SORT_STATE);
```
b-d055, line 744: Once this kicks in there's no way back to unsorted order. Clearing the sort persists an empty string which parses back to null, so shouldApplyDefaultScoreSort fires again and re-applies score desc. That means every non-collection list/gallery block gets score sorting whether or not the user wants it, including pre-existing blocks. I get that default score sorting is the point of this commit, but should clearing the sort stick? Not sure what the intended behavior is here, worth confirming.

apps/web/partials/blocks/table/table-block-list-item.tsx
```diff
@@ -105,14 +114,64 @@ export function TableBlockListItem({
...
+      c.slotId !== SystemIds.DESCRIPTION_PROPERTY &&
+      c.slotId !== SCORE_SYSTEM_PROPERTY
   );
 
   const imageUrl = useImageUrlFromEntity(image || undefined, currentSpaceId || '');
   if (image && imageUrl) {
     image = imageUrl;
   }
 
+  const useRankedScoreListLayout = !isEditing && showVoteButtons && listRank != null && !isPlaceholder;
```
b-d055: When this layout kicks in it only renders rank, name, and votes, dropping the image, description, and any other shown columns. Since showScoreRankedList ends up true by default for non-collection list blocks in browse mode, existing list blocks will look different after this ships. Is that the intended design? Checking since the issue only mentions gating the vote buttons.

b-d055 (changes requested):
The Score property work itself looks good, but I have questions about the vote gating on a few views, plus a durability issue with removing Score. Details inline.

apps/web/partials/blocks/table/table-block.tsx
webmagic123: @b-d055 
implemented  persisting feature.

apps/web/partials/blocks/table/table-block.tsx
```diff
@@ -702,16 +709,40 @@ const ConfiguredTableBlock = ({
...
+    () => mergeBlockMenuProperties(filterableProperties, properties),
+    [filterableProperties, properties]
+  );
+
+  // List/gallery: persist Score when missing, unless the user durably dismissed it.
+  const hasScoreShown = isScorePropertyShown(shownColumnIds);
+  const scoreDismissed = useScoreDismissed();
+  const scoreBrowseDefault = isScoreVisibleOnBrowseView(shownColumnIds, relationId);
+  React.useEffect(() => {
+    if (!canEdit) return;
```
b-d055, line 723: Going into edit mode now shows an edit I never made. Just loading a page that has a list or gallery block is enough: toggle edit mode afterwards and there's a pending Score property plus a Score sort on the block, already checked and ready to publish.

The cause is that `canEdit` here is `useCanUserEdit`, which is only permission (`isMemberOfDao || isEditorOfDao`). The edit-mode hook is `useUserIsEditing`, which this component already binds at :559. So both effects run on a plain page load for anyone with access. The sort effect just below has no gate at all, and `useSort(canEdit)` hands members the persisting setter, so that one writes too.

Could these gate on `useUserIsEditing` instead, or hold off the write until the user actually edits the block?

b-d055 (changes requested):
@webmagic123  one more issue

b-d055:
no longer needed

PR #1977, opened 2026-07-10
feat: show link ingestion tool on data blocks via Tools relation

webmagic123:


apps/web/partials/blocks/table/table-block-ranking.tsx
```diff
@@ -92,152 +97,135 @@ export function TableBlockRanking({ spaceId, rankingStartDate = '', rankingEndDa
...
+                      />
+                    ))}
+                  </div>
+                )}
+              </div>
+            </motion.div>
+          </AnimatePresence>
+        )}
+
         <RankingBlockBody state={state} presentation="embedded" />
```
b-d055: I think the rebase dropped some code here. On current master this renders RankingGalleryView / RankingListView / RankingPillView / RankingExploreView depending on the active view, falling back to RankingBlockBody. Here RankingBlockBody renders unconditionally, and RankingBlockBody doesn't dispatch to those views internally, so gallery/list/pill/explore ranking views no longer render. That includes the Explore view the news stories block uses, which this feature targets.

The view components and the isExploreView / isListView / isPillView / isGalleryView flags are still imported and computed above but never used now, and DataBlockViewMenu and RankingHeaderActions also went missing from the header in the same commit. Looks like a conflict resolution accident rather than intentional, but let me know if I'm wrong. The old ternary should be recoverable from master.

apps/web/partials/blocks/table/block-link-ingestion-tool.tsx
```diff
@@ -0,0 +1,210 @@
...
+}
+
+type ProviderProps = {
+  spaceId: string;
+  children: React.ReactNode;
+};
+
+export function BlockLinkIngestionProvider({ spaceId, children }: ProviderProps) {
+  const { hasLinkIngestionTool } = useBlockTools();
+  const [expanded, setExpanded] = React.useState(false);
```
b-d055, line 37: The panel starts collapsed and only opens when the chip is clicked. Issue describes the tool as collapsible but open by default. I can't tell if the collapsed chip is the newer intended design or not, can someone confirm? If it should be open by default this just needs useState(true).

apps/web/partials/blocks/table/block-link-ingestion-tool.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const setChatOpen = useSetAtom(isChatOpenAtom);
+  const [url, setUrl] = React.useState('');
+  const [submitting, setSubmitting] = React.useState(false);
+
+  if (!enabled || !canEdit || !expanded) return null;
+
+  const normalizedUrl = normalizeHttpUrl(url);
+  const canSubmit = normalizedUrl !== null && !submitting;
+
+  const handleSubmit = async (event: React.FormEvent) => {
```
b-d055: normalizeHttpUrl (line 16) and this handleSubmit are copied nearly verbatim from add-data-panel.tsx (only the console.warn tag differs). Could we pull the shared logic into a hook or util? Otherwise a fix to the ingestion flow in one place will get missed in the other.

apps/web/core/blocks/data/block-ontology-ids.ts
```diff
@@ -1,7 +1,13 @@
...
 export const MAX_DATA_BLOCK_PAGE_SIZE = 100;
+export const DATA_BLOCK_PAGE_SIZE_PROPERTY_ID = 'fd8d9863342d4dcea8fd9687805a88a4';
+export const DATA_BLOCK_VIEW_PROPERTY_ID = '1907fd1c81114a3ca378b1f353425b65';
+export const DATA_BLOCK_PROPERTIES_PROPERTY_ID = '01412f8381894ab1836565c7fd358cc1';
+export const DATA_BLOCK_TOOLS_PROPERTY_ID = '721ac0d146404b0f93edf52535151d19';
+
+export const GALLERY_VIEW_ID = 'ccb70fc917f04a54b86e3b4d20cc7130';
+export const LIST_VIEW_ID = '7d497dba09c249b8968f716bcf520473';
+export const PILL_VIEW_ID = '60022d24d9df4e1888faa71e915f8101';
+export const EXPLORE_VIEW_ID = 'f4063f3d5e904e4694ab04729caad52b';
```
b-d055: These six constants aren't referenced anywhere, and four of them duplicate existing SystemIds values (VIEW_PROPERTY, PROPERTIES, GALLERY_VIEW, LIST_VIEW). Only DATA_BLOCK_TOOLS_PROPERTY_ID and LINK_INGESTION_TOOL_ID are actually used, in use-block-tools.ts. Can we drop the unused ones and lean on SystemIds where they exist, so we don't end up with two sources of truth for the same IDs?

b-d055 (changes requested):
The link ingestion mechanics look right, IDs match the issue and the gating on the Tools relation works. Main concern is that the rebase seems to have dropped the ranking view dispatch from `table-block-ranking.tsx`. Details inline.

b-d055 (approved):
lgtm if conflicts are resolved @webmagic123

webmagic123:
@b-d055 
Resolved conflicts.

PR #1996, opened 2026-07-15
feat: claim-aware vote controls with upvote % and factual check/x var…

webmagic123:
…iant


apps/web/partials/entity-page/entity-interaction-bar.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * data blocks, in order: votes, the voters who ranked it, comments, and — with the
+ * debates flag on — the debate megaphone.
+ */
+export function EntityInteractionBar({ entityId, spaceId, commentCount, className }: EntityInteractionBarProps) {
+  const isClaim = useIsClaimEntity(entityId, spaceId);
+  const { data: fetchedCommentCount = 0 } = useQuery({
+    queryKey: ['entity-comment-count', entityId],
+    queryFn: ({ signal }) => Effect.runPromise(getEntityCommentCount(entityId, signal)),
+    staleTime: 60_000,
+    enabled: commentCount === undefined,
```
b-d055: I think this fires the comment count query for every entity that mounts the bar, not just claims, since `enabled` only checks `commentCount === undefined`. The `if (!isClaim) return null` below skips the render but the fetch already went out. And `getEntityCommentCount` isn't cheap, it pages through backlinks (1000 per page) and returns the length, so non-claim rows end up paying for a query they throw away. This looks like it'd hit `ranking-block-body`, `ranking-list-view`, `ranking-gallery-view`, and the table-block list/gallery/bulleted items, since those render the bar unconditionally (table-block-table gates it with `ClaimInteractionRow`). Could we add `isClaim` to the `enabled` check?

apps/web/partials/entity-page/entity-vote-buttons.tsx
```diff
@@ -193,8 +227,54 @@ export function EntityVoteButtons({ entityId, spaceId }: EntityVoteButtonsProps)
...
+      entityId={entityId}
+      spaceId={spaceId}
+      objectType={ENTITY_VOTE_OBJECT_TYPE}
+      totalVoters={totalVoters}
+    />
+  ) : null;
+
+  const claimVoterAvatarsClassName = 'inline-flex h-5 shrink-0 items-center';
+
+  if (hideWhenClaim && isClaim) return null;
```
b-d055: One thing I noticed: this bailout runs after `useQueryEntity` and both vote queries, so a claim row still fetches all of that and then renders nothing, while `EntityInteractionBar` mounts a second `EntityVoteButtons` for the same entity. The query keys dedupe the actual network calls so it's not double requests, but it is a couple of extra render trees and subscriptions per claim row. Might be cleaner to check `useIsClaimEntity` in the caller and do `{!isClaim && <EntityVoteButtons … />}` rather than hiding it from inside. Not a blocker, just a thought.

apps/web/partials/entity-page/entity-vote-buttons.tsx
```diff
@@ -30,18 +37,45 @@ type OptimisticVote = 0 | 1 | 'none' | null;
...
     spaceId,
   });
   const { smartAccount } = useSmartAccount();
   const { isPending: isAccountSetupPending } = usePendingPersonalSpace();
+
+  // Claim entities render a different vote control.
+  const { entity } = useQueryEntity({ id: entityId, spaceId });
+  const isClaim = entity?.types.some(t => normalizeId(t.id) === CLAIM_TYPE) ?? false;
+  const isFactualClaim =
+    isClaim && parseBoolean(entity?.values.find(v => v.property.id === CLAIM_IS_FACTUAL_PROPERTY_ID)?.value);
```
b-d055: Small thing, but this compares the property id directly while the type check just above normalizes both sides. If `property.id` ever comes through with dashes (I think sync-store ids sometimes do), `isFactualClaim` would quietly be false and the claim would render as thumbs instead of chevrons, even though it's still correctly detected as a claim. Probably worth normalizing here too for consistency, something like `normalizeId(v.property.id) === normalizeId(CLAIM_IS_FACTUAL_PROPERTY_ID)`.

apps/web/core/io/queries.ts
```diff
@@ -615,6 +620,39 @@ export function getDebateVoteEntities(debateEntityId: string, signal?: AbortCont
...
+        debateTypeId: DEBATE_TYPE_ID,
+        first: BACKLINKS_PAGE_SIZE,
+        offset,
+      },
+      signal,
+    })
+  );
+}
+
+export function getClaimDebateCount(claimEntityId: string, signal?: AbortController['signal']) {
```
b-d055: Minor, but this collects every debate backlink id (1000 per page) just to return the length. It's gated to the icon badge so it's fairly scoped, but a claim with a lot of debates would page through all of them to show a count. If the indexer has a count/aggregate query that'd be nicer, otherwise maybe cap it at something like "99+" so we can stop early. Fine to leave for now if there's no easy count query.

apps/web/partials/entity-page/entity-vote-buttons.tsx
```diff
@@ -30,18 +37,45 @@ type OptimisticVote = 0 | 1 | 'none' | null;
...
+}: EntityVoteButtonsProps) {
   const { upvote, downvote, unvote, isConnected, personalSpaceId } = useEntityVote({
     entityId,
     spaceId,
   });
   const { smartAccount } = useSmartAccount();
   const { isPending: isAccountSetupPending } = usePendingPersonalSpace();
+
+  // Claim entities render a different vote control.
+  const { entity } = useQueryEntity({ id: entityId, spaceId });
```
b-d055: `useQueryEntity` also gives back `isLoading` and we're dropping it here. While the entity is loading, `entity` is null so `variant` falls back to the default arrows and then swaps to thumbs/chevrons once it resolves, which shows up as a little icon flash on claim rows. Could use `isLoading` to hold off on rendering the icon (or skeleton it) similar to how the ranking rows already handle it. Low priority.

apps/web/partials/entity-page/entity-vote-buttons.tsx
```diff
@@ -30,18 +37,45 @@ type OptimisticVote = 0 | 1 | 'none' | null;
...
+  const isClaim = entity?.types.some(t => normalizeId(t.id) === CLAIM_TYPE) ?? false;
```
b-d055: This is basically the same logic as `useIsClaimEntity`, which `EntityInteractionBar` already uses. Since we also need `isFactualClaim` off the same entity here, maybe the hook could return the entity (or an `isFactualClaim` flag) so both spots share one subscription. Or just reuse the hook for `isClaim` and keep a small lookup for the factual bit. Not important, just noticed the duplication.

b-d055 (changes requested):
@webmagic123  see comments

apps/web/partials/entity-page/entity-vote-buttons.tsx
webmagic123: @b-d055 
Updated

apps/web/partials/entity-page/entity-vote-buttons.tsx
webmagic123: @b-d055 
Updated

b-d055 (approved):
lgtm

PR #2004, opened 2026-07-16
feat: update default explore card layout with compact Join and left t…

webmagic123:
…humbnail


apps/web/core/utils/relation-video.ts
```diff
@@ -0,0 +1,10 @@
+import type { Relation } from '~/core/types';
+
+export const isPlayableVideoUrl = (value: string) => value.startsWith('ipfs://') || value.startsWith('http');
+
+export function getRelationVideoUrls(relations: Relation[], propertyId: string): string[] {
+  return relations
+    .filter(relation => relation.type.id === propertyId)
```
b-d055: This compares relation type IDs raw, but the caller in `fetch-explore-feed.ts` normalizes IDs everywhere else, that's why `normId` exists in that file:

- `types` filter: `normId(r.type.id) === typesRelationIdNorm`
- `textValueForProperty`: `normId(v.property.id)` / `normId(v.spaceId)`
- `imageFromRelationMedia`: `normId(r.type.id) === coverT`

`imageFromRelationMedia` is the direct analogue (it also reads a URL off `relation.toEntity.value`) and it normalizes *and* falls back to scanning all relations when the space-scoped pool comes back empty. This does neither.

If the GraphQL source returns UUID-formatted relation type IDs, `recordingUrls` / `debateVideoUrls` are silently `[]` on the server  which makes the `serverRecordingUrls` fallback in `useRecordingSources` dead, and video only appears once the sync store hydrates client-side.

Could you confirm against a real payload? If normalization genuinely isn't needed here, that's an argument for simplifying the rest of `fetch-explore-feed.ts`, not for the two styles coexisting.

apps/web/partials/explore/explore-feed-card.tsx
```diff
@@ -70,61 +242,25 @@ export function ExploreFeedCard({ item, hideSpaceLink = false, hideJoinButton =
...
-          <ExploreCommentsIcon className="text-grey-04" />
-          <span className="text-[14px] font-normal tabular-nums">{item.commentCount}</span>
-        </Link>
-      </div>
+      {isCommunityCall ? (
+        <CommunityCallCardBody item={item} actions={cardActions} />
+      ) : isDebate ? (
+        <DebateCardBody item={item} actions={cardActions} />
+      ) : isRanking ? (
+        <RankingCardBody item={item} />
```
b-d055: `CardBodyProps` above still says the actions row is "owned by the shell so every body renders it identically", which is no longer true now that ranking opts out. Worth a one-line update so the next reader doesn't file this as a bug like I nearly did.

apps/web/partials/explore/explore-ranking-card-body.tsx
```diff
@@ -0,0 +1,267 @@
...
+}
+
+function useRankingBlockPlacement(blockEntityId: string, spaceId: string) {
+  return useQuery({
+    queryKey: ['explore-ranking-block-placement', blockEntityId, spaceId],
+    enabled: Boolean(blockEntityId && spaceId),
+    staleTime: 60_000,
+    queryFn: async () => {
+      const relations = (await Effect.runPromise(
+        getRelationsByToEntityIds([blockEntityId], SystemIds.BLOCKS, spaceId)
```
b-d055: This is the third copy of this logic, `resolveBlockPlacement` in `core/io/subgraph/fetch-featured-rankings.ts` (~L114) is nearly line-for-line identical, down to the local `ToEntityRelation` type above, the `as unknown as` cast, and the `relations.find(...) ?? relations[0]` fallback. That file's own comment already notes it "mirrors resolve-ranking-share's placement resolution", so this is copy 3 of something flagged as duplicated.

Worth extracting a single `resolveBlockPlacement(blockEntityId, spaceId)` helper. It also gives one place to fix the `as unknown as` double cast, which currently defeats typechecking on an undecoded GraphQL shape.

Smaller thing: `spaceId` is passed as a query variable *and* re-filtered client-side, and then `?? relations[0]` discards the filter anyway.

apps/web/partials/explore/explore-ranking-card-body.tsx
```diff
@@ -0,0 +1,267 @@
...
+          )}
+          {showPagination ? (
+            <div className="ml-auto self-end [&>div:first-child]:hidden [&>div:last-child]:!mt-0 [&>div:last-child]:!mb-0 [&>div:last-child]:!justify-end">
+              <RankingBlockGlobalPagination
+                pageNumber={safePage}
+                hasPreviousPage={safePage > 0}
+                hasNextPage={safePage < totalPages - 1}
+                onSetPage={next => {
+                  if (next === 'previous') setPageNumber(p => Math.max(0, p - 1));
+                  else if (next === 'next') setPageNumber(p => Math.min(totalPages - 1, p + 1));
```
b-d055: These updaters read the raw `pageNumber`, but everything rendered uses the clamped `safePage`.

If `globalRankingEntityIds` shrinks while the user is on a later page (sync store update, entries unpublished), `safePage` clamps down but `pageNumber` stays high. "Previous" then decrements the stale value with no visible change — e.g. `pageNumber: 5`, `totalPages: 2` gives three dead clicks before the UI moves.

Seeding from `safePage` instead of `p` fixes it, or reset `pageNumber` when the ID list changes identity.

apps/web/core/explore/explore-constants.ts
```diff
@@ -1,3 +1,5 @@
+import { RANKING_BLOCK_TYPE_ID } from '~/core/ranking-block-ids';
```
b-d055: `RANKING_BLOCK_TYPE_ID` is imported but never used in this file — `eslint` warns (`@typescript-eslint/no-unused-vars`). Leftover from an earlier revision?

b-d055 (changes requested):
A few things I'd like to resolve before merge, inline comments below. The one I care most about is the raw ID comparison for recordings/debate videos, which likely makes the server-side fallback dead.

apps/web/core/utils/relation-video.ts
b-d055: Thanks @webmagic123  the other four are all resolved, and the shared `resolveBlockPlacement` came out clean.

This one went the other way. `f41f0dd34` didn't normalize the recording/debate lookups; it removed `normId()` from the left side of *every* comparison in `fetch-explore-feed.ts`, title/description values, cover and avatar images, and type chips, not just recordings.

That's a fine resolution if the decoder always emits hyphenless lowercase IDs. But it's now load-bearing for the whole card rather than one field: if that assumption is ever wrong, cards render untitled, imageless and chip-less, silently.

Two things before I sign off:

1. Did you confirm the ID format against a real payload? That was the original question, and it now covers a lot more surface than when I asked it.

2. The right-hand sides are still normalized — `pid`, `sid`, `coverT`, `avatarT`, `displaySpaceIdNorm`, `typesRelationIdNorm` are all still built by `normId()`, and the names still say "Norm". So each comparison normalizes one side only. Please either drop `normId()` from both sides and rename accordingly, or put it back on both. As written it reads like a normalized comparison but isn't, which is how this quietly breaks on someone later.

webmagic123: @b-d055 
Confirmed against a live payload. I queried the testnet API. Every id f41f0dd3 made load-bearing comes back hyphenless and lowercase.
 Six comparisons restored: textValueForProperty (property + space), imageFromRelationMedia (pool, cover, avatar), relationsInDisplaySpace, and the types filter.

b-d055 (changes requested):
needs rebase + conflict resolutions

b-d055 (approved):

PR #2017, opened 2026-07-17
fix: restore readable entity content widths

jwalkingjew:
## Summary

- restore the default entity/content reading column to 900px
- add an explicit 1142px `with-sidebar` layout that falls back to 900px when the rail hides at the 1023px breakpoint
- keep covers at 1192px and align avatars and community-call recordings with the normal content column
- opt the space-home community-calls rail into the wide layout only when call series exist
- add focused regression tests and the implementation plan

## Why

The community-calls work coupled the shared `EntityPageContentContainer` width to cover geometry, widening it from the historical 880px to 1144px so one custom right rail would fit. Because that container is shared, articles, posts, profiles, space headers, governance, claims, activity, imports, and side-panel entity content all inherited an overly wide reading measure.

This change makes readable content the default again and requires sidebar pages to opt into their wider shell.

## User impact

Content-heavy pages return to a readable line length. Space homes with community calls still have room for their right rail, while spaces without calls do not reserve an empty sidebar or change width after rendering.

## Validation

- `bun run test partials/entity-page/entity-page-content-container.test.tsx` — 4 passed
- `bunx tsc --noEmit` from `apps/web` — passed
- targeted ESLint on changed files — passed
- `bun run lint` — passed with existing repository warnings
- `bunx prettier --check ...` on changed files — passed
- full web Vitest run — 1,355 passed; 7 unrelated analytics/feature-flag tests failed because jsdom received an invalid localStorage backing path in this environment
- root test command remains blocked by `packages/auth` exiting when no test files are found
- production build compiled `@geogenesis/auth`, then stalled in Next/Turbopack after inferring the workspace root above the repo from an outer `package-lock.json`; it was interrupted after 4m48s with no application error

## Review focus

Please compare the proposed 900px normal content width and 1142px sidebar shell in the deploy preview, especially navigation between a long article/post and a space home with community calls.


apps/web/app/space/[id]/(space)/layout.tsx
```diff
@@ -47,9 +48,10 @@ export default async function Layout(props0: LayoutProps) {
     notFound();
   }
 
-  const props = await getSpaceFrontPage(spaceId);
+  const [props, communityCalls] = await Promise.all([getSpaceFrontPage(spaceId), fetchCommunityCalls(spaceId)]);
 
   const typeIds = props.space?.entity?.types?.map(t => t.id) ?? [];
+  const contentVariant = communityCalls.length > 0 ? 'with-sidebar' : 'content';
```
Comment (b-d055) [inline apps/web/app/space/[id]/(space)/layout.tsx]: This layout wraps all six space routes, so on a space with community calls the header and tabs render at 1142px on every tab, but only the home page's body opts into that width. Activity, claims, community, governance, debates, and `TopicEntityBody` all stay at the 900px default, so the header ends up noticeably wider than the content under it on those pages.

Might be worth scoping the variant decision to the home route instead of the shared layout?

Example, at a sceen size above ~1400px look at 
https://geogenesis-git-codex-content-width-layout-plan-geo-browser.vercel.app/space/4582fbbee28a16589154f7e36f1ee3c5
vs
https://geogenesis-git-codex-content-width-layout-plan-geo-browser.vercel.app/space/c9f267dcb0d270718c2a3c45a64afd32
notice the width difference?

apps/web/app/space/[id]/(space)/layout.tsx
```diff
@@ -47,9 +48,10 @@ export default async function Layout(props0: LayoutProps) {
     notFound();
   }
 
-  const props = await getSpaceFrontPage(spaceId);
+  const [props, communityCalls] = await Promise.all([getSpaceFrontPage(spaceId), fetchCommunityCalls(spaceId)]);
```
Comment (b-d055) [inline apps/web/app/space/[id]/(space)/layout.tsx]: Moving this fetch out of the home page (where it was inside `<Suspense>`) and into the layout means a failure here now takes down every space route instead of just the home page. There's no `error.tsx` under `app/space/`, so it goes all the way up to the root error boundary.

Adding `.catch(() => [])` here and at `page.tsx:80` would degrade to "no rail" instead, which is how `fetchCommunityCallsForExplore` already handles it.

b-d055 (commented):
@jwalkingjew  see comments (you also need to resolve conflicts)

apps/web/app/space/[id]/(space)/layout.tsx
Comment (jwalkingjew) [reply apps/web/app/space/[id]/(space)/layout.tsx]: Addressed in 8512cc2a. The shared header now uses the with-sidebar width only on the exact space home route and only when that home page will actually render the community-call rail. Nested space routes retain the standard content width. I also added regression coverage for the home route, a nested route, no-sidebar behavior, and the debate chrome gate.

apps/web/app/space/[id]/(space)/layout.tsx
Comment (jwalkingjew) [reply apps/web/app/space/[id]/(space)/layout.tsx]: Addressed in 8512cc2a. Both layout.tsx and page.tsx now catch fetchCommunityCalls failures and fall back to an empty list, so the route degrades to no rail instead of reaching the root error boundary. The focused tests and TypeScript check pass.

PR #2023, opened 2026-07-20
feat: show explore sidebar on root space home

webmagic123:


apps/web/core/explore/fetch-explore-side-panel-data.ts
```diff
@@ -0,0 +1,96 @@
...
+  communityCalls: ExploreCall[];
+};
+
+type FetchExploreSidePanelDataOptions = {
+  wallet?: string | null;
+  memberSpaceId?: string | null;
+  featuredSpacesPromise?: Promise<FeaturedSpace[]>;
+};
+
+export async function fetchExploreSidePanelData(
```
b-d055: This gets called once in `layout.tsx:59` and again in `page.tsx:88`. Both sit in the same route segment, so a request to the root space home runs the whole thing twice. Nothing dedupes it: this is a bare `export async function`, and every fetcher underneath it is uncached too.

Neither half can lean on Next's fetch memoization, for different reasons. The REST calls (the membership fan out, via `fetch-proposed-members.ts:32-38`) are GETs that Next would memoize, but `rest/client.ts:255,268-269` stamps a fresh `crypto.randomUUID()` into `x-request-id` on every call, which changes the key. The GraphQL calls are POSTs (`graphql-request` v7 defaults to POST, and `subgraph/graphql.ts:16-28` posts directly), and Next only memoizes GET. So stripping the UUID headers would not buy anything, the fix has to be at this level.

On cost: the fan out is capped at `MAX_FEATURED = 60` (`fetch-featured-spaces.ts:32`), so a signed in visitor with a resolvable member space makes up to 120 REST calls instead of 60. But it is not only the fan out. `fetchFeaturedSpaces` runs a sequential BFS (up to 2500 nodes at 200 per round trip) and `fetchCommunityCallsForExplore` does a 200 result search plus a batch hydrate. Those double for every visitor, anonymous included.

The correctness side is what I would actually worry about. Both call sites `.catch(() => null)` independently, and each derives the container variant from its own result: `layout.tsx:63-67` feeds `SpaceHeaderContentContainer`, `page.tsx:110` feeds `EntityPageSidebarLayout`. If one succeeds and the other fails you get a 900px header over a 1142px body, or the reverse. Only visible above 1023px, since `lg:` is max-width here, but that is most desktop traffic.

`cache()` fixes all of it, and there is precedent right next door: `fetchCommunityCalls` is already wrapped at `fetch-community-calls.ts:19`. Both call sites pass zero args so they share a key:

```ts
import { cache } from 'react';

export const fetchExploreSidePanelData = cache(
  async (options: FetchExploreSidePanelDataOptions = {}): Promise<ExploreSidePanelData> => {
    // ...
  }
);
```

apps/web/app/space/[id]/(space)/layout.tsx
```diff
@@ -49,13 +52,19 @@ export default async function Layout(props0: LayoutProps) {
     notFound();
   }
 
-  const [props, communityCalls] = await Promise.all([
+  const isRootSpace = spaceId === ROOT_SPACE;
+  const [props, communityCalls, exploreSidePanelData] = await Promise.all([
     getSpaceFrontPage(spaceId),
-    fetchCommunityCalls(spaceId).catch(() => []),
+    isRootSpace ? Promise.resolve([]) : fetchCommunityCalls(spaceId).catch(() => []),
+    isRootSpace ? fetchExploreSidePanelData().catch(() => null) : Promise.resolve(null),
```
Comment (b-d055) [inline apps/web/app/space/[id]/(space)/layout.tsx]: Worth flagging that this also stopped streaming. `ae0600212` originally had the panel behind `<React.Suspense fallback={null}>` via a `RootExploreSidePanelContainer`. Upstream `41f8dd082` (#2017) removed that streaming container pattern in favour of an eager `Promise.all` for `communityCalls`, and the merge in `aa58f4fb7` re-applied the explore panel in the new eager style while adding this second call site.

So it reads like a merge resolution side effect rather than a deliberate change. There is no `loading.tsx` under `app/space/[id]/` either, so Next does not insert an implicit boundary. Net effect is the root space shell now blocks on the full fan out before anything renders, where it used to stream.

If putting the Suspense back is awkward, the `cache()` change at least halves the wait.

apps/web/app/space/[id]/(space)/space-chrome-gate.tsx
```diff
@@ -25,7 +27,9 @@ type SpaceHeaderContentContainerProps = {
 
 export function SpaceHeaderContentContainer({ children, spaceId, hasSidebar }: SpaceHeaderContentContainerProps) {
   const pathname = usePathname();
-  const isSpaceHome = pathname === `/space/${spaceId}` || pathname === `/space/${spaceId}/`;
+
```
Comment (b-d055) [inline apps/web/app/space/[id]/(space)/space-chrome-gate.tsx]: Nit: trailing whitespace, `prettier --check` flags it. Not a CI failure, CI only runs Build, Test and eslint, and the `format` script is `--write` rather than `--check`. Just noting it since it is the only formatting deviation in the branch.

```suggestion

```

apps/web/app/space/[id]/(space)/space-chrome-gate.tsx
```diff
@@ -25,7 +27,9 @@ type SpaceHeaderContentContainerProps = {
 
 export function SpaceHeaderContentContainer({ children, spaceId, hasSidebar }: SpaceHeaderContentContainerProps) {
   const pathname = usePathname();
-  const isSpaceHome = pathname === `/space/${spaceId}` || pathname === `/space/${spaceId}/`;
+  
+  const homePath = NavUtils.toSpace(spaceId);
```
Comment (b-d055) [inline apps/web/app/space/[id]/(space)/space-chrome-gate.tsx]: `NavUtils.toSpace` returns `/root` for the root space (`core/utils/utils.ts:26`), which is what makes `/root` work here. But `/space/a19c345ab9866679b001d7d2138d88a1` is still served by the `[id]` route, and I could not find anything that redirects it: no `/space/*` rule in `next.config.ts`, no `middleware.ts`, and `SpaceRedirect` lives in the entity route group rather than `(space)`. Both `page.tsx:84` and `layout.tsx:55` branch on `spaceId === ROOT_SPACE`, so that route explicitly supports rendering the root space.

On that URL `isSpaceHome` is false, so the header renders at 900px over a 1142px body. That is a 242px mismatch on any viewport above 1023px.

The underlying thing is that the header asks "am I on home" from the pathname while `entity-page-sidebar-layout.tsx:14` derives its variant from sidebar presence. Two sources of truth for the same question, so they can disagree. Threading the state down would be the real fix. If you want the cheap version, this passes the existing suite unchanged:

```ts
const isSpaceHome = [homePath, `/space/${spaceId}`].some(p => pathname === p || pathname === `${p}/`);
```

apps/web/core/explore/explore-side-panel-has-content.ts
```diff
@@ -0,0 +1,18 @@
+import type { ExploreSidePanelData } from '~/core/explore/fetch-explore-side-panel-data';
+import { normId } from '~/core/utils/norm-id';
+
+/**
+ * Server-side approximation of whether {@link ExploreSidePanel} will render.
+ * Omits curator-onboarding (client-only); used to avoid reserving sidebar width
+ * when the panel would otherwise be empty.
+ */
+export function exploreSidePanelHasServerContent(data: ExploreSidePanelData): boolean {
```
b-d055: The comment above is accurate about omitting curator onboarding, but both callers use this to decide whether to *mount* `ExploreSidePanel` at all, not just whether to reserve width. That makes the `showOnboarding` branch in `explore-side-panel.tsx:51-53` unreachable whenever this returns false.

Worth noting `useCuratorOnboardingStatus` never "completes": `isVisible` is just `Boolean(personalSpaceId) && isPersonalSpaceFetched` (`use-curator-onboarding-status.ts:95`), so the section is meant to render for every signed in user indefinitely.

The realistic trigger is failure rather than membership. `fetch-explore-side-panel-data.ts:47-49` wraps all three fetches in `.catch(() => [])`, and `page.tsx:88` wraps the whole call in `.catch(() => null)`. An indexer hiccup makes this return false and silently drops the panel, onboarding included.

Cheapest fix is to mount `ExploreSidePanel` unconditionally and let its own `hasContent` decide, using this helper only to pick the container variant.

apps/web/partials/explore/explore-page.tsx
```diff
@@ -28,9 +30,31 @@ export function ExplorePage({
   editorSpaceIds,
   communityCalls,
 }: Props) {
+  const hasSidebar = exploreSidePanelHasServerContent({
```
b-d055: Same root cause as the helper, from the other direction. `explore-side-panel.tsx:48` also subtracts `dynamicPendingSet` from `joinableSpaces`, which this helper cannot know about, so the server can say "has content" while the client renders `null`.

The most reachable path is not the optimistic atom, it is the cookie: `pendingMembershipSpaceIds` is only computed when `memberSpaceId` resolves from `WALLET_ADDRESS` (`fetch-explore-side-panel-data.ts:32-44, 69-86`). If that cookie is missing or stale while the client is signed in, the server sees nobody as a member, says has content, and the client then filters everything out.

When that happens the container is stuck wide: `entity-page-sidebar-layout.tsx:11` tests `sidebar !== null` on the *element*, which stays truthy even when the component returns `null`, so it can never react.

apps/web/partials/explore/explore-page.tsx
```diff
@@ -28,9 +30,31 @@ export function ExplorePage({
...
+            pendingMembershipSpaceIds={pendingMembershipSpaceIds}
+            memberOrEditorSpaceIds={memberOrEditorSpaceIds}
+            editorSpaceIds={editorSpaceIds}
+            communityCalls={communityCalls}
+          />
+        ) : null
+      }
+    >
+      <main className="min-w-0 pt-5">
         <div className="mx-auto w-full max-w-[880px]">
```
b-d055, line 43: Withdrawing this one, I got it wrong. `max-w-[880px]` is not dead. It binds whenever `hasSidebar` is false (container drops to 900px), whenever `ExploreSidePanel` returns `null` at runtime while the element stays non-null (main column becomes the full 1142px), and at 1023px and below where the rail is `lg:hidden` and the container falls back to 900px. It is inert only in the desktop case where the sidebar is present and rendering. Sorry for the noise.

apps/web/partials/explore/explore-page.tsx
```diff
@@ -45,15 +69,6 @@ export function ExplorePage({
           feedTopSpacingClassName=""
         />
       </main>
```
b-d055, line 56: The `<div aria-hidden className="w-px shrink-0 self-stretch bg-divider lg:hidden" />` came out here and nothing replaced it. `EntityPageSidebarLayout` does not render one, so `/explore` and `/root` lose their vertical divider while every other space page keeps one. The `ml-8` added to the aside looks like it was meant to stand in for the spacing. Was dropping the divider intentional?

apps/web/partials/explore/explore-side-panel.tsx
```diff
@@ -86,7 +86,7 @@ export function ExploreSidePanel({
     // the offset, the panel's bottom sits below the viewport at top-of-page.
     // `self-start` stops the flex container from stretching the aside past its
     // declared height.
-    <aside className="sticky top-11 flex h-[calc(100dvh-2.75rem)] w-[360px] shrink-0 flex-col self-start lg:hidden">
+    <aside className="sticky top-11 ml-8 flex h-[calc(100dvh-2.75rem)] w-[360px] shrink-0 flex-col self-start lg:hidden">
```
b-d055: The commit says "share rail width" but these two still differ:

- here: `ml-8 w-[360px]`, no border, no `pl-8`
- `space-community-calls-section.tsx:65`: `ml-8 w-[300px] border-l border-divider pl-8` (unchanged by this PR)

Inside the 1142px container that is a 750px main column on root and explore against 810px elsewhere. The rail content gap is wider than that: the 300px rail is border-box, so its usable width is `300 - 32 - 1 = 267px` against 360px here, a 93px difference.

If 360 with no divider is the target, `SpaceCommunityCallsSection` should follow so the two rails actually match.

apps/web/styles/styles.css
```diff
@@ -75,6 +75,9 @@
   /* Breakpoints (max-width variants defined below) */
   --breakpoint-*: initial;
 
+  /* Match CONTENT_MAX_WIDTH */
+  --width-content-max: 1144px;
```
b-d055: This variable is added by this PR but nothing consumes it. Grepping the tree for `content-max` (excluding `.next`) turns up only this definition, no `w-content-max` or `max-w-content-max` usage anywhere.

The comment is stale too. There is no symbol named `CONTENT_MAX_WIDTH` any more, only `ENTITY_PAGE_CONTENT_MAX_WIDTH = 900`, `ENTITY_PAGE_WITH_SIDEBAR_MAX_WIDTH = 1142` and `ENTITY_PAGE_COVER_MAX_WIDTH = 1192` (`entity-page-layout.ts:1-3`). 1144 matches none of them. Looks like leftovers from an earlier iteration, worth dropping.

apps/web/core/explore/fetch-explore-side-panel-data.ts
```diff
@@ -0,0 +1,96 @@
...
+  options: FetchExploreSidePanelDataOptions = {}
+): Promise<ExploreSidePanelData> {
+  const wallet =
+    options.wallet !== undefined ? options.wallet : ((await cookies()).get(WALLET_ADDRESS)?.value ?? null);
```
b-d055: Minor, carried over from the last round. `explore/page.tsx` passes `string | undefined` from `cookies().get()?.value`, so on a signed out visit this falls through and re-reads cookies to land on the same `null`. Two explicit call shapes would read better than distinguishing absent from explicitly null.

apps/web/app/space/[id]/(space)/page.tsx
```diff
@@ -77,17 +81,33 @@ export default async function SpacePage(props0: Props) {
     return <TopicEntityBody spaceId={spaceId} topicEntityId={space.topicId} />;
   }
 
-  const [props, communityCalls] = await Promise.all([
+  const isRootSpace = spaceId === ROOT_SPACE;
```
Comment (b-d055) [inline apps/web/app/space/[id]/(space)/page.tsx]: The `isRootSpace` and `hasSidebar` logic is duplicated close to verbatim between here and `layout.tsx:55-67`. Once `fetchExploreSidePanelData` is memoized, pulling this into a small shared helper would be easy and would keep the two from drifting.

apps/web/app/space/[id]/(space)/space-chrome-gate.test.tsx
```diff
@@ -29,6 +31,18 @@ describe('space chrome layout', () => {
     expect(screen.getByText('Header').dataset.entityPageContentVariant).toBe('with-sidebar');
   });
 
+  it('aligns the header with the sidebar layout on the root space home', () => {
```
Comment (b-d055) [inline apps/web/app/space/[id]/(space)/space-chrome-gate.test.tsx]: Nice to see the `/root` case pinned. One gap: nothing covers the null sidebar path in `EntityPageSidebarLayout`, which is what fixes the empty rail reserving width. That is the load bearing behavior now, so it would be good to lock in.

b-d055 (commented):
Re-reviewed after the rework onto `EntityPageSidebarLayout`. Moving off the custom `OverviewSideRail` abstraction cleaned up most of what I flagged last round: the empty rail no longer reserves width, the sticky rail works now that the aside is a direct flex child, and the `-mt-8` coupling to `Main` is gone.

One thing I think should block: `fetchExploreSidePanelData` runs twice per root space load with nothing deduping it, and the two copies can disagree in a way that misaligns the header by 242px. It looks like it came in as a merge resolution side effect rather than on purpose.

Everything else is a suggestion. Nice to see the `/root` header case pinned in a test.

apps/web/app/space/[id]/(space)/layout.tsx
Comment (webmagic123) [reply apps/web/app/space/[id]/(space)/layout.tsx]: @b-d055 
Updated.
Added stream root space explore panel.

apps/web/core/explore/explore-side-panel-has-content.ts
webmagic123: @b-d055 
Updated

apps/web/partials/explore/explore-page.tsx
webmagic123: @b-d055 
Updated.

apps/web/partials/explore/explore-side-panel.tsx
webmagic123: @b-d055 
Updated.

apps/web/core/explore/fetch-explore-side-panel-data.ts
webmagic123: @b-d055 
Updated.

apps/web/app/space/[id]/(space)/page.tsx
Comment (webmagic123) [reply apps/web/app/space/[id]/(space)/page.tsx]: @b-d055 
Updated.

apps/web/app/space/[id]/(space)/space-chrome-gate.test.tsx
Comment (webmagic123) [reply apps/web/app/space/[id]/(space)/space-chrome-gate.test.tsx]: @b-d055 
Added entity-page-sidebar-layout.test.tsx

apps/web/partials/entity-page/entity-page-content-container.tsx
```diff
@@ -25,9 +25,11 @@ export function EntityPageContentContainer({ children, variant = 'content' }: Pr
     <div
       className={cx(
         'mx-auto w-full',
-        variant === 'with-sidebar'
-          ? 'max-w-[var(--entity-page-with-sidebar-max-width)] lg:max-w-[var(--entity-page-content-max-width)]'
-          : 'max-w-[var(--entity-page-content-max-width)]'
+        variant === 'with-sidebar' &&
+          'max-w-[var(--entity-page-with-sidebar-max-width)] lg:max-w-[var(--entity-page-content-max-width)]',
+        variant === 'auto-sidebar' &&
+          'max-lg:[&:has(>div>aside)]:max-w-[var(--entity-page-with-sidebar-max-width)] max-w-[var(--entity-page-content-max-width)]',
```
b-d055: `max-lg` doesn't exist in this repo, so this class never compiles and the widening never happens.

`styles.css:76` sets `--breakpoint-*: initial`, which wipes the breakpoint theme keys that `max-*` variants are built from. `lg` only exists as a `@custom-variant` at line 247, and `max-*` doesn't pick those up. I compiled `styles.css` through `@tailwindcss/postcss` with the real content scan and there's no `max-lg` rule and no `:has()` rule in the output at all.

The only token here that survives is the 900px cap, which makes `auto-sidebar` identical to `content`. On `/explore` the rail still takes its 392px (`ml-8` + `w-[360px]` + `shrink-0`), so the feed column lands at 508px instead of the 750px you'd get at 1142px. It's ~847px on master today. `/root` is unaffected since it uses the default `reserve`.

`not-lg:` does work with the custom variant and compiles to `@media not (max-width: 1023px)`:

```suggestion
          'not-lg:[&:has(>div>aside)]:max-w-[var(--entity-page-with-sidebar-max-width)] max-w-[var(--entity-page-content-max-width)]',
```

One thing worth checking either way: was Explore meant to get the wider sidebar layout width here? If 900px is actually the design, then `auto-sidebar` isn't doing anything and it'd be clearer to drop the variant and pass `variant="content"`.

b-d055 (commented):
@webmagic123  one new ask before I merge

apps/web/partials/entity-page/entity-page-content-container.tsx
webmagic123: @b-d055 
Updated.
I've removed the broken auto-sidebar variant (and the unused sidebarWidth prop) so Explore and space home both resolve through the same with-sidebar path — 1142px container, 360px rail, 750px feed.

webmagic123:
@b-d055 
- Feature: The root space home reuses the Explore page's sidebar on its Overview tab, just like other spaces have a rail on their overview
- Solutions:
  1. Fixed: the rail was leaking onto the non-overview tabs. /root reuses the shared space overview page through a thin wrapper, but the wrapper dropped searchParams —  ?tabId=… was never seen and the rail rendered on the content tabs (Education, Ontology, About) too.
  Forwarded searchParams through the wrapper, so the rail stays on Overview + Community only (Community mirrors Overview by design).
  2. Made the header width follow the actual rail. The header width was pinned to a hardcoded Overview/Community route check, so it couldn't respond to the rail on any other tab. I changed it so the rail's content (via the atom) drives the header width on whichever tab
  is showing, using the route-gated server signal only as the SSR seed (no flash, non-rail tabs stay narrow).
  3. Updated the chrome-gate tests to match.

b-d055 (approved):

PR #2029, opened 2026-07-21
Feat/set specific image in gallery or list view

webmagic123:


apps/web/core/hooks/use-block-main-media-property.ts
```diff
@@ -0,0 +1,43 @@
...
+} from './use-block-media-dimensions';
+
+export type BlockMainMedia = MainMediaProperty & {
+  dimensions: BlockMediaDimensions;
+};
+
+/**
+ * Block-level main media for list/gallery views
+ */
+export function useBlockMainMedia(
```
b-d055: Naming nit, but it'll cost someone real time later: this file (`use-block-main-media-property.ts`) exports `useBlockMainMedia`, while `use-block-main-media.ts` exports `useBlockMainMediaUrl`. The file named after the hook doesn't contain it.

Suggest `use-block-main-media.ts` → `use-block-main-media-url.ts`, and this file → `use-block-main-media.ts`. Pure rename, no behavior change.

Same vein: `blockMainMediaDimensions()` below is a one-line `?? NO_BLOCK_MEDIA_DIMENSIONS` wrapper around a constant that lives in `use-block-media-dimensions.ts` — probably belongs next to the constant, or just inlined at the two call sites.

apps/web/core/hooks/use-block-main-media.ts
```diff
@@ -0,0 +1,153 @@
...
+import { useEntityMedia } from '~/core/utils/use-entity-media';
+
+function isDirectMediaUrl(value: string | null | undefined): value is string {
+  return Boolean(value && (value.startsWith('ipfs://') || value.startsWith('http://') || value.startsWith('https://')));
+}
+
+/**
+ * Resolves the gallery/list main image URL for an entity.
+ */
+export function useBlockMainMediaUrl({
```
b-d055: This hook carries five concerns in ~150 lines: selected-property relation lookup (`:38`), legacy cover/avatar fallback (`:44`), keyframe store lookup (`:52`), keyframe network fetch with its own `useState` + effect + `cancelled` flag + `videoEntityId === selectedEntityId` staleness guard (`:60-99`), and entity-id → URL resolution (`:142-152`).

The keyframe half is self-contained and already has its own cache key — pulling it out as `useVideoKeyframeUrl(videoEntityId, spaceId)` would leave this as a readable three-branch selector.

While it's being moved: `cache.fetchQuery` inside an effect with local state is `useQuery` in disguise. Switching to `useQuery` deletes the state, the effect, the `cancelled` flag, and the staleness guard. It mirrors the existing pattern in `use-entity-media.ts` so it's consistent rather than novel — but no reason to carry it into a new hook.

b-d055 (commented):
@webmagic123  see comments

apps/web/core/hooks/use-block-main-media.ts
webmagic123: @b-d055 
Updated

apps/web/core/hooks/use-block-main-media-property.ts
webmagic123: @b-d055 
Updated

PR #2030, opened 2026-07-21
feat: add Daily activities checklist to space overview side panel

webmagic123:


apps/web/partials/space-page/add-data-panel.tsx
```diff
@@ -96,6 +98,7 @@ export function AddDataPanel({ spaceId }: Props) {
 
     setSeed({ mode: 'ingestion', url: normalizedUrl });
     setChatOpen(true);
+    completeDailyUploadActivity(spaceId);
```
b-d055: This fires the completion on the fallback/error path too, so a failed upload still checks off "Upload a news story." If the request throws, 500s, or comes back without a jobId, we shouldn't be marking the daily activity done. Can we move this so it only runs when the ingestion actually succeeded?

apps/web/partials/space-page/add-data-panel.tsx
```diff
@@ -81,6 +82,7 @@ export function AddDataPanel({ spaceId }: Props) {
           if (body.jobId) {
             setSeed({ mode: 'inject', url: normalizedUrl, jobId: body.jobId, injectType: classification.type });
             setChatOpen(true);
+            completeDailyUploadActivity(spaceId);
```
b-d055: Minor follow-on to the above: this one fires at the 202/accept, before ingestion actually completes. Defensible as "job queued," but worth confirming that's the signal we want to count, vs waiting for the job to finish.

apps/web/core/space/use-space-daily-activities.ts
```diff
@@ -0,0 +1,301 @@
...
+        r =>
+          !r.isDeleted &&
+          ID.equals(r.type.id, DATA_BLOCK_TOOLS_PROPERTY_ID) &&
+          ID.equals(r.toEntity.id, LINK_INGESTION_TOOL_ID)
+      );
+    });
+  }, [blocksRelationEntityIds, initialBlockEntities, toolRelations.length]);
+
+  const tasks = React.useMemo(() => {
+    if (!isSignedIn) return [];
```
b-d055, line 116: The task shows for any signed-in user, but the only UI that completes it (AddDataPanel) early-returns on !canEdit. So a signed-in non-editor gets a row they can never check and a progress bar stuck below 100%. Either gate the task on the same edit permission, or give non-editors a path to complete it.

apps/web/core/space/use-space-daily-activities.ts
```diff
@@ -0,0 +1,301 @@
...
+      const relationsForBlock = [
+        ...typeRelations.filter(r => ID.equals(r.fromEntity.id, blockId)),
+        ...(initial?.relations ?? []).filter(
+          r =>
+            ID.equals(r.fromEntity.id, blockId) &&
+            r.type.id === SystemIds.TYPES_PROPERTY &&
+            ID.equals(r.spaceId, spaceId) &&
+            !r.isDeleted
+        ),
+      ];
```
b-d055: This hand-rolls a concat of live + snapshot relations, and the snapshot copies still carry isDeleted: false. A relation the editor just deleted gets re-added from the snapshot, so a removed ranking block keeps counting as a task. We have mergeRelations for exactly this (merges by id + relationKey and respects deletes). Can we use that here instead of the manual concat?

apps/web/core/space/use-space-daily-activities.ts
```diff
@@ -0,0 +1,301 @@
...
+    () => (myRankEntity ? parseTimestampMs(myRankEntity.updatedAt) : 0),
+    [myRankEntity]
+  );
+
+  const isSubmissionLive = React.useMemo(() => {
+    if (!isRolling || !myRankEntity) return true;
+    const windowElapsed =
+      submissionFrequencyHours != null &&
+      submittedAtMs > 0 &&
+      Date.now() >= submittedAtMs + submissionFrequencyHours * MS_PER_HOUR;
```
b-d055: Date.now() is captured inside this memo, so the roll-off state freezes at first render and won't update until something else re-renders the hook. For a long-lived open tab the rolling-ranking status can go stale. Worth deriving this off a ticking value or recomputing on an interval.

apps/web/core/space/use-space-daily-activities.ts
```diff
@@ -0,0 +1,301 @@
...
+        setComplete(readDailyUploadComplete(spaceId));
+      }
+    };
+
+    window.addEventListener('storage', onStorage);
+    window.addEventListener('geo:daily-activities-upload', onCustom);
+
+    const timeoutId = window.setTimeout(() => {
+      setComplete(readDailyUploadComplete(spaceId));
+    }, msUntilNextLocalMidnight() + 50);
```
b-d055: Single timeout to next midnight means the reset only fires once. If the tab stays open past a second midnight it won't reset again. Consider re-arming the timeout after it fires, or keying off a date value.

apps/web/core/space/use-space-daily-activities.ts
```diff
@@ -0,0 +1,301 @@
...
+}
+
+export function useDailyUploadActivityComplete(spaceId: string): boolean {
+  const [complete, setComplete] = React.useState(false);
+
+  React.useEffect(() => {
+    setComplete(readDailyUploadComplete(spaceId));
+
+    const onStorage = (event: StorageEvent) => {
+      if (event.key?.startsWith(`geogenesis.daily-activities.upload.v1:${spaceId}:`)) {
```
b-d055: The storage key prefix is re-typed here as a literal. It's defined in daily-activities-storage.ts already, so this can drift if that prefix ever changes. Can we export and reuse the constant?

apps/web/app/space/[id]/(space)/page.tsx
```diff
@@ -238,9 +238,9 @@ const SubtopicGalleryContainer = async ({ spaceId }: SubtopicGalleryContainerPro
   return <SubtopicGallery spaceId={spaceId} subtopics={subtopics} />;
 };
 
-const SpaceCommunityCallsContainer = async ({ spaceId }: { spaceId: string }) => {
+const SpaceOverviewSidePanelContainer = async ({ spaceId }: { spaceId: string }) => {
   const series = await fetchCommunityCalls(spaceId);
```
Comment (b-d055) [inline apps/web/app/space/[id]/(space)/page.tsx]: The whole side panel (including Daily activities) awaits fetchCommunityCalls before anything renders, so community-calls latency blocks paint of the daily checklist. These feel independent. Could the daily section render without waiting on the calls fetch?

apps/web/partials/space-page/space-daily-activities-section.tsx
```diff
@@ -0,0 +1,170 @@
...
+          className="mt-1.5 inline-flex h-6 w-6 shrink-0 items-center justify-center rounded text-grey-04 transition-colors hover:text-text"
+        >
+          <span className={cx('transition-transform', expanded ? 'rotate-180' : 'rotate-0')}>
+            <ChevronDownSmall />
+          </span>
+        </button>
+      </div>
+
+      <div className="mt-4 flex flex-col gap-2">
+        <div className="h-1.5 w-full overflow-hidden rounded-full bg-grey-02">
```
b-d055: The progress bar has no role="progressbar" / aria-valuenow, so it's invisible to screen readers. Small add if a11y is in scope.

apps/web/partials/space-page/space-daily-activities-section.tsx
```diff
@@ -0,0 +1,170 @@
...
+          <div
+            className="h-full rounded-full transition-[width] duration-300"
+            style={{
+              width: `${progressPercent}%`,
+              backgroundColor: DAILY_ACTIVITIES_PROGRESS_COLOR,
```
b-d055: This sets the fill color with an inline hardcoded hex while the checkmark above uses the bg-purple token. Same color, two sources of truth. Prefer the token here too.

apps/web/partials/community-calls/space-community-calls-section.tsx
```diff
@@ -83,8 +91,12 @@ export function SpaceCommunityCallsSection({ spaceId, series }: { spaceId: strin
       ) : (
         <UpcomingCard row={highlight.row} isMember={isMember} isEditor={isEditor} accessLoading={accessLoading} />
       )}
-    </aside>
+    </>
   );
+
+  if (embedded) return <div>{content}</div>;
```
b-d055: The non-embedded branch below (the aside wrapper) looks like it's now dead once everything renders through embedded. Worth removing if nothing hits it anymore.

b-d055 (changes requested):
@webmagic123  A few things to look at before merge. The top cluster (completion firing on failure, task unachievable for non-editors, and the stale-delete leak in task discovery) are behavioral. The rest are should-fixes and nits.

apps/web/core/space/use-space-daily-activities.ts
webmagic123: @b-d055 
Updated.

apps/web/core/space/use-space-daily-activities.ts
webmagic123: @b-d055 
Updated.

apps/web/partials/space-page/space-daily-activities-section.tsx
webmagic123: @b-d055 
Updated.

apps/web/partials/community-calls/space-community-calls-section.tsx
webmagic123: @b-d055 
Fixed.

b-d055 (approved):
lgtm

PR #2039, opened 2026-07-21
fix: dedupe @radix-ui/react-dialog to stop search dialog crash

ohohoreilly:
## Summary
- A recent dependency bump left `cmdk` resolving its own separate physical copy of `@radix-ui/react-dialog` (`1.1.17`) instead of sharing the app's hoisted copy (`1.1.19`). Confirmed in `bun.lock` via the now-removed `cmdk/@radix-ui/react-dialog` entry.
- Two physical installs means two independent Dialog React contexts, so `Dialog.Title` (imported directly in `apps/web/partials/search/dialog.tsx`) couldn't find the context provided by `cmdk`'s internally-rendered `Dialog.Root`, throwing `` `DialogTitle` must be used within `Dialog` `` on every click of the search icon.
- This is [GEOGENESIS-A4](https://geo-7y.sentry.io/issues/7625286711/) in Sentry — escalating, 32+ events / 5+ users since 2026-07-21.
- Fix: pin `@radix-ui/react-dialog` in the root `overrides` block (same mechanism already used for the `viem` pin) so every consumer, including `cmdk`, resolves to a single shared instance.

## Test plan
- [x] Clean `rm -rf node_modules && bun install --frozen-lockfile` confirms only one `@radix-ui/react-dialog` package remains in the tree and `cmdk` no longer has a nested copy.
- [x] Built the actual production bundle (`next build`) with the fix and served it locally; clicked the search icon 6x in a row — dialog opened successfully every time with no `DialogTitle` errors (previously crashed on every click on production).

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -177,6 +177,7 @@
     },
   },
   "overrides": {
+    "@radix-ui/react-dialog": "1.1.19",
     "@tanstack/react-query": "^5.101.1",
     "permissionless": "^0.3.4",
     "prosemirror-model": "^1.25.9",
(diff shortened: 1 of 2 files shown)


ohohoreilly:
Closing as already fixed. The identical override landed in #2038 (`b5420cad6`, "fix: search crash"), which is on `master`:

```json
"@radix-ui/react-dialog": "1.1.19"
```

`bun.lock` has no nested `cmdk/@radix-ui/react-dialog` entry any more, so there's a single physical copy and one Dialog context. The two PRs were the same fix arrived at in parallel; #2038 won and this one was never closed.

Confirmed against Sentry: [GEOGENESIS-A4](https://geo-7y.sentry.io/issues/GEOGENESIS-A4) last fired **2026-07-23**, 192 events / 16 users, and has been silent for 34 days. Resolving it there too.

PR #2042, opened 2026-07-22
feat(inject): surface enrich outcome with target pill and review-panel preview

0xTaneja:
## Problem

The chat inject flow has two outcomes from the news pipeline: a **create** (new story entity) and an **enrich** (the URL matches a story that already exists on-chain, and the edit updates it with new claims/sources instead of minting a duplicate). The create path has a preview: the widget picks the first named `createEntity` op as the primary entity,
pills it in the summary, and navigates to its page where the staged data renders.

The enrich path had none of that: an enrich edit contains no named `createEntity` for the story, so `primaryEntityId` stays null - no pill, no navigation, and the summary degrades to a generic "Imported N entities". The staged changes were only discoverable by manually entering edit mode and opening Review edits. The worker already reports the enrich outcome in its job status (`enrich: { mode, targetGeoId, targetHeadline, newClaims, sourcesMinted }`), but the poll proxy dropped the field.

## Change

- `core/chat/inject-types.ts` — new `InjectEnrichInfo` type; the `completed` poll response optionally carries `enrich`.
- `app/api/chat/inject/[jobId]/route.ts` — `extractEnrichInfo()` validates and passes the worker's `enrich` field through (only `mode: 'enriched'`; the target id is format-checked like the route's other inputs).
- `core/hooks/use-inject-job.ts` — job state carries `enrich` to the widget.
- `partials/chat/chat-widget.tsx` — when the completed job is an enrichment: summary becomes "Enriched [story pill] with N new claims and M sources", the widget navigates to the enriched story, and opens the review panel (`bumpReviewVersion()` +
  `setIsReviewOpen(true)`, same calls the chat's `openReviewPanel` tool uses).

Why open the review panel instead of relying on the entity page: an existing story renders its claims through server-seeded data blocks, so locally staged additions are not visible in read mode - the review diff is the only surface that can show what the enrichment adds. The create path renders a brand-new entity entirely from the local store, which is why navigation alone works there.

## Scope / safety

- Ops staging is unchanged: `applyInjectOpsToStore` runs before the branch for every inject type, exactly as before.
- Post, tweet, and create injects are unaffected: the worker only sets `mode: 'enriched'` on the news enrich track, so `extractEnrichInfo` returns null for everything else and the pre-existing code path runs unchanged. The poll response
  change is an additive optional field.
- If the worker doesn't send `enrich` (older deploys), behavior is identical to today.
- No other global state is touched — no `activeSpace`, publish, or editor state writes; the review panel self-selects the space containing the staged actions as it does for a manual open.

## Testing

Tested locally against the live inject worker: enrich inject (existing World Affairs story) → "Enriched …" summary with pill - navigation to the story - review panel opens showing the staged claims/sources diff; create inject on a fresh article - unchanged
behavior (new-entity preview + navigation); typecheck clean on all four files.

### DIFF
--- a/apps/web/app/api/chat/inject/[jobId]/route.ts
+++ b/apps/web/app/api/chat/inject/[jobId]/route.ts
@@ -7,6 +7,7 @@ import { type Edit, decodeEdit } from '@geoprotocol/grc-20';
 import { cookies } from 'next/headers';
 
 import type {
+  InjectEnrichInfo,
   InjectPollResponse,
   SerializedOp,
   SerializedPropertyValue,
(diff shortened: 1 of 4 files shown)


ohohoreilly:
Closing this, and the reason is our side rather than yours.

This has been open since 22 July with no review and no response, and the Vercel check never got past "a member of the Team first needs to authorize it", so it could not even produce a preview to look at. Two months without an answer is not something you should have to keep waiting on, so I would rather close it honestly than leave it open as though a review were still coming.

The gap you identified is real and **still live on master**. `apply-inject-ops.ts:182` still sets the primary from the first `createEntity` carrying a non-empty Name, and an enrich edit has none — the story already exists on-chain, so nothing names it. `primaryEntityId` stays null, and since that is the only thing `chat-widget.tsx` gates the pill and the navigation on, an enrich still lands silently while a create gets a pill and a destination.

Written up as **GEO-2983**, crediting this PR and describing the asymmetry as you framed it. I have recorded your approach as a sketch rather than a verified design, only because it was never reviewed — not as a judgement on it.

Thanks for the write-up, which was clearer about the create/enrich split than anything we had. Sorry it sat this long.

PR #2055, opened 2026-07-24
fix: prevent accidental mobile row drag while scrolling

webmagic123:
Feat:
 On touch devices, trying to scroll the list was being misread as the start of a drag, so the page wouldn't scroll properly.

Fixed:
  The list uses @dnd-kit sensors that decide when a touch becomes a drag. Previously a single `PointerSensor` with `{ distance: 8 }` handled both mouse and touch, so any 8px touch move — including the start of a scroll — began a drag. The fix splits those into separate sensors so each input gets the right activation rule:

  - Mouse: `MouseSensor` with `MOUSE_ACTIVATION = { distance: 10 }`. Unchanged in spirit from the old pointer behaviour — a small deliberate move starts a drag.
  - Touch: `TouchSensor` with `TOUCH_ACTIVATION = { delay: 250, tolerance: 6 }`. This is the actual fix. A distance constraint can't distinguish a drag from a scroll, so touch now requires a *hold* rather than a move: 250ms stationary before a drag begins.
  - Rows are no longer draggable outside edit mode — `disabled: !isEditing` on `useSortable`, plus the listeners spread gated on `isEditing`. Previously read-mode rows were draggable and could fire `onUpdateRelation`.
  - In ranking compose, row swipe actions are disabled while a reorder drag is in progress (`swipeEnabled={!isReordering}`), so the two gestures stop competing.

Note on `tolerance` — it reads backwards, and an earlier revision of this PR got it wrong:

  Under a *delay* constraint, `tolerance` is a **cancel radius**, not a grace allowance. Moving more than that distance before the timer fires *aborts* the pending drag (see the `hasExceededDistance` checks in `@dnd-kit/core`'s `core.cjs.development.js` around :1546). So a **lower** tolerance makes scrolling win more often, and raising it would make accidental drags **more** likely, not less.

  This PR briefly tried `{ delay: 450, tolerance: 12 }`, which was the wrong direction on both counts — the higher tolerance gave the finger a larger movement budget before the drag gave up. That was reverted in 707b5e0d7; the values above are the shipping ones. The reasoning now lives in a comment next to the constants so it doesn't get lost again.

Also reverted `ranking-my-ranking-dnd.tsx` back to master in 70df011a2 — that list never needed the sensor swap, and the change was causing a desktop regression.


b-d055 (changes requested):
resolve conflicts + re-test

apps/web/partials/blocks/table/ranking-my-ranking-dnd.tsx
```diff
@@ -130,7 +134,7 @@ export function RankingMyRankingDndList({
   itemsRef.current = items;
 
   const sensors = useSensors(
-    useSensor(PointerSensor, { activationConstraint: POINTER_ACTIVATION }),
+    useSensor(MouseSensor, { activationConstraint: MOUSE_ACTIVATION }),
```
b-d055: Thanks for restoring the touch window in `707b5e0d` — that one's resolved.

This swap is still open though, and it kills desktop mouse reorder outright. Flagging the mechanism because CI is green and can't see it.

`MouseSensor`'s activator is `onMouseDown`, where `PointerSensor`'s was `onPointerDown` (`@dnd-kit/core` 6.3.1 — `core.cjs.development.js:1679` vs `:1638`). The sortable wrapper above declares its own `onMouseDown` **after** the listeners spread:

```jsx
:99    {...attributes}
:100   {...listeners}                                  // now contributes onMouseDown
:101   onClick={suppressClickAfterDrag}
:102   onClickCapture={suppressClickAfterDrag}
:103   onMouseDown={event => event.stopPropagation()}  // later JSX prop wins - clobbers it
```

Line 103 isn't new. It was harmless while the activator key was `onPointerDown`, a different prop name, so the spread and the explicit handler never collided. With `MouseSensor` they collide and the sensor's own handler is overwritten: no drag, no overlay, `onDragStart` never fires. `useSensors` registers only `MouseSensor` + `TouchSensor`, so there's no keyboard sensor to fall back on either.

I verified this with a probe rather than by reading — mocking `listeners: { onMouseDown: spy }` and firing `mousedown` on `[data-ranking-drag-activator]`, the spy is never called. The same probe with `onPointerDown` does fire, so it isn't a bad selector.

**The updated mock can't catch it.** `ranking-my-ranking-dnd.test.tsx:28` still returns `listeners: { onPointerDown: ... }`, so the mock's listeners never collide with the explicit `onMouseDown` and the suite stays green. Adding `MouseSensor` to the `@dnd-kit/core` mock in `33d9f569` fixed the import error but not the blind spot — the `@dnd-kit/sortable` mock needs `onMouseDown` too.

Please don't fix this by moving the spread last — that would drop the `stopPropagation` the row wants. Merging the two handlers keeps both:

```jsx
onMouseDown={event => {
  event.stopPropagation();
  listeners?.onMouseDown?.(event);
}}
```

b-d055 (approved):

PR #2079, opened 2026-07-29
 feat: add Upvoted and Downvoted tabs to ranking compose

webmagic123:
Adds **Upvoted** and **Downvoted** tabs alongside Global in the ranking compose 
browse section, so you can rank from the things you've already voted on instead 
of hunting for them in the global list.

## How the tabs get their entities

The voted list is its own query rather than a filter over the feed — the feed is
 space-scoped and votes aren't. `userVotes` is paged 50 at a time by `voteType`,
 and each page of ids is hydrated into display entries one page at a time, accum
ulated in `useVoteTabEntities`. The consumer gets `entries` back directly, so th
ere's no second entity query over the accumulated list.

All three tabs share **one panel** rather than one per tab. That's deliberate: r
emounting the panel on every switch would throw away the hydrated pages and re-f
etch the whole list, which is the thing `1289d7800` exists to prevent. It's also
 why this doesn't use Radix `Tabs` despite it already being a dependency — `Tabs
.Content` unmounts on switch. The ARIA wiring is hand-rolled instead: `tablist`/
`tab`/`tabpanel`, roving tabindex, arrow/Home/End with focus following selection, and an `sr-only` `<h2>` labelling the tablist (the old markup had `aria-selected` on `role="button"`, where it's invalid and silently dropped).

Search on a vote tab filters the loaded entries by name on the client. The global tab's search is a server fuzzy query where the id list *is* the result set, so that path can't be reused for an arbitrary voted-id list — the global tab keeps using it unchanged.

## Three things that are less obvious than they look

**A vote has to appear in its tab before the indexer has it.** Voting writes a pending override that's overlaid onto page 0 and deduped against the server list, then cleared once reconciliation confirms indexing. The store is rooted at `user-entity-votes-pending` — a **sibling** of `user-entity-votes`, not a child. TanStack v5 `invalidateQueries` matches by prefix, so a child key means every refresh of the votes list silently wipes every pending override at once.

**The list is ordered server-side, by `orderBy: [VOTED_AT_DESC, OBJECT_ID_ASC]` on the `userVotes` list field.** Sorting `votedAt` client-side only orders the pages already fetched, so past the first 50 a freshly cast vote landed wherever the server happened to put it — and the row you'd just watched appear would vanish the moment its override cleared. `VOTED_AT_*` genuinely does 500 on `userVotesConnection` + cursor, which is what the original cursor-based query was working around; it's fine on the list field with offset, as `ParticipantPositions` already relies on in production.

**Pages are tracked by the ids they were hydrated from, not by index.** Index-only tracking meant a page whose *contents* changed was never re-fetched — so a restored vote landing mid-list never surfaced, and a cached revisit that delivered several id pages at once skipped straight to the newest and stranded the ones before it. Keying on the joined id list fixes both with one mechanism, and lets the gap-walk catch up page by page.

## Hydration failures stop the list

A failed entity fetch settles as `isFetched` with the local-store fallback, so it used to be banked as an empty page and never retried. It now leaves the page as a gap, surfaces the error with a retry that re-fetches both queries, and retires the scroll sentinel — without which the sentinel walks the viewer's entire vote history back to back, nothing hydrating and no spinner to show for it.

Claim kind is resolved against the entity's own space via upstream's `resolveEntitySpaceId`, so a claim verified in its home space isn't downgraded and dropped when it's browsed from another.

## Verification

**3018 tests / 291 files passing.** eslint 0 errors. `next build` exit 0.

Where a test pins a fix, I checked it actually fails without it rather than assuming — the pagination guard, the error-surfacing path, and the content-keyed rehydration were each verified failing against the reverted code.

The voted-ids query is generated rather than hand-written: it lives in `query-fragments.tsx` where codegen's glob finds it, and `core/gql/*` is regenerated against the live schema


apps/web/partials/blocks/table/ranking-compose-global-ranking.tsx
```diff
@@ -404,7 +445,7 @@ export function RankingComposeGlobalRanking({
     setSearchListStableHeight(computeSearchListStableHeight(pageScrollRoot, listContainerRef.current));
   }, [getPageScrollRoot, needsSearchStablePlaceholder]);
 
-  const canLoadMore = hasNextPage && hasVisibleRankableEntities;
+  const canLoadMore = hasNextPage && (browseTab !== 'global' || hasVisibleRankableEntities);
```
b-d055: I wonder if the vote tabs want to be their own data source rather than a filter layered over the global feed.

On a vote tab this reduces to just `hasNextPage`, and the sentinel also renders inside `voteTabEmptyState`, so when the tab is empty or short the sentinel ends up sitting in the viewport and re-firing after every page. Unless I'm missing something upstream, that means opening Upvoted pages through the entire entity list 10 rows at a time until `hasNextPage` flips false, with a "Loading more…" parked under an apparently empty list the whole time.

The knock-on is that every page grows `allRankableEntityIds`, and `useRankingEntryEntities` re-syncs the whole accumulated set each time (`first: stableIds.length`, new query key), so the cost is closer to quadratic than linear. `queries.ts` also notes that `id.in` should stay under `ENTITY_ID_BATCH_SIZE` (50), and this would go well past it.

Something you upvoted a few hundred entities ago won't appear until the global feed has paged past it, so the tab is only ever as complete as how far you happened to scroll.

`userVotesConnection` already returns exactly the ids the tab wants, ordered `VOTED_AT_DESC`, which is arguably a nicer sort for "stuff I upvoted" than global created-at anyway. What if we made it the paging source instead of the filter?

- turn `useUserVotedEntityIds` into a `useInfiniteQuery`, ~50 per page, `getNextPageParam` off `endCursor`. That also removes the `while (hasNextPage)` crawl in `getUserEntityVoteObjectIds`, which currently blocks on N sequential 200-row requests inside one `queryFn` before anything paints.
- hydrate that page directly via `useQueryEntities({ where: { ...blockWhere, id: { in: pageIds } }, first: pageIds.length })`, so the block's filters still apply and we stay under the batch limit.
- point the scroll sentinel at the vote connection's `fetchNextPage` when `browseTab !== 'global'`.

`globalRankByEntityId` comes from `useRankingBlockRelations` rather than the paginated feed, so the ranked/unranked split should still work, and `tabFilteredRankedIds` / `tabFilteredUnrankedIds` / `isLoadingVoteFilter` would all drop out. Opening a tab becomes one vote page plus one entity batch.

b-d055 (changes requested):
@webmagic123 see comment, this is a performance bottleneck/correctness issue

apps/web/partials/blocks/table/ranking-compose-global-ranking.tsx
webmagic123: @b-d055 
Updated.

apps/web/core/blocks/ranking/use-vote-tab-entities.ts
```diff
@@ -0,0 +1,50 @@
...
+  const { filterState, filterMode } = useDataBlock();
+  const blockWhere = React.useMemo(() => filterStateToWhere(filterState, filterMode), [filterState, filterMode]);
+
+  const enabled = direction !== null;
+  const { ids, isLoading, hasNextPage, isFetchingNextPage, fetchNextPage } = useUserVotedEntityIds(
+    direction ?? 'up',
+    enabled
+  );
+
+  const where = React.useMemo(() => ({ ...blockWhere, id: { in: ids } }), [blockWhere, ids]);
```
b-d055: Two separate things here, and the second is a one-liner worth taking on its own.

**The hydration re-does itself every page.** `ids` is the flattened accumulation across all fetched pages, not just the newest, so `where` changes shape on every `fetchNextPage`. That serializes into a new `GeoStore.queryKeys` entry each time, so `E.syncMany` runs again over the entire list: 50 ids, then 100, then 150.

Worth splitting the cost, since only part of it is cacheable. The `merge` loop in `orm.ts` re-runs across every accumulated id on each page, and that work is quadratic no matter what the network does. The network side is quadratic too, but only incidentally: the per-50 batch keys inside `syncMany` are actually stable, so they would dedupe on their own if anything set a `staleTime`. The `QueryClient` is constructed with no defaults, so `staleTime` is 0 and every batch is re-requested. After k pages that is k(k+1)/2 batch calls.

`useRankingAccumulatedRows` already has the shape that avoids all of this, using `upsertRowPage` / `flattenRowPages` from `accumulate-row-pages`: hydrate one page against a stable key, accumulate client-side, never refetch what you already have. Adopting it needs `useUserVotedEntityIds` to expose the per-page id slices rather than only the flattened array, which seems like a small change to make.

**Separately, could we pass `deferUntilFetched: true`?** A configured ranking block always carries at least a `TYPES_PROPERTY` filter (`ranking-block-state.ts` gates on exactly that), so `blockWhere` is never empty on this screen and `where` always has two or more keys. That means the single-key `id.in` fast path at `use-store.tsx:414` never applies here, and on each new key, with `data` still undefined, the selector falls through to the tail branch at `:436`: `new EntityQuery(store.getEntities())`, which materializes the whole store, filters it, sorts the entire filtered set by `updatedAt`, and only then paginates. It lives inside `useSelector` on an atom created with `compare: () => false`, so it re-runs on every store mutation while the fetch is in flight, and with the point above it re-enters once per page rather than once on mount.

`useRankingAccumulatedRows` passes both `deferUntilFetched: true` and `placeholderData: keepPreviousData` for what looks like exactly this reason.

apps/web/core/hooks/use-entity-vote.ts
```diff
@@ -80,6 +81,7 @@ export function useEntityVote({ entityId, spaceId }: UseEntityVoteArgs) {
   const onSuccess = () => {
     queryClient.invalidateQueries({ queryKey: ['entity-vote-count', entityId, 0] });
     queryClient.invalidateQueries({ queryKey: ['user-entity-vote', personalSpaceId, entityId, spaceId, 0] });
+    queryClient.invalidateQueries({ queryKey: [USER_ENTITY_VOTES_QUERY_KEY_ROOT, personalSpaceId ?? null] });
```
b-d055: This prefix-matches the `useInfiniteQuery` in `use-user-voted-entity-ids.ts`, whose key is `[root, personalSpaceId, direction]`, so the two-segment key here matches both directions.

The part worth knowing: on react-query 5.101.2 (what the lockfile resolves), refetching an infinite query with no direction set replays every loaded page in a sequential `do...while` in `infiniteQueryBehavior`, awaiting each before starting the next. It has to be sequential, since `getNextPageParam` can only read the cursor off the page that just resolved. No `maxPages` is set, so it is all of them. Someone paged five deep into Upvoted pays five sequential round trips for one vote, plus the re-hydration from the other comment.

One thing that limits the blast radius: `invalidateQueries` defaults to `refetchType: 'active'`, and only one direction is ever mounted, so the inactive direction just goes stale rather than refetching. But the case that does fire is the bad one, since the vote buttons live inside the rows themselves, so it is the list you are looking at that re-pages under you.

A `setQueryData` would cut this to roughly one request. Removal from the cached pages is clean. Insertion is fuzzier and worth thinking about before committing to it: a cursor-paginated list has no defensible local position for a new id, `endCursor` / `hasNextPage` drift from the server's view, and `useVoteTabEntities` takes display order straight from array index. If that is more than you want to own, `maxPages` is the cheap cap and the only supported one in v5 now that `refetchPage` is gone.

b-d055 (commented):
@webmagic123  Follow-up purely on performance, since the structure is now right. Two things inline, and the second one is independent of the first if you want to take it on its own.

apps/web/core/blocks/ranking/use-vote-tab-entities.ts
webmagic123: @b-d055 
Updated

apps/web/core/hooks/use-entity-vote.ts
webmagic123: @b-d055 Updated.
-The blanket two-segment invalidation is gone. 
A vote removes the entity from the lists it isn't in via setQueryData
- maxPages cap (use-user-voted-entity-ids.ts),

apps/web/core/io/user-entity-votes-by-type-document.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import { parse } from 'graphql';
+
+const USER_ENTITY_VOTES_BY_TYPE_SOURCE = /* GraphQL */ `
+  query UserEntityVotesByType($userId: UUID!, $voteType: Int!, $objectType: Int!, $first: Int!, $after: Cursor) {
+    userVotesConnection(
+      first: $first
+      after: $after
+      condition: { userId: $userId, voteType: $voteType, objectType: $objectType }
+      orderBy: [VOTED_AT_DESC]
```
b-d055: this `orderBy` combined with a cursor returns a 500 from the live API.**

`userVotesConnection(after: ..., orderBy: [VOTED_AT_DESC])` returns `INTERNAL_SERVER_ERROR` for every page after the first. I checked against `https://api-testnet.geobrowser.io/graphql` three ways (no condition, with the `userId` condition, and with `VOTED_AT_ASC`) and all of them fail the same way as soon as a cursor is passed. The identical query with the default `NATURAL` ordering pages fine, so it is the `VOTED_AT_*` ordering plus `after` that the server rejects, not the condition or the page size.

What that means at runtime: a user with more than 50 upvotes opens the Upvoted tab, the sentinel prefetches page 2, `graphql()` throws on `json.errors` (`subgraph/graphql.ts:44`), and `Effect.runPromise` rejects. `useUserVotedEntityIds` never reads `query.error`, so `hasNextPage` stays true and `isFetchingNextPage` drops back to false. The prefetch effect at `ranking-compose-global-ranking.tsx:494` then re-fires on every dependency change, so the tab sits under a permanent "Loading more..." in a retry loop, capped at the first 50 votes. There is no error state and no way out except closing the tab.

Two things worth doing whichever way the ordering gets resolved:

1. Read `query.error` and stop the sentinel on failure. That turns an invisible infinite loop into something the user can see and we can debug.
2. This document lives in a `.ts` file, which sits outside the `./**/*.tsx` glob in `codegen.ts`, so codegen never validates it against the schema. Moving it into the glob would have caught this at build time rather than at runtime.

If the server side cannot be fixed quickly, `NATURAL` ordering plus a client-side sort of the accumulated pages would unblock the tab, at the cost of ordering being correct only within what has been fetched so far.

b-d055 (commented):
@webmagic123  see comments

apps/web/core/io/user-entity-votes-by-type-document.ts
webmagic123: @b-d055 Updated.

I have not widen to all document files in this PR.

apps/web/partials/blocks/table/ranking-compose-global-ranking.tsx
```diff
@@ -326,6 +333,51 @@ export function RankingComposeGlobalRanking({
...
+    hasNextPage: voteTabHasNextPage,
+    isFetchingNextPage: isFetchingVoteTabNextPage,
+    isError: isVoteTabError,
+    retry: retryVoteTab,
+    fetchNextPage: fetchVoteTabNextPage,
+  } = useVoteTabEntities(voteDirection);
+
+  const isVoteTab = voteDirection !== null;
+
+  const { entries: voteTabEntries } = useRankingEntryEntities(spaceId, isVoteTab ? voteTabIds : []);
```
b-d055: This reintroduces the accumulated re-hydration one layer up. `voteTabIds` is the full ordered accumulation across every fetched page, and `useRankingEntryEntities` feeds it straight into `useQueryEntities({ where: { id: { in: stableIds } }, first: stableIds.length })`. So every `fetchNextPage` grows the list, which serializes into a new `GeoStore.queryKeys` entry, and `E.syncMany` runs again over everything fetched so far: 50 ids, then 100, then 150. It is also a second hydration of entities `useVoteTabEntities` already loaded on the vote-tab path, since it fetched them to build `orderedIds` and check vote kinds, so each page ends up synced twice.

Suggestion: have `useVoteTabEntities` return display entries rather than bare ids, and skip `useRankingEntryEntities` entirely when `isVoteTab`. That hook already holds the full entity objects per page (it reads them for `resolveEntityResponseKind`) and already accumulates page slices with `upsertIdPage`, so accumulating `RankingEntryDisplay` rows there instead of id strings is a small change. The pickers it needs (`pickValueEntryBySpace`, `pickValueBySpace`, `pickImage`) are already exported from `use-ranking-entry-entities.ts`, so the display mapping can be reused as is. That removes the quadratic re-sync and the duplicate fetch in one move, and the global tab keeps using `rankableEntriesById` untouched.

b-d055 (approved):

PR #2086, opened 2026-07-30
fix(entity-page): scale entity, space and topic titles on mobile

webmagic123:
Entity, space and topic page titles now scale down on smaller screens instead of staying at the full desktop size, so long titles fit and read well on mobile.

## Sizes

Titles use a dedicated `entityTitle` token rather than sharing `mainPage` with section headings, which has a larger desktop size and no responsive ramp:

| Viewport | Size / line height |
| --- | --- |
| ≥768px | 44px / 46px |
| 410–767px | 36px / 38px |
| ≤409px | 26px / 30px |

The steps live in `styles.css` as plain `:root` media queries redefining the token's custom properties, so every consumer scales without restating sizes at the call site.

## The three-line clamp is intentionally gone

Browse-mode titles previously ran through `Truncate` / `ClampedText` with `maxLines={3}`. **That clamp has been removed on purpose, and this PR should not be read as having dropped it by accident.**

[The designer asked for it directly](https://github.com/EE-Solutions/geo-project-issues/issues/223#issuecomment-5148164646) — jwalkingjew, Jul 31: *"can we remove the truncation at 3 lines?"*, with a screenshot of a clamped title on a phone.

It also largely stops earning its keep: it compensated for a title fixed at 52px (`mainPage`) at every width, and the token now steps down to 26px on a phone.

Being precise about the trade, since the two headers clamped differently. The **space** header used `Truncate`, which cuts with no way to reveal the rest — removing that is a straight win. The **entity** header used `ClampedText`, which does render a More/Less toggle, so there a pathological name (a pasted paragraph) now runs full length instead of collapsing behind More. That cost is accepted deliberately. The description below the title still clamps to three lines (`ENTITY_DESCRIPTION_MAX_LINES`); long titles wrap via `wrap-break-word`.

`partials/entity-page/entity-page-title.test.tsx` pins this, so changing it back has to be deliberate.

## Changes

- `EntityPageTitle` is the one title component for entity and space headers, replacing the inline `Truncate` / `Text` / `PageStringField` blocks in `editable-entity-header.tsx` and `editable-space-header.tsx`.
- `EntityPageActions` (menu, history, create, votes) is split out and composed beside `EntityPageMetadataHeader`. On master these controls lived in `EditableHeading`, which the side panel rendered *outside* the `!isRelationPage` gate — so a relation page kept its menu and history by accident of layout. Extracting them put them under that gate and relation pages lost both; `640fc513d` fixes it. **To be clear: that regression was introduced by this PR, not inherited from master.** The split now enforces what was previously incidental — "is there type metadata to show" is a different question from "can this entity be acted on".
- `TopicPageView` uses the same `entityTitle` token. On master both it and the entity header used `mainPage`, so they matched; converting only the entity header would have shipped the same route drawing the same slot 8px apart depending on whether the entity was typed as a Topic.
- The title accessory (the personal-space verify badge) is centred by the row instead of master's `mt-[9px]`, which was tuned against a title fixed at 52px and could not follow a token that now steps 44/36/26.
- The profile layout passes the route's `entityId` to `EntityPageActions`, not `profile.id`, so the title, metadata, actions, history and voting all act on one canonical id.
- Both create links carry an accessible name; they are icon-only and otherwise reach a screen reader as the bare URL. Their visibility is unchanged from master — note that `sm:hidden` here means hidden *on phones*, since this repo's breakpoints are desktop-first (`sm` = max-width 639px).

## Regression coverage

The refactor moved rendering responsibilities between components, and each review pass surfaced another surface it had quietly changed. Those paths are now covered directly:

- `showHeader={false}` hides the metadata strip and every action, including the vote controls inside it, while leaving the title alone — and the reverse.
- A relation side panel keeps the menu and history while hiding type metadata and voting.
- `fallbackName` titles browse mode only and never pre-fills the editing textarea, where one keystroke would commit a preview name into the graph.
- A space subroute still renders a supplied `actionsComponent`, without the space-only history and menu beside it.
- Names are read with both the entity id and the space id, on the read and the write side.
- Both create links have an accessible name.
- The title uses the token and does not clamp.
- The profile layout keys the title, the store and the actions on one id.


apps/web/partials/entity-page/entity-page-title.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+import { ZERO_WIDTH_SPACE } from '~/core/constants';
+
+import { Spacer } from '~/design-system/spacer';
+
+/**
+ * ≥768 → 44/46, 410–767 → 36/38, ≤409 → 26/30.
+ */
+const titleTypographyClassName =
+  'font-semibold tracking-[-0.5px] text-[44px] leading-[46px] max-[767px]:text-[36px] max-[767px]:leading-[38px] max-[409px]:text-[26px] max-[409px]:leading-[30px]';
```
b-d055: This hardcodes what the `mainPage` token already defines.

`mainPage` (`styles.css:92-95`) is 52px/56px, no responsive behavior, and is still used by six page level `<h1>`s: `topic-page-view.tsx:131`, `space/pending/[topicId]/page.tsx:76`, `home/page.tsx:124`, `export-wallet/page.tsx:6`, `import/generate.tsx:389`, `community-calls/pre-join.tsx:89`. Entity and space titles are now 44/46 against those six at 52/56.

`entity-page-body.tsx:188-190` routes a topic entity to `TopicPageView` (52/56) and a non-topic entity here (44/46), so the same route renders the same slot 8px apart by entity type. The comment at `topic-page-view.tsx:123-133` documents the opposite invariant and is now inaccurate.

Suggested: keep `variant="mainPage"` here and make the token itself responsive. `text-mainPage` compiles to `font-size: var(--text-mainPage)`, with the same pattern for line height, letter spacing and weight, so redefining those custom properties in a plain `:root` media query scales every consumer. The overrides have to sit in ordinary CSS, not in `@theme`, since `@theme` variables must be declared top level and cannot be nested in a media query.

Separately, the boundaries are off by one against the doc comment on line 14. Tailwind v4 compiles arbitrary `max-[Npx]:` to a strict `<`:

```
.max-[767px]:text-[36px]  ->  @media (width < 767px)
.max-[409px]:text-[26px]  ->  @media (width < 409px)
```

Line 14 specifies `410-767 -> 36/38` and `<=409 -> 26/30`. Widths of exactly 767 and 409 land in the wrong band, as does any fractional width in [767, 768) and [409, 410). Corrected:

```
max-[768px]:text-[36px] max-[768px]:leading-[38px] max-[410px]:text-[26px] max-[410px]:leading-[30px]
```

Not `md:` for the upper bound. `@custom-variant md` is an inclusive `max-width: 767px`, but Tailwind v4 emits arbitrary `max-[...]` variants before custom variants regardless of class order. `md:text-[36px]` with `max-[410px]:text-[26px]` compiles `max-[410px]` first, and equal specificity gives `md` the win at 400px, dropping the 26px band.

`design-system/theme/typography.ts:16-22` defines `mainPage` as `2.75rem`/`2.875rem`, matching 44/46. Those numbers feed `typography` at `:137`, which is not exported; only the `textStyles` class map at `:171` is consumed.

apps/web/partials/entity-page/entity-page-title.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+type EntityPageTitleProps = {
+  value: string;
+  isEditing: boolean;
+  onChange: (value: string) => void;
+  placeholder?: string;
+  className?: string;
+  /** Rendered inline directly after the title in browse mode, e.g. a verification badge. */
+  accessory?: React.ReactNode;
```
b-d055: `accessory` is declared unconditionally but read only in the browse branch (`:56`, `:61`), so passing it with `isEditing={true}` is a no-op. Master behaved the same way, so this is about the type rather than behavior.

The same shape let the two browse branches diverge on `wrap-break-word`. A discriminated union (`{ mode: 'edit' } | { mode: 'browse'; accessory? }`) or two components sharing the typography constant would prevent that.

apps/web/partials/entity-page/editable-entity-header.tsx
```diff
@@ -50,98 +33,16 @@ export function EditableHeading({
...
-            )}
-          </HistoryPanel>
-          <EntityPageContextMenu entityId={entityId} entityName={name || ''} spaceId={spaceId} />
-        </div>
-      </div>
-
-      <HistoryDiffSlideUp selection={diffSelection} onClose={clearDiffSelection} />
-    </>
+    <EntityPageTitle
+      value={name ?? fallbackName ?? ''}
```
b-d055: This passes `name ?? fallbackName ?? ''` for both modes. Master used `name ?? ''` in edit and fell back only in browse, and the prop doc comment still says "Shown in browse mode".

Effect: the ranking compose flow prefills the textarea with the preview name, and one keystroke persists it via `storage.entities.name.set`. Intended?

apps/web/partials/entity-page/entity-page-title.tsx
b-d055: This hardcodes what the `mainPage` token already defines.

`mainPage` (`styles.css:92-95`) is 52px/56px, no responsive behavior, and is still used by six page level `<h1>`s: `topic-page-view.tsx:131`, `space/pending/[topicId]/page.tsx:76`, `home/page.tsx:124`, `export-wallet/page.tsx:6`, `import/generate.tsx:389`, `community-calls/pre-join.tsx:89`. Entity and space titles are now 44/46 against those six at 52/56.

`entity-page-body.tsx:188-190` routes a topic entity to `TopicPageView` (52/56) and a non-topic entity here (44/46), so the same route renders the same slot 8px apart by entity type. The comment at `topic-page-view.tsx:123-133` documents the opposite invariant and is now inaccurate.

Suggested: keep `variant="mainPage"` here and make the token itself responsive. `text-mainPage` compiles to `font-size: var(--text-mainPage)`, with the same pattern for line height, letter spacing and weight, so redefining those custom properties in a plain `:root` media query scales every consumer. The overrides have to sit in ordinary CSS, not in `@theme`, since `@theme` variables must be declared top level and cannot be nested in a media query.

Separately, the boundaries are off by one against the doc comment on line 14. Tailwind v4 compiles arbitrary `max-[Npx]:` to a strict `<`:

```
.max-[767px]:text-[36px]  ->  @media (width < 767px)
.max-[409px]:text-[26px]  ->  @media (width < 409px)
```

Line 14 specifies `410-767 -> 36/38` and `<=409 -> 26/30`. Widths of exactly 767 and 409 land in the wrong band, as does any fractional width in [767, 768) and [409, 410). Corrected:

```
max-[768px]:text-[36px] max-[768px]:leading-[38px] max-[410px]:text-[26px] max-[410px]:leading-[30px]
```

Not `md:` for the upper bound. `@custom-variant md` is an inclusive `max-width: 767px`, but Tailwind v4 emits arbitrary `max-[...]` variants before custom variants regardless of class order. `md:text-[36px]` with `max-[410px]:text-[26px]` compiles `max-[410px]` first, and equal specificity gives `md` the win at 400px, dropping the 26px band.

`design-system/theme/typography.ts:16-22` defines `mainPage` as `2.75rem`/`2.875rem`, matching 44/46. Those numbers feed `typography` at `:137`, which is not exported; only the `textStyles` class map at `:171` is consumed.

apps/web/partials/entity-page/entity-page-metadata-header.tsx
```diff
@@ -261,36 +270,93 @@ export function EntityPageMetadataHeader({ id, spaceId, isVoteable = false }: En
...
+            <ReadableRelationsGroup
+              entityId={id}
+              spaceId={spaceId}
+              propertyId={SystemIds.TYPES_PROPERTY}
+              isMetadataHeader={true}
+            />
+          )}
+        </div>
+        <div className="flex shrink-0 items-center gap-5">
+          <EntityPageContextMenu entityId={id} entityName={name || ''} spaceId={spaceId} />
```
b-d055: These moved here from `EditableHeading`, but this component does not render everywhere `EditableHeading` does.

`entity-page-body.tsx:212` renders `EditableHeading` unconditionally; `:221` gates this component on `!isRelationPage`. On master both controls were inside `EditableHeading` (`editable-entity-header.tsx:100-144`), so relation entities had them. Nothing else supplies them now: `EntitySidePanelHeader` renders only close, breadcrumb, mode toggle and fullscreen, and `editable-entity-header.tsx` is title only.

Side panel only. `isRelationPage` exists on that path alone (`use-entity-page-surface-data.ts:36` into `entity-side-panel.tsx:242`). Route level pages go through `entity-page-header.tsx:30` and are unaffected.

Repro: open an entity in the side panel, click the `RelationSmall` icon on a relation chip. The types and actions row is absent.

Underlying cause is that page actions are now gated by a condition answering a different question, whether the entity has type metadata to show. Same coupling at `entity-page-header.tsx:30`. Extracting `<EntityPageActions entityId spaceId isVoteable />` and composing it beside the metadata strip resolves this case and decouples the two.

b-d055 (commented):
@webmagic123  Two changes requested.

The title typography no longer goes through the mainPage token. Six other page levels still do, so entity and space titles render 44/46 where every other page title renders 52/56. On the entity route a topic entity and a non-topic entity now render the same slot at different sizes. Inline on entity-page-title.tsx.

Relation entities in the side panel lose the context menu and version history. Both moved into EntityPageMetadataHeader, which is gated on !isRelationPage. Route level pages are unaffected. Inline on entity-page-metadata-header.tsx.

PR #2096, opened 2026-07-31
fix: add category and status filters to space governance

webmagic123:
Feature:
 The space governance tab now has category and status dropdown filters (like governance home), replacing the old single proposal-type toggle.

Worked:
  - Added a shared query module governance-proposal-query.ts with the filter types (`GovernanceProposalCategory`, `GovernanceProposalStatusFilter`), URL parsers (`parseGovernanceCategory`, `parseGovernanceStatus`), and a single REST fetch
  - Swapped the old `GovernanceProposalTypeFilter `for a new `GovernanceProposalFilters `(the two dropdowns) on the governance page, driven by `proposalCategory`/`proposalStatus `search params (with backward-compat parsing of the legacy proposalType param).
  - Reworked `governance-proposals-list.tsx `to fetch through the new filtered query instead of the old three-bucket (executable/active/completed) fetch.
  - Constrained the governance content width to match the space header/tabs by wrapping it in `EntityPageContentContainer`.


apps/web/partials/governance/governance-proposal-query.ts
```diff
@@ -0,0 +1,162 @@
...
+};
+
+const PAGE_SIZE = 100;
+
+const SETTINGS_ACTION_TYPES = [
+  'UpdateVotingSettings',
+  'SetTopic',
+  'UnsetTopic',
+  'TopicDeclared',
+  'TopicRemoved',
```
b-d055, line 41: **Settings filter always shows "No proposals yet"**

`TopicDeclared` and `TopicRemoved` aren't valid action types on the REST API, so it rejects the whole request with a 400 (`Invalid action types: TopicDeclared, TopicRemoved`). `fetchProposalsForSpaceByGovernanceFilters` logs that and returns `[]`, so every Settings view renders the empty state.

Repro on the preview: open https://geogenesis-git-fix-add-dropdowns-to-governance-tabs-geo-browser.vercel.app/space/a19c345ab9866679b001d7d2138d88a1/governance?proposalStatus=accepted (100 proposals), then pick **Settings** in the category dropdown. The list goes empty, and Pending and Rejected are empty too.

Direct API check:
```sh
# 400 with this list
curl "https://api-testnet.geobrowser.io/proposals/space/a19c345ab9866679b001d7d2138d88a1/status?limit=5&status=ACCEPTED&actionTypes=UpdateVotingSettings,SetTopic,UnsetTopic,TopicDeclared,TopicRemoved,SubspaceVerified"
```
Without these two entries, the root space returns 18 accepted settings proposals (15 "Add Verified Space", 3 "Add Related Space").

`SetTopic`/`UnsetTopic` already cover topics, so these two can go. `VALID_ACTION_TYPES` in `core/io/rest/validation.ts` also lists them, which is why `validateActionTypes` let them through; worth bringing that in line with the API. This list came from home's fetcher on master, so the fix also repairs the Settings filter on governance home.

```suggestion
```

apps/web/partials/governance/governance-proposal-filters.tsx
```diff
@@ -0,0 +1,69 @@
...
+}: {
+  spaceId: string;
+  category: GovernanceProposalCategory;
+  status: GovernanceProposalStatusFilter;
+  proposalId?: string | null;
+}) {
+  const params = new URLSearchParams();
+  if (category !== 'all') params.set('proposalCategory', category);
+  if (status !== 'pending') params.set('proposalStatus', status);
+  if (proposalId) params.set('proposalId', proposalId);
```
b-d055, line 37: **Opening a proposal resets the filters**

These filter links keep `proposalId`, but going the other way drops the filters. Row links in `governance-proposals-list.tsx` are `/space/${spaceId}/governance?proposalId=${p.id}`, and `useCloseProposal` pushes the bare `/space/${spaceId}/governance`.

Repro: on https://geogenesis-git-fix-add-dropdowns-to-governance-tabs-geo-browser.vercel.app/space/a19c345ab9866679b001d7d2138d88a1/governance?proposalStatus=accepted (100 proposals), click any proposal. The URL becomes `?proposalId=...`, the list behind the drawer switches back to Pending (1 proposal on this space), and closing the drawer stays on Pending.

Master has the same href, but its default view listed everything, so nothing visibly changed. With Pending as the default, anyone browsing Accepted or Rejected loses their place every time they open a proposal.

Fix: keep `proposalCategory`/`proposalStatus` in the row href, and in the non-home branch of `use-close-proposal.ts` (for example, the current search params minus `proposalId`).

apps/web/partials/governance/governance-proposal-query.ts
```diff
@@ -0,0 +1,162 @@
...
+  spaceId: string;
+  memberSpaceId: string;
+  proposalType?: 'membership' | 'content';
+  category?: GovernanceProposalCategory;
+  status?: GovernanceProposalStatusFilter;
+}): Promise<readonly ApiProposalListItem[]> {
+  const config = Environment.getConfig();
+
+  const params = new URLSearchParams();
+  params.set('limit', String(PAGE_SIZE));
```
b-d055, line 115: **"Load more" can never trigger, so each view stops at 100**

This fetches a single page with `limit=100`, and `GovernanceProposalsList` paginates that with `PAGE_SIZE = 100`. So `hasMore = combinedProposals.length > endIndex` is always false, and `GovernanceProposalsListInfiniteScroll`, its new `key`, and the category/status args on `loadMoreProposalsAction` never run.

The root space has more than 100 accepted and more than 100 rejected proposals (the API returns a non-null `nextCursor` for both). Each view shows exactly 100 with no way to reach older ones: https://geogenesis-git-fix-add-dropdowns-to-governance-tabs-geo-browser.vercel.app/space/a19c345ab9866679b001d7d2138d88a1/governance?proposalStatus=rejected

Either page with `cursor` (as `core/io/fetch-sidebar-counts.ts` does) and pass it through the load-more action, or remove the infinite-scroll path and treat the view as capped. Copilot's suppressed comment on this is still unaddressed.

b-d055 (commented):
@webmagic123  see comments

b-d055 (approved):

PR #2098, opened 2026-07-31
Feat/community tab

webmagic123:


apps/web/core/community/community-graphql.ts
```diff
@@ -0,0 +1,82 @@
...
+  return after ? `, after: "${after}"` : '';
+}
+
+export async function runQuery<T>(label: string, query: string, signal?: AbortController['signal']): Promise<T | null> {
+  const result = await Effect.runPromise(
+    Effect.either(graphql<T>({ endpoint: Environment.getConfig().api, query, signal }))
+  );
+
+  if (Either.isLeft(result)) {
+    console.error(`community: "${label}" query failed`, result.left);
```
b-d055, line 69: I think this makes backend failures invisible to the whole feature.

Every caller coalesces this `null` into an empty array or map (`data?.x?.nodes ?? []`, and `collectConnection` just breaks at line 73). `fetchProfilesBySpaceIds` does the same thing, returning default profiles rather than failing. So an outage doesn't surface as an error, it surfaces as "no data". `fetchCuratorLeaderboard` resolves normally, the route returns a 200 with an all-zero leaderboard, and the user sees "0 active curators, 0 rankings", which looks exactly like a quiet space.

Partial failure is the nastier case. If just `fetchVoteCounts` fails, the table still renders with believable rows, quietly missing vote counts, and the tiebreaker on line 362 ends up sorting on data that isn't there.

Worth flagging that this also means adding an `isError` branch in `curator-leaderboard-section.tsx` wouldn't help. `useQuery` resolves fine with zeros, so `isError` never fires. I think the fix has to live here, either letting `runQuery` propagate the failure or returning a discriminated result so callers can tell "empty" apart from "broken".

(I did check `graphql()` and `restFetch()`, and they genuinely do funnel network, HTTP status, JSON parse and GraphQL `errors[]` failures into Effect's typed channel, so `Effect.either` catches all of those. The one thing that escapes is a synchronous `Environment.getConfig()` throw on line 48, but that only fires with unset mainnet env vars, so it isn't the current testnet setup.)

apps/web/core/community/community-graphql.ts
```diff
@@ -0,0 +1,82 @@
...
+export async function collectConnection<TNode>(
+  label: string,
+  buildQuery: (after: string | null) => string,
+  selectConnection: (data: Record<string, Connection<TNode> | undefined>) => Connection<TNode> | undefined,
+  signal?: AbortController['signal']
+): Promise<TNode[]> {
+  const nodes: TNode[] = [];
+  let after: string | null = null;
+
+  for (let page = 0; page < MAX_PAGES; page++) {
```
b-d055, line 87: This caps out silently, and I think it's feeding numbers that users read as authoritative.

There are only two real breaks in here: query failure on 73, and `hasNextPage: false` on 77. If neither fires by page 19, the loop condition just goes false and line 81 returns `nodes` with no exception, no flag, and no warning. From the caller's side, "I got everything" and "I hit the cap" look identical.

The trouble is where those partial sets end up:

* `rankings.total` (line 465) is counted off the truncated `relations` array at 117-125, not a server `totalCount`
* `newsStories.total` (line 466) is `storyIds.length` at 335, off the truncated `stories` array

Ceilings per call site: rankings and submissions 20 x 500 = 10,000 relations, votes 20 x 1,000 = 20,000, news stories 20 x `ENTITY_PAGE_SIZE` 200 = 4,000. `fetch-space-bounties.ts:23` uses the same helper at 20 x 100 = 2,000 bounties.

Cheapest fix is probably just surfacing it, a `truncated` flag on the return or even a `console.warn`. Better would be sourcing the headline counts from a server-side `totalCount` instead of `array.length`.

apps/web/core/community/fetch-curator-leaderboard.ts
```diff
@@ -0,0 +1,471 @@
...
+
+    for (const node of data?.entitiesConnection?.nodes ?? []) {
+      rankCreatedAt.set(node.id, node.createdAt);
+    }
+  }
+
+  let total = 0;
+
+  for (const relation of relations) {
+    if (!isWithinWindow(rankCreatedAt.get(relation.fromEntityId), window)) continue;
```
b-d055: Related to the pagination cap.

That relations query up at 76-92 doesn't carry a time filter at all. The period window gets applied here, in memory, instead. Compare `fetchVoteCounts` on 139 with `votedAtFilter`, and `fetchNewsStoryCounts` on 249 with `createdAtFilter`, both of which push the bound server-side. (`fetchSubmissionCounts` is halfway there: its `executedAt` filter on 201-204 is server-side, but only at the proposals stage, so its first-stage links fetch on 180-196 is just as unbounded as this one.)

Two things fall out of that:

1. Picking "week" still pulls the whole historical relation set, up to the 10,000 row cap, before throwing nearly all of it away. So a short period doesn't reduce the truncation risk, which is the opposite of what you'd assume.
2. There's no `orderBy` on the query, so nothing guarantees recent relations come back first. Once the cap is hit, which rows survive is arbitrary. A "week" count can quietly miss rows inside the window because older data ate the cap.

Pushing the window into the GraphQL filter would fix the correctness exposure and a good chunk of the latency at the same time.

apps/web/core/community/fetch-curator-leaderboard.ts
```diff
@@ -0,0 +1,471 @@
...
+    data => data.entitiesConnection,
+    signal
+  );
+
+  const storyIds = [...new Set(stories.map(story => story.id).filter(Boolean))];
+  if (storyIds.length === 0) return { perCurator, total: 0 };
+
+  const firstVersionKeyByStory = new Map<string, bigint>();
+
+  for (const ids of chunk(storyIds, ID_CHUNK_SIZE)) {
```
b-d055: This one's the expensive one. It's a sequential chunk loop with a paginated call nested inside it, so worst case is chunks x up to 20 pages, all in series.

There are four sequential chunk loops in this file (99, 206, 273, 307), all shaped `for (const ids of chunk(...)) { await runQuery(...) }`, and all four could be `Promise.all`'d. This is just the one that compounds.

Worth saying the top-level fan-out on 418-423 is already right, `Promise.all` over the four metric fetchers. The sequential work is purely in the inner loops, so the fix belongs there and not at that level.

For context on why this adds up: `community-tab-page.tsx:15-18` mounts four sections that fire four separate network calls on mount. The three `/bounties` calls use different `taskStatusId` values, so they get different react-query keys and don't dedupe. Both routes declare `export const revalidate = 60` but read `request.url` and `await context.params`, which forces them dynamic in Next 15/16, so the revalidate is inert and every visit pays full freight. `staleTime: 60_000` only helps on refetch within a session, not on first mount.

Each of those three `/bounties` calls also re-fetches contributor avatars and profiles independently (`fetch-space-bounties.ts:170-173`). I couldn't find any memo, LRU, `React.cache()` or HTTP caching in `core/io/rest`, so overlapping contributors get fetched from scratch each time. I didn't measure how much overlap there actually is across the three status sets, so treat that last bit as directional.

b-d055 (changes requested):
@webmagic123  see comments

b-d055 (approved):

PR #2105, opened 2026-08-03
feat: replace the space template picker with one fixed overview template

webmagic123:


apps/web/core/hooks/use-deploy-space.ts
```diff
@@ -284,6 +285,19 @@ async function createDaoSpace({
     throw new Error('Timed out waiting for DAO space to index.');
   }
 
+  // Seed the overview tab. Deliberately after indexing and deliberately non-fatal
+  try {
+    await seedNewSpaceOverview({
```
b-d055: This also fires from the "Turn into space" flow, and I don't think that path was considered.

entity-to-space-dialog.tsx:40 still has its own copy of the 10 template options this PR deletes from create-space-dialog.tsx, and it's reachable with no gate (entity-page-context-menu.tsx:200). It calls deploy() at :87 with topicId set to the entity id, so spaceHomeEntityId is the same entity that cloneEntityIntoSpace copies BLOCKS onto at :98.

For the template backed types (dao, industry, and so on) that leaves three sets of blocks on one home entity: the template blocks from generateOpsForSpaceType, these four seeded blocks, and the cloned ones. The clone keeps each relation's original position while the seeded blocks start a fresh generateBetween(cursor, null) chain from null, and since Position.generate() is itself generateBetween(null, null) both sets land around a0 and interleave once sorted (use-blocks.ts:33). So the overview comes out jumbled instead of in the 1 to 4 order the spec asks for.

GEO-2466 says to remove the picker entirely and use the same template for all newly created spaces, so I think this path needs covering too. Either skip seeding there or point that dialog at the fixed template as well.

apps/web/core/hooks/use-deploy-space.ts
```diff
@@ -284,6 +285,19 @@ async function createDaoSpace({
     throw new Error('Timed out waiting for DAO space to index.');
   }
 
+  // Seed the overview tab. Deliberately after indexing and deliberately non-fatal
```
b-d055: Can we drop the await here and let this run in the background?

Right now deploy() doesn't resolve until the whole seeding chain finishes, which is an IPFS proposeEdit plus two userops with receipt waits (create and vote), three if the vote doesn't auto execute. PendingCreatedSpaceRunner only calls router.push after deploy() resolves (pending-created-space-runner.tsx:68-84), and EntityToSpaceDialog blocks dismissal the entire time (entity-to-space-dialog.tsx:133), so both paths get noticeably slower.

The wait doesn't buy anything either, since nothing waits for the seeded edit to be indexed, so the user can still land on an empty overview. It's already non fatal, so firing it without awaiting keeps it off the path. Worth updating the comments at create-space-dialog.tsx:185-190 and pending-created-space-runner.tsx:20-22 at the same time, since they still describe the chain as ending at the index wait.

apps/web/core/utils/contracts/seed-new-space-overview.ts
```diff
@@ -0,0 +1,61 @@
...
+export async function seedNewSpaceOverview({
+  smartAccount,
+  spaceId,
+  spaceAddress,
+  spaceHomeEntityId,
+  authorSpaceId,
+}: SeedNewSpaceOverviewParams): Promise<void> {
+  const ops = generateNewSpaceTemplateOps({ spaceId, spaceHomeEntityId });
+
+  const proposal = await geo.daoSpaces.proposeEdit({
```
b-d055, line 34: Could we wrap these two calls (this one and sendUserOperation at :43) the way use-publish.ts does?

The same calls there are retry wrapped: proposeEdit at use-publish.ts:362-388 with retrySchedule('proposeEdit', Duration.minutes(1)) and sendUserOperation at :428-451 with a 10s schedule. executeFastProposal below is fine since it's inside runEffectEither and retries its own governance userops, so this is the only unprotected part of the chain. A transient IPFS or bundler hiccup here drops the whole template.

They also have no Effect.withSpan, so unlike every other write in this flow a failure leaves nothing in Sentry. Combined with the catch in use-deploy-space.ts, a failed seed is completely invisible today. Adding the span is the cheap half of the fix even if the retry is more work.

Heads up that retrySchedule isn't exported from use-publish.ts yet, so it needs an export or to move somewhere shared.

apps/web/core/utils/contracts/generate-new-space-template.ts
```diff
@@ -0,0 +1,243 @@
...
+ *   4. Recent news     — query block, explore view, infinite scroll, link-ingestion tool
+ */
+
+const ROLLING_RANKING_SUBMISSION_FREQUENCY_HOURS = 24;
+
+const RECENT_DEBATES_PAGE_SIZE = 3;
+const TRENDING_CLAIMS_PAGE_SIZE = 5;
+
+const RECENT_NEWS_TYPE_IDS = [
+  'e550fe517e904b2c8fffdf13408f5634',
```
b-d055: These four are byte identical to News story, Episode, Tweet and Paper in explore-constants.ts:3-8, and those are the only two places each string appears. Since CLAIM_TYPE_ID and DEBATE_TYPE_ID right below are named imports, could we export named constants for these four and use them here too? This array defines what Recent news means for every space from now on, so a typo in a bare hex id would be hard to spot.

Don't reuse EXPLORE_ENTITY_TYPE_IDS wholesale though, the lists genuinely differ. Explore carries six types this block excludes, and this block includes Claim, which Explore doesn't.

apps/web/core/utils/contracts/generate-new-space-template.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    .flatMap(op => ('values' in op && Array.isArray(op.values) ? (op.values as OpValue[]) : []))
+    .map(value => value.value?.value)
+    .filter((value): value is string => typeof value === 'string' && value.includes('"filter"'))
+    .map(parseFiltersSync);
+}
+
+const typeFilters = (parsed: FilterStateResult) =>
+  parsed.filters.filter(filter => filter.columnId === SystemIds.TYPES_PROPERTY);
+
+describe('generateNewSpaceTemplateOps', () => {
```
b-d055, line 75: Could we add one test that asserts the shape of the overview itself?

Right now these three only cover filter mode and space scoping, and deleting three of the four blocks still passes all of them, since blockFilters() samples whatever it finds and the assertions are toHaveLength(1), length > 0 and every(...). I ran the generator to check: it emits 46 ops and 4 Blocks relations off the home entity in ascending position order, and none of that is covered.

Something asserting the four Blocks relations (count, ascending positions, the view relation on each) would catch the failure mode this file actually has, which is the template silently drifting from the spec.

b-d055 (commented):
@webmagic123  Five things inline. The first one is the only one I'd ask about/resolve - the rest are small and I'd rather get them in now than carry them.

apps/web/core/utils/contracts/generate-new-space-template.test.ts
webmagic123: @b-d055 
Updated

apps/web/core/utils/contracts/generate-new-space-template.ts
webmagic123: @b-d055 
Updated.

apps/web/core/hooks/use-deploy-space.ts
webmagic123: @b-d055 
Updated.

apps/web/core/hooks/use-deploy-space.ts
webmagic123: @b-d055 
Fixed.

apps/web/partials/entity-page/entity-to-space-dialog.tsx
```diff
@@ -102,19 +88,22 @@ export function EntityToSpaceDialog({
...
     }
-  };
+  }, [address, deploy, entityName, entityId, sourceSpaceId, storage, reportError, onOpenChange]);
 
-  const onSelectType = (spaceType: SpaceType) => {
-    setStep('creating');
-    createSpace(spaceType);
-  };
+  const startedRef = React.useRef(false);
+  React.useEffect(() => {
```
b-d055: This auto-fires the deploy as soon as the dialog opens, which means there's no confirmation step anymore before minting a space on chain.

Before this change the template picker doubled as the confirm: you clicked "Turn into space", saw ten options, and picked one. Now entity-page-context-menu.tsx:207 sets open to true and this effect immediately calls createSpace(). That menu item also has no isMember or isEditor gate (compare "Move to..." at :180 which does), and Dialog.Root at :121 ignores every close attempt, so a single misclick in the context menu mints a DAO space and there's no way to back out.

Could we put a confirm step in front of it? Even just a "Create space from this entity" button in the dialog would restore what the picker was doing, without bringing the templates back.

apps/web/partials/entity-page/entity-to-space-dialog.tsx
```diff
@@ -102,19 +88,22 @@ export function EntityToSpaceDialog({
...
       const message = describeError(error);
-      // Send the user back to the type-selection step so they can pick again or retry.
-      setStep('select-type');
-      reportError(`Space creation failed: ${message}`, () => {
-        setStep('creating');
-        createSpace(spaceType);
-      });
+      resetState();
+      onOpenChange(false);
+      reportError(`Space creation failed: ${message}`, () => onOpenChange(true));
```
b-d055: This retry can mint a second space.

Reopening sets startedRef back to false at :100 and the effect refires createSpace(). If the original failure came after the factory tx landed, the most likely one being the "Timed out waiting for DAO space to index" throw in use-deploy-space.ts after the ~120s wait, the space already exists on chain and the retry creates another one. pending-created-space.ts:19-23 calls this out explicitly: DAO deploy is not idempotent and re-running mints a second space.

The create-space flow guards this two ways, a module level autoRunFired flag (create-space-dialog.tsx:74-78, whose comment explains why a component ref wasn't enough) and the 'pending' status check on the atom. This dialog has neither. Could the retry either resolve to the already-created space or be dropped in favour of pointing the user at the space that did get made?

apps/web/partials/entity-page/entity-to-space-dialog.tsx
```diff
@@ -68,27 +51,30 @@ export function EntityToSpaceDialog({
...
+    setStep('creating');
+
     try {
       const spaceId = await deploy({
-        type: spaceType,
+        type: NEW_SPACE_TYPE,
         spaceName: entityName,
         governanceType: 'DAO',
         topicId: entityId,
+        seedOverviewTemplate: false,
```
b-d055: Just want to confirm this is the intended product call rather than only a fix for the ordering problem I raised.

GEO-2466 says to use the same template for all newly created spaces, and this path now gets no template at all. The comment at :38 gives the rationale, that the cloned blocks are the overview here, which reads fine to me. But it does mean two ways of making a space produce two different overviews, so worth a quick check with Preston or Jeff before it ships.

b-d055 (approved):

PR #2108, opened 2026-08-05
feat(debates): publish attributed debate claims read from geo-chat

TheHoneySeller:
## What

Publishes **attributed, fact/opinion-classified claims** for a debate — sourced from geo-chat's extraction step rather than extracted inline.

When the publish sweep publishes a completed debate, it now:
- Reads pre-computed claims from geo-chat's `GET /debates/{id}/claims` (`loadDebateClaims` in `debate-source.ts`), using geo-chat's canonical `turns` for the transcript blocks and its `claims` (keyed by `turn_index`) for the claim entities.
- Mints each claim as a Geo **Claim** entity via `buildDebatePublishDraft`: `Is factual?` (BOOLEAN), block → `Claims` relation, `Sources` → Debate, attributed to the speaker's profile space (`Authors`).
- Falls back to the plain transcript (no claims) when a debate predates extraction or extraction produced nothing.

Pairs with **geobrowser/geo-chat#35** (Part A), which does the extraction + attribution inside the media job and serves the payload.

## Why this changed from the original inline approach

This PR originally extracted claims **inline** during the publish sweep — a 120s synchronous poll bolted onto a short, bounded cron task, which caps publish throughput (Nik's review). The publish sweep is the wrong place for a long, retryable job.

geo-chat already owns every long debate step (compose, render, Whisper transcription) in a leased worker queue. Extraction now runs there, next to transcription; geogenesis just **reads the result**. This drops the inline extractor entirely:
- **Deleted** `core/debates/server/extract-claims.ts` (+ test) and its call in `publish-debate.ts`.
- **Removed** the `DEBATE_CLAIM_EXTRACTION_*` / `EXTRACTION_API_*` env from geogenesis (that config now lives in geo-chat).
- `buildDebatePublishDraft` is unchanged — it already accepts `claims` and wires `Is factual?` / `block → Claims` / `Sources`.

## Design note

geo-chat is the single authority on **turn boundaries**: it merges Whisper segments into turns once and returns turns + claims keyed by `turn_index`, so published transcript blocks and claims align (no dual-merge drift between the two services).

## Testing

- `buildDebatePublishDraft` tests unchanged and passing (attributed Claim draft: `Is factual?`, `Claims`, `Sources`).
- `debate-source` tests cover the `/claims` read + the transcript fallback.
- Validated end-to-end against a live extraction dry-run: the same attributed, character-identical claims reach the draft.


nikgraf (approved):
On a high level the PR looks "acceptable".

one concern from my side: Claims extraction is a blocking interaction. In case it fails publishing fails. Not sure this is intentional or not.
Since I don't have the credentials to a claims extractions server or code to run it locally I can't test it.

Here two suggestions Codex found:

#1 -- The referenced extraction service [requires ](https://github.com/geo-explorers/extraction-api/blob/de61046f96be11a0cf738c35f5d02f14c1b4a36e/src/api/main.py#L75-L101)X-API-Key[ on every non-docs request](https://github.com/geo-explorers/extraction-api/blob/de61046f96be11a0cf738c35f5d02f14c1b4a36e/src/api/main.py#L75-L101). An operator can follow this PR's example, omit the "optional" key, and get a 401 for every enqueue. The catch then publishes without claims, so the enabled feature silently does nothing. Mark the key as required when extraction is enabled and add an explicit missing-key configuration guard.
#2 -- Neither the enqueue request nor a poll request has an abort signal. If either server accepts the connection but never completes the response, fetch never returns, the loop deadline is not checked, and the [] fallback does not run. Use one overall deadline and pass its AbortSignal to both requests.

PR #2109, opened 2026-08-06
feat: keep action-initiated onboarding on-page and replay the action

webmagic123:


apps/web/partials/comments/comments-section.tsx
```diff
@@ -222,14 +225,32 @@ export function CommentSection({ entityId, spaceId }: CommentSectionProps) {
...
     (text: string, ancestorComments?: Array<{ id: string; spaceId: string }>) => {
+      // The composer only lets a signed-in user type, so smartAccount is present here.
+      if (!personalSpaceId) {
+        const id = `comment:${entityId}:${(commentSeqRef.current += 1)}`;
+        enqueuePendingAction({
+          id,
+          label: 'your comment',
+          requires: 'personalSpace',
+          run: () =>
+            createComment({
```
b-d055, line 327: This closure captures `createComment`, which is a `useCallback` in `use-create-comment.tsx` with `personalSpaceId` in its dep array (line 478). We only reach this branch when `personalSpaceId` is null, so the captured copy permanently sees null.

When the runner replays it, it hits the guard at `use-create-comment.tsx:152`, which returns null instead of throwing, so the runner's success path runs and drops the action from the queue. The comment is never posted, and the user gets "Personal space required to comment" right when their account became ready.

I don't think this path can ever succeed. Worth a repro before merge.

apps/web/partials/space-page/space-members-popover-members-request-button.tsx
```diff
@@ -30,9 +37,29 @@ export function SpaceMembersPopoverMemberRequestButton({
...
+      return;
+    }
+    setOptimisticRequested(true);
+    enqueuePendingAction({
+      id: `join:${spaceId}`,
+      label: 'your membership request',
+      requires: 'personalSpace',
+      run: () =>
+        new Promise<void>((resolve, reject) => {
+          requestToBeMember(undefined, { onSuccess: () => resolve(), onError: err => reject(err) });
```
b-d055: The `onSuccess`/`onError` passed to `mutate` are dispatched through the MutationObserver, which gates on `hasListeners()` (query-core 5.101.2, mutationObserver.js:77). If this component unmounts before the mutation settles, neither fires and this promise never resolves or rejects, so `pending-actions-runner.tsx` keeps the id in `runningRef` and skips it forever, with no error and no retry.

This button is the most exposed case: `space-members-popover.tsx:27` renders `{open ? content : null}` with no `forceMount`, so closing the popover unmounts it. The same pattern is in the other three join buttons and in `queueVote`.

apps/web/partials/entity-page/entity-vote-buttons.tsx
```diff
@@ -279,23 +318,19 @@ export function EntityVoteButtons({
...
             ? 'Sign in to vote'
-            : isAccountSetupPending
-              ? 'Finishing account setup…'
-              : isConnected
-                ? upvoteActive
-                  ? 'Remove upvote'
-                  : 'Upvote'
-                : 'Connect wallet to vote'
+            : !isConnected
+              ? 'Vote now — saved until your account is ready'
```
b-d055: Dropping the `disabled` guard means the arrow is now clickable while the account is still setting up. If the user clicks and then dismisses the Privy modal, `hasQueuedVote` stays true, so the clear effects at lines 120 and 128 never run, and the filled arrow plus bumped score persist until reload, for a vote that was never cast and never will be.

Previously this was disabled with a "Finishing account setup" tooltip.

b-d055 (commented):
@webmagic123  see suggested spots to fix

apps/web/partials/comments/comments-section.tsx
b-d055: Resolved by the cached-identity refactor. `createComment` now reads the smart account and personal space from the query cache at call time (`use-create-comment.tsx:162,168`), so the replayed closure no longer carries a stale null. The flow also changed shape: the optimistic row is created up front and only the publish is replayed with the same `commentId`, and the replay throws when the result is not published (`comments-section.tsx:333`) instead of resolving as success.

Note GitHub re-anchored this thread to line 327, which is now inside the new replay closure, so it points at the fix rather than the defect. Resolving.

apps/web/partials/space-page/space-members-popover-members-request-button.tsx
b-d055: Half of this is resolved. Switching to `mutateAsync` removes the mechanism I described: the promise now settles because it comes from `Mutation.execute()` rather than the observer callbacks, so there is no silent permanent wedge here anymore.

The other half is still open, in a different file. `use-request-to-be-member.ts` did not pick up the `readCachedPersonalSpace` helper the vote and comment paths adopted, so `handleRequestToBeMember` still closes over `personalSpaceId` (deps at :76) and throws at :42. If this button unmounts before registration lands, the replay runs against frozen options, reports "You need a registered personal space ID to request membership" at the moment the user has one, and retry loops on the same frozen closure.

Non-blocking since it fails loudly and nothing is lost, but it is a one-line fix with the helper already imported two files over. Leaving this thread open to track it.

PR #2146, opened 2026-08-13
fix(explore): stop showing published-entity affordances on drafts

jwalkingjew:
## What

Two oddities reported while edits were in review:

1. Feed rows in explore-style data blocks showed **"—"** where the age goes, healing once the edits were published or discarded.
2. Unpublished entities showed a **"Publish changes before responding"** notice in browse mode — once per row across a whole import.

## Why (corrected mid-review)

My first pass assumed drafts simply have no creation time and hid the placeholder. That was wrong, and follow-up screenshots showed published entities with no age at all.

The real cause is **`GeoStore.merge`**. When an entity has any local edit, merge rebuilds it from `id, name, spaces, description, types, values, relations` and **dropped `createdAt`/`updatedAt`**. Entities with no local edits skip that branch and return the remote object intact — which is exactly why rows healed on publish or discard. Anything reading a timestamp through the sync store lost it for as long as an edit sat in review.

Two related problems surfaced from the same investigation:

- The `Entity` query never selected `createdAt`, only `updatedAt`, so block rows showed *time since last update* — days off on entities that keep getting edited.
- The response controls were gated on `isLocal && !hasBeenPublished`, but **nothing in production ever sets `hasBeenPublished: true`** (only tests do). A row therefore looks unpublished forever once written locally, which turned "hide the notice" into "hide responses permanently on published claims".

## How

- `GeoStore.merge` carries the indexed timestamps through a local edit.
- `Entity` query selects `createdAt`; block rows prefer it for age. Regenerating the codegen document map was required — `graphql()` resolves documents by exact string match and returns `{}` for anything unrecognised, so an edited query silently becomes a no-op. The type check caught it (`Type 'unknown' is not assignable to TypedDocumentNode`); `bun run gql:codegen` fixed it, and the generated diff is confined to that one document.
- Responses hide only when the entity carries **no indexed timestamps** — genuinely never published — instead of when local-edit flags are set. Published entities keep their controls even with stale local rows.
- `formatExploreRelativeTime` returns `null` rather than `'—'` when there is no timestamp, so a true draft has no age instead of a dash that reads as broken.

## Testing

- New: merge carries timestamps through a local edit, and leaves a local-only entity without them.
- New: responses hidden for an entity with no indexed timestamps; still shown for a published entity carrying local-edit flags.
- New: `formatExploreRelativeTime` returns null for absent/zero timestamps.
- Sync, explore, entity-page, responses and table-block suites: 200 passing. `next build` (with type check) and eslint clean.


jwalkingjew:
Closing — after further testing this turned out not to be a bug. Scrapping the work rather than landing changes to the sync store and shared GraphQL queries on a false premise.

PR #2153, opened 2026-08-13
fix(debates): snappier claim responses — instant voter avatar and an unblocked Debate switch

jwalkingjew:
Two fixes for the same feeling: acting on a claim makes you wait on indexing you shouldn't have to wait on.

## 1. The voter avatar appears immediately

Adding your position to a claim puts your space id into the responder avatar group right away, but the avatar itself took seconds to paint.

The cause is cache shape, not indexing. `RankingAggregatedSubmitterAvatars` cached profiles and spaces under a key built from the **whole set** of responder ids. Adding one id minted a new key with nothing behind it, so the entire group — not just yours — dropped to gradient placeholders until a fresh `POST /profile/batch` landed. In the batched claim views it was worse: those render with `queriesEnabled={false}` and rely on a cache primed *before* your response existed, so the new key had no data **and** nothing would ever fetch it. The placeholder sat there until indexing rebuilt the primed caches.

Cache profiles **per space id** (`profileBySpaceIdQueryKey`) via a new `useProfilesBySpaceIds` hook:

- Ids already resolved keep rendering when one is added — no group-wide flash.
- Your own profile is already cached from the navbar's `useGeoProfile`, which now seeds the by-space-id key from its address lookup (`/profile/address` and `/profile/space` return the same record). Your avatar paints with **no request at all**.
- Same-tick per-id lookups are coalesced by `loadProfileBySpaceId` into a single `POST /profile/batch`, so the finer granularity costs no extra round trips.
- `loadClaimResponderMetadataCaches` primes the per-id keys, and `ClaimResponderAvatars` warms your own profile while an optimistic response is showing.

Spaces keep their set-shaped key but hold previous data through the refetch (`keepPreviousData`). Entries are read by id, so a held-over map can only serve ids it genuinely resolved.

## 2. The Debate switch moves before the response indexes

On a claim's entity page or in a data block, taking a side lets you flip Debate straight away. In the claims side panel or the rematch picker, the same action disabled the switch and showed "Publishing your response…" for the minute or so indexing took — making standing ready a second trip back to a claim you had just answered.

The underlying constraint is real: geo-chat only accepts readiness once it can see an *indexed* active response. But the entity page already solved it. A click there records an **intent** rather than sending — the switch moves now, and the queue request is held until geo-chat's copy of the response agrees with the one the intent was recorded against. The hub simply never used that machine; it disabled itself instead.

This extracts that machine out of `ClaimDebateReadiness` into `useClaimDebateReadiness` and has both surfaces share it. The hub keeps its own presentation (the explanation line, the `activeDebate` block) and drops the indexing gate along with the one-shot `useClaimReadiness` mutation it no longer needs.

Sharing it also makes the two surfaces **one** switch rather than two racing on the same server state: the intent lives in the query cache under a key both resolve to, so it survives an unmount (tab switch, re-sorted list, navigation) and cannot be double-sent.

## Testing

New tests:
- `use-profiles-by-space-ids.test.tsx` — single batched request; adding an id keeps the resolved ones and fetches only the new one; a seeded profile serves with queries disabled and no request.
- `ranking-aggregated-submitter-avatars.test.tsx` — prepending a submitter renders all three avatars with zero fetches.
- `use-geo-profile.test.tsx` — seeds the by-space-id cache; does not cache a wallet address as a space id for unregistered wallets.
- `claim-readiness-toggle.test.tsx` rewritten for the shared machine, including the case this PR is about: the switch moves while the response is publishing, sends exactly once after geo-chat confirms, and keeps a held intent across an unmount.

`bunx tsc --noEmit` and `bun run build` pass. The 32 existing `ClaimDebateReadiness` tests pass unchanged against the extracted hook.

**Full suite: 232 files / 2391 tests, all passing.** 19 of those files can't even load under the default local config (a Node webstorage/jsdom `localStorage` conflict unrelated to this change), so they were run against a temporary Storage shim to get real CI-equivalent signal — which is how the four suites that needed their mocks updated for the new machine (`matchmaking-claim-card`, `claims-tab`, `matches-tab`, `rematch-page-client`) were caught. The shim is local-only and is not part of this branch.

Not verified in a live browser: reproducing either fix needs a real wallet and an on-chain response.


apps/web/core/debates/use-claim-debate-readiness.ts
```diff
@@ -0,0 +1,283 @@
...
+
+      if (intent.desiredReady === intent.confirmedReady) {
+        if (backendReady === intent.confirmedReady && !intent.error) setIntent(null);
+        return;
+      }
+
+      // Only watch the response run while we are still waiting on it. The run is retired the moment
+      // geo-chat reports the position, so policing it past that point dropped the intent at the
+      // instant it came good — the switch flicking itself off just before the card resettled.
+      if (intent.desiredReady && readiness.viewer_response?.position !== intent.expectedPosition) {
```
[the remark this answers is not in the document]
jwalkingjew: Confirmed and fixed in 0ea0e442 — thanks, this was a real regression.

`RematchClaimCard` was filling `readiness.viewer_response` from `localPosition`, so the shared machine saw the server agreeing the moment the viewer clicked and sent `joinQueue` before there was an indexed response. Before this PR the hub toggle was simply disabled during publishing, so the path was unreachable; sharing the machine opened it.

Fixed at the caller, as you suggested: the picker now passes `serverLocalPosition` in `viewer_response`. Display is unaffected because `RespondableControls` already prefers the indexing snapshot and only falls back to this field. `waits for the response to settle before standing the viewer ready` now asserts `joinQueue` stays uncalled until geo-chat catches up, which fails without the change.

apps/web/core/debates/matchmaking/claim-readiness-toggle.tsx
```diff
@@ -8,37 +8,39 @@ import { Text } from '~/design-system/text';
...
+    isSaving,
+    error,
+    viewerPosition,
+    toggle,
+  } = useClaimDebateReadiness({
+    readiness,
+    entityId: claim.claim_entity_id,
+    spaceId: claim.space_id,
+    canEnable: !activeDebate,
+  });
```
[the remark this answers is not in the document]
jwalkingjew: Confirmed and fixed in 0ea0e442. Reachable exactly as described: the auto-opt-in arms on the first position and fires on settle, while `viewer_debate_ready` stays false until it lands — so the switch still reads off, and pressing it recorded a second intent. Both then POSTed.

Routed through the same machine, per your first suggestion. `useClaimDebateReadiness` now exposes `setReady(boolean)` alongside `toggle()`, and `useReadinessOnFirstPosition` records `setReady(true)` instead of calling `useClaimReadiness().mutate`. The intent already owned the wait for geo-chat, so the hook lost its `responseSettled` plumbing and its second effect along with the duplicate send.

`stands the viewer ready only once, even as the claim keeps refetching` now exercises both mechanisms against one intent. Worth noting it only caught this after I fixed the queue mocks not being cleared between tests — it was passing on leaked call history.

apps/web/core/debates/use-claim-debate-readiness.ts
```diff
@@ -0,0 +1,283 @@
...
+            };
+          });
+
+          if (retryAfterRefetch) {
+            joinQueue.reset();
+            void queryClient.refetchQueries({ queryKey: ['debates', 'claims', spaceId] }).finally(() => {
+              updateIntent(current =>
+                current?.desiredReady && current.refreshing ? { ...current, refreshing: false } : current
+              );
+            });
```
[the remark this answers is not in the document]
jwalkingjew: Confirmed and fixed in 0ea0e442. The `refetchQueries` promise resolves immediately when nothing matches, so on the hub the retry re-sent against exactly the state that had just 409ed — spending the one retry `hasRetried` allows for nothing.

Now refetches all three families that can carry the response — `[debates, claims, spaceId]` plus `debateQueryKeys.matchmakingClaimsRoot(accountKey)` and `debateQueryKeys.matches(accountKey)` — and waits on all of them before clearing `refreshing`. Used the canonical key factory rather than re-deriving the shapes, so this stays correct if they move.

New test `refetches every readiness family before retrying claim_response_required` spies on `refetchQueries` and asserts all three; it fails against the old single-family refetch.

apps/web/core/debates/matchmaking/matchmaking-claim-card.tsx
```diff
@@ -193,7 +194,35 @@ function RespondableControls({
...
+  const { profilesBySpaceId } = useProfilesBySpaceIds(viewerSpaceIds);
+  const viewerProfile = personalSpaceId ? profilesBySpaceId.get(personalSpaceId) : undefined;
+
+  const optimisticPositions = React.useMemo(
+    () =>
+      withViewerPosition({
+        positions,
+        responseKind: readiness.response_kind,
+        serverPosition: readiness.viewer_response?.position ?? null,
+        viewerPosition,
```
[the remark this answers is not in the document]
jwalkingjew: Confirmed and fixed in 0ea0e442 — same root cause as your comment on `use-claim-debate-readiness.ts`, so one change covers both.

You are right that `withViewerPosition` needs the position the summaries already reflect, and that the rematch adapter was handing it the optimistic one instead — `viewerPosition === serverPosition` returned early, so the picker never got the avatar the hub card does.

Rather than add a second parameter, I fixed the field: `RematchClaimCard` now puts `serverLocalPosition` in `readiness.viewer_response`, which is what both consumers were already documented to expect. `RespondableControls` derives the displayed side from the indexing snapshot, so nothing needed the optimistic value there.

One thing the fix exposed: the rematch suite mocked `useEntityResponseIndexingSnapshot` and `useEntityResponse` independently, so a test could assert an optimistic side the snapshot denied. In production `optimisticResponse` is derived from that snapshot. The mock now derives it the same way.

apps/web/core/debates/matchmaking/matchmaking-claim-card.tsx
```diff
@@ -262,6 +288,77 @@ function RespondableControls({
...
+
+  const withViewer = (side: DebateClaimPositionSummary): DebateClaimPositionSummary => ({
+    ...side,
+    total_count: side.total_count + 1,
+    participants: [viewer, ...side.participants],
+  });
+  const withoutViewer = (side: DebateClaimPositionSummary): DebateClaimPositionSummary => ({
+    ...side,
+    total_count: Math.max(0, side.total_count - 1),
+    participants: side.participants.filter(participant => participant.profile_space_id !== viewerSpaceId),
```
b-d055: Consider `ID.equals` here instead of raw `!==`. This is the only raw comparison against `profile_space_id` in the repo, and `use-debate-votes.tsx:153` normalizes before the same comparison.

`viewerSpaceId` is a graph id, and `core/io/schema.ts:5` defines `HexId` as `/^[0-9a-f]{32}$/i`, so hyphens are rejected there. geo-chat ids are treated as possibly dashed elsewhere in this directory (`use-debate-votes.tsx:123` runs `ID.uuidToHex` on one), and `validateSpaceId` accepts both forms, so a dashed id passes validation and misses this filter.

Tested against the exported function with `019fedb1-0c41-...` vs `019fedb10c41...`: the viewer appears in both sides' `participants`, with their avatar beside a count of zero since line 337 still decrements. Worth confirming the geo-chat format on your end.

```suggestion
    participants: side.participants.filter(participant => !ID.equals(participant.profile_space_id, viewerSpaceId)),
```

The removal branch is also untested. The new test passes `viewer_response: null`, so `serverPosition` is null and line 343 never fires.

apps/web/app/space/[id]/(space)/debates/rematches/[sessionId]/rematch-page-client.tsx
```diff
@@ -696,16 +686,33 @@ function RematchClaimCard({
...
+  // response off the indexing snapshot for display, and uses this field for the two questions only
+  // the server can answer: whether it is safe to send readiness yet, and whether the position
+  // summaries already count the viewer. Handing it the optimistic answer claimed the server agreed
+  // the instant the viewer clicked, which sent readiness before there was an indexed response to
+  // hang it on and suppressed the optimistic avatar the card would otherwise add.
   const readiness: MatchmakingReadiness = {
     response_kind: responseKind,
     viewer_response:
-      localPosition === null
+      serverLocalPosition === null
```
Comment (b-d055) [inline apps/web/app/space/[id]/(space)/debates/rematches/[sessionId]/rematch-page-client.tsx:713]: Re: `serverLocalPosition` at :646, which feeds this field.

`positions` comes from `rematchPositionSummaries`, which derives holders from `claim.participants` without reading `currentUserId`. `serverLocalPosition` reads it, and it is the only input to `readiness.viewer_response`.

`useCurrentGeoChatUserId` returns null until the token exchange lands. Is this sequence reachable: viewer already holds Agree, summaries show their avatar, they click Disagree, so `viewerPosition = false` and `serverPosition = null`. `withViewer` adds them to Disagree and the removal branch does not run, since no side has `position === null`. Avatar on both sides, Agree keeps its count.

Deriving `serverPosition` from the same source as the summaries, or skipping the optimistic adjustment while `currentUserId` is null, would close it.

apps/web/core/hooks/use-profiles-by-space-ids.ts
```diff
@@ -0,0 +1,48 @@
...
+  const uniqueSpaceIds = React.useMemo(() => [...new Set(spaceIds.filter(Boolean))], [spaceIds]);
+
+  return useQueries({
+    queries: uniqueSpaceIds.map(spaceId => ({
+      queryKey: profileBySpaceIdQueryKey(spaceId),
+      queryFn: () => loadProfileBySpaceId(spaceId),
+      enabled,
+      staleTime: 60_000,
+    })),
+    combine: results => {
```
b-d055: Consider wrapping this `combine` in `React.useCallback` keyed on `uniqueSpaceIds`.

Recreated each render, it misses query-core's guard at `queriesObserver.js:148` (`combine !== this.#lastCombine`) and re-runs. `replaceEqualDeep` recurses into the returned `Map`, fails `isPlainObject`, and passes it through, so `profilesBySpaceId` gets a new identity every render. Measured 5 renders producing 5 distinct Maps with unchanged inputs; the `useQuery` this replaced returned a stable one. Adding the `useCallback` brought it to 1.

Memoizing the `queries` array had no effect on its own, since `setQueries` already early-returns. Downstream memos in `ranking-period-metadata.tsx` depend on this identity. Worth pinning in the test file this already ships.

apps/web/core/debates/use-claim-debate-readiness.ts
```diff
@@ -0,0 +1,305 @@
...
+    spaceId,
+    responseKind: readiness.response_kind,
+  });
+  // Privy blanks the account while it rehydrates, and the intent is keyed on it — re-keying
+  // switches to an empty entry *and* wipes the one being held. A card remounting through that
+  // window (which is what the hub's list does when a claim changes section) would silently drop
+  // readiness the viewer is waiting on, flicking the switch off. Hold the last settled account
+  // until Privy actually has an answer; a real sign-out still re-keys, which is the point.
+  const settledAccountKey = React.useRef(accountKey);
+  if (authReady) settledAccountKey.current = accountKey;
```
b-d055: Is this hold needed? `privy.ready` is a one-shot false to true flag on a provider mounted once at the root, so `authReady` should not return to false, and during boot `accountKey` is null for every instance. The rehydration case in the comment happens with `ready` staying true, which this does not catch. The only thing exercising the backwards transition is `claim-readiness-toggle.test.tsx:191-212`, which flips the mock directly.

The ref write is also in the render body. `reactStrictMode` is commented out at `next.config.ts:25`, and `eslint.config.mjs` pulls in only `eslint-config-next/typescript`, which omits the `react-hooks` plugin, so nothing flags it. If the hold stays, `useState` in render is the sanctioned form.

apps/web/app/space/[id]/(space)/debates/rematches/[sessionId]/rematch-page-client.tsx
```diff
@@ -749,29 +756,36 @@ function RematchClaimCard({
...
+  readiness: MatchmakingReadiness;
   localPosition: boolean | null;
   /** Whether {@link localPosition} is this client's own in-flight submission. */
   pickedHere: boolean;
-  /** Whether geo-chat's copy of the response has caught up with {@link localPosition}. */
-  responseSettled: boolean;
   alreadyReady: boolean;
 }) {
-  const setReadiness = useClaimReadiness();
+  const { setReady } = useClaimDebateReadiness({
```
Comment (b-d055) [inline apps/web/app/space/[id]/(space)/debates/rematches/[sessionId]/rematch-page-client.tsx:808]: This card runs two `useClaimDebateReadiness` instances over one shared cache entry, this one and the card's `ClaimReadinessToggle`. They pass different `canEnable`: `true` here, `!activeDebate` in the toggle. That works while `activeDebate` is unset on this page. Consider aligning them now.

The duplicate-send guard holds, and the test at :1064-1092 covers it with both mounted.

apps/web/core/hooks/use-spaces-by-ids.ts
```diff
@@ -33,6 +33,11 @@ export function useSpacesByIds(spaceIds: string[] = [], enabled = true): UseSpac
       };
     },
     enabled: enabled && normalizedIds.length > 0,
+    // The key is the whole id set, so adding one id (the viewer's own space, the moment they
+    // respond to a claim) would otherwise empty `spacesById` until a refetch lands and blank
+    // out every space image that was already on screen. Entries are looked up by id, so the
+    // held-over map can only ever serve ids it genuinely resolved.
+    placeholderData: keepPreviousData,
```
b-d055, line 46: `enabled: false` and empty id lists no longer return an empty map. `data` is undefined and status is pending, so the placeholder branch fires and callers receive the previous fetch's map. This affects `ranking-period-metadata.tsx:104`, `data-block-scope-dropdown.tsx:257`, and `table-block-editable-title.tsx:20`, all of which only `.get()` current ids today. Worth covering in the comment alongside the `spacesById` rationale.

b-d055 (commented):
@jwalkingjew a few commnets I think are worth looking at/addressing to keep things clean

apps/web/core/debates/matchmaking/matchmaking-claim-card.tsx
jwalkingjew: Fixed in 00a09d49. Took the `ID.equals` suggestion, and went a bit further on the surrounding logic.

Confirmed the raw comparison was the only one against `profile_space_id` in the repo, and that `use-debate-votes.tsx:153` normalizes for the same reason. I have not confirmed the geo-chat format either — which is the argument for normalizing rather than for finding out.

The removal also no longer keys off which side the server reports. It rebuilds the participant list on every side and adds the viewer to the picked one, so the viewer can only ever appear once regardless of whether `serverPosition` and the summaries agree about who they are. Counts still follow `serverPosition`, since `participants` is capped and absence from it does not mean absence from the count.

You were right that the removal branch was untested. Now covered by `moves the viewer off the side they left when geo-chat reports a %s id`, parameterized over both id forms; the hyphenated case fails against the old `!==`.

apps/web/app/space/[id]/(space)/debates/rematches/[sessionId]/rematch-page-client.tsx
Comment (jwalkingjew) [reply apps/web/app/space/[id]/(space)/debates/rematches/[sessionId]/rematch-page-client.tsx:713]: Reachable, and fixed in 00a09d49. Confirmed `useCurrentGeoChatUserId` returns null until the token exchange resolves, and that `rematchPositionSummaries` never reads it — so the two disagree exactly as you describe.

Took your second option. `MatchmakingClaimCard` gained `viewerIdentityPending`, and the picker passes `currentUserId === null`; the card then renders the summaries untouched rather than adjusting from a "no position" that only means "not known yet". Deriving `serverPosition` from the summaries would have needed the same id to locate the viewer, so it moves the problem rather than closing it.

The rebuilt-every-side removal from your other comment also makes this non-catastrophic on its own — the viewer can no longer land on two sides — but the count would still have been wrong, so the gate is doing real work.

Covered by `leaves the summaries alone while the viewer cannot be identified in them`.

apps/web/core/hooks/use-profiles-by-space-ids.ts
jwalkingjew: Fixed in 00a09d49 — `combine` is now a `useCallback` keyed on `uniqueSpaceIds`.

Your reading of the guard is right, and so is the consequence. I had noticed the Map identity churn while writing this and talked myself out of it on the grounds that it cannot loop — which was true and beside the point. It regressed the stable identity the `useQuery` it replaced provided, and the memos in `ranking-period-metadata.tsx` are downstream of exactly that.

Pinned in the test file as you suggested: `keeps one map identity across renders while the inputs are unchanged` asserts `toBe` across two rerenders, and fails against the inline `combine`.

Also matches your finding that memoizing `queries` does nothing on its own — I left that unmemoized.

apps/web/core/debates/use-claim-debate-readiness.ts
jwalkingjew: No, it is not needed. Removed in 00a09d49, along with the test that only passed because it drove the mock backwards.

You are right on every point. `PrivyProvider` is mounted once in `providers.tsx`, so `ready` is one-shot; during boot the account is null for every instance and no intent exists to protect; and the rehydration case I described in the comment keeps `ready` true, so the guard never sees it. It guarded nothing, and the only evidence it worked was a test asserting a transition that cannot happen.

Worth being explicit about how it got there: I never reproduced the toggle flicker. I found that a changing `accountKey` re-keys the intent and that `useDebateReadinessIntent` wipes the old entry, wrote a test that forced that transition, and treated it passing as confirmation. That is backwards, and your comment is the reason it is out.

The `!authReady` early return went with it rather than staying as a lone guard, for the same reason — nothing demonstrates it firing. The re-key-wipes-the-entry behaviour is still there if a real case for it turns up.

Left the `ready` field on the `useGeoChatAuth` mocks, since it matches the real hook.

apps/web/app/space/[id]/(space)/debates/rematches/[sessionId]/rematch-page-client.tsx
Comment (jwalkingjew) [reply apps/web/app/space/[id]/(space)/debates/rematches/[sessionId]/rematch-page-client.tsx:808]: Aligned in 00a09d49. Agreed it works today only because the picker never sets `activeDebate` — two instances over one cache entry disagreeing about whether readiness is allowed is the kind of thing that stays harmless right up until it does not.

`RematchClaimCard` now holds a single `activeDebate` local, passes it to the card and `!activeDebate` to `useReadinessOnFirstPosition`, and the hook takes `canEnable` as a parameter rather than hardcoding `true`. It is a `const false` with a comment, which is blunt, but it makes the picker having no active-debate signal explicit and gives the two paths one source instead of two that happen to agree.

`DebateRematchClaim` carries no `active_debate`, so wiring it to something real needs geo-chat to send it.

apps/web/core/hooks/use-spaces-by-ids.ts
jwalkingjew: Documented in 00a09d49. Confirmed your reading: with no data of its own the placeholder branch fires, so a disabled hook or an empty id list now yields the previous fetch's map rather than an empty one.

The comment now states the trade and, more usefully, the condition that keeps it safe — every caller looks entries up by an id it is currently rendering, which a stale map either answers correctly or not at all. What would break is enumerating the map or reading its size as the current id count. That is the thing worth warning the next person about, so it is spelled out rather than left as "all callers `.get()` today", which stops being true silently.

Checked the three call sites you listed plus `navbar-breadcrumb.tsx:87`, which reads `spaces` — that stays correct independently, since `select` derives it from `requestedIds`.

PR #2155, opened 2026-08-13
fix: let readers expand a long entity name in view mode

webmagic123:


apps/web/design-system/clamped-text.tsx
```diff
@@ -0,0 +1,153 @@
...
+    const measure = () => {
+      setIsOverflowing(isEllipsisActive(el, maxLines));
+      setLastLinePx(readLineHeightPx(el));
+    };
+    measure();
+
+    if (typeof ResizeObserver === 'undefined') return;
+
+    const ro = new ResizeObserver(measure);
+    ro.observe(el);
```
b-d055: One thing I'd like to sort out before merge, plus a few smaller notes.

**Observer now watches the element whose width it changes** (this line)

The gutter moved from the text element onto the wrapper (line 122), but the `ResizeObserver` still observes the text element itself. Adding `pr-11` to the parent shrinks the observed child, so the observer re-fires:

1. Measure at width `W`, overflowing, `setIsOverflowing(true)`
2. Parent gains `pr-11`, child is now `W - 44`
3. Observer fires, full re-measure at `W - 44`
4. Same result, React bails, stable

It does converge (narrower can only mean more overflow, never less), so it isn't a hang. But every overflowing element pays an extra full measure: clone, append, `getBoundingClientRect()` (forced sync reflow), remove, on mount and on every resize. An entity page with a long name and a long description pays it twice. It's also the exact shape that produces `ResizeObserver loop completed with undelivered notifications`, which is what the removed code's comment was guarding against: "so we don't mutate layout-affecting classes on the visible (and observed) element."

The fix should be small: observe the wrapper `<div>` rather than the text element. Its outer width doesn't change when you add padding to it, so the feedback edge goes away, while `isEllipsisActive` still measures the child at its true narrowed `clientWidth`, so you keep the accuracy win.

**Two latent bugs in the new shared API.** Neither fires today since both call sites pass `maxLines={3}`, but worth guarding while the component is new:

- Line 55: the `maxLines === 1` branch never sets `temp.style.width`, and the clone is `position: fixed`, so any width class carried over resolves against the viewport instead of the container. Since the header already passes `textClassName="w-full"`, a future `<ClampedText maxLines={1} textClassName="w-full" />` makes `full.width > e.clientWidth + 1` always true, giving a permanent "More" on text that fits. The `maxLines > 1` path only escapes this because it sets an explicit inline width at line 59.
- Line 125: `LINE_CLAMP_CLASS[maxLines]` is `undefined` outside 1 to 6. The text renders unclamped and the clone measures the same unclamped height, so no toggle renders either and the full text just spills. The `Truncate` component this replaces threw `Maximum lines is currently 6.`, which seems worth keeping loud.

**Minor:**

- Line 97: `setExpanded(false)` runs in a passive effect while the measure effect bails when `expanded` is true, so when sync swaps in a shorter name mid-expand you get at least one painted frame of short text with a trailing "Less".
- Line 58: the clone forces `white-space: normal`, so a caller passing `whitespace-pre-line` or `pre-wrap` (plausible for descriptions with newlines) measures short and never gets a toggle.

**Question:** in `entity-page-body.tsx:158`, `sidePanelHeadingClassName` (line 27) force-disables `line-clamp-*` on descendants (`display:block!important; overflow:visible!important`, and I confirmed it's emitted in the built CSS). So in the side panel the `h1` is never clamped, `isOverflowing` stays false, and the More/Less toggle never appears there. Nothing is hidden from the reader, but is that intended or an oversight?

b-d055 (commented):
@webmagic123  see comments

apps/web/design-system/clamped-text.tsx
webmagic123: @b-d055 Fixed.
Answer: It's an oversight.
sidePanelHeadingClassName has been added in June.
EditableHeading only switched to ClampedText on Aug 18th.
Removed the constant and the className from the wrapper at entity-page-body.tsx:158. The side panel now clamps to 3 lines with a working More/Less, same as the main page.

b-d055 (approved):

PR #2156, opened 2026-08-13
feat(explore): add the "Best" sort, backed by the Phase A ranking

ohohoreilly:
Adds **Best** as the default Explore sort, served by `entitiesRankedForFeedConnection` (gaia #892, deployed). Best ranks on quality, structural completeness and recency computed server-side. `New` stays reverse-chronological, `Top` stays the integer score property.

Requires gaia #892 — merged and deployed to the v2 testnet, verified below.

## Best sends no `filter`, deliberately

The other two sorts build one with `buildFeedFilter`. Every clause of it is now enforced inside `entities_ranked_for_feed`:

| clause | enforced by |
|---|---|
| name presence | gaia 0075 |
| system entities | 0076, via the unforgeable System Type relation |
| data / text / ranking blocks | `entity_type_exclusions` config |
| space and type scoping | 0077, as function arguments |
| recency window | `createdAfter`, the function's own argument |

Sending them again would be redundant and **not free**: `filter` together with `totalCount` and `edges` on this connection exceeds the statement timeout — totalCount scans the filtered candidate set while edges walks it again. Omitting both keeps Best on an index-ordered walk that stops as soon as `first` rows are found.

Because re-adding either one looks like an improvement, tests pin their absence.

`requireName` is not honoured for Best: a nameless entity is never a candidate server-side and cannot be opted back in. Nothing passes `requireName: false` today, and for this feed it would be a request to serve rows that render as a raw uuid.

## No codegen

The explore documents are hand-built GraphQL parsed at runtime and typed `TypedDocumentNode<any, any>`, so nothing here depends on `core/gql` being regenerated. Worth stating because it looked like a prerequisite.

## Shared card selection extracted

The per-entity selection was duplicated **byte-identically** across the New and Top documents — differing only in fragment name, since they're parsed separately and a shared name would collide — and Best would have been a third copy of ~120 lines. A field added for one sort would then render a subtly different card on the others, which the shared decoder can't detect.

The extraction is behaviour-preserving: the **printed ASTs of both existing documents are identical before and after** (185 lines each), and a test asserts all three documents select the same field set.

## Verification

Against the deployed API, using the real document and `EXPLORE_ENTITY_TYPE_IDS`:

| scoping | rows |
|---|---|
| none | 5 |
| `spaceIds` = spaces the top entities live in | 5 |
| + `typeIds` = their types | 5 |
| + `createdAfter` 7d | 5 |
| `typeIds` = full EXPLORE set | 5 |

Scoping to six arbitrary spaces correctly returns zero — those spaces hold no ranked content.

21 tests pass. The 7 new document tests are mutation-tested:

| mutation | caught by |
|---|---|
| add a `$filter` variable | sends no `filter` |
| Best selects a field the others don't | all three select an identical field set |
| Best points at the wrong connection | each sort targets its own connection |

## Note for reviewers

The top-ranked entities currently all live in **one space**, and `Tweet` occupies ranks 2–4 — a type that has never been triaged in the feed config. Neither is caused by this PR (both fell out of removing the type boosts), but Best makes them the first thing a logged-out user sees. The per-creator diversity cap is Phase C.

### DIFF
--- a/apps/web/app/api/explore/feed/route.ts
+++ b/apps/web/app/api/explore/feed/route.ts
@@ -14,12 +14,12 @@ function normId(id: string): string {
   return id.replace(/-/g, '').toLowerCase();
 }
 
-const SORTS: ExploreSort[] = ['new', 'top'];
+const SORTS: ExploreSort[] = ['new', 'top', 'best'];
 const TIMES: ExploreTime[] = ['today', 'week', 'month', 'year', 'all'];
 
 function parseSort(raw: string | null): ExploreSort {
   if (raw && (SORTS as string[]).includes(raw)) return raw as ExploreSort;
-  return 'top';
+  return 'best';
 }
 
 function parseTime(raw: string | null): ExploreTime {
--- a/apps/web/core/explore/explore-best-document.ts
+++ b/apps/web/core/explore/explore-best-document.ts
@@ -1,0 +1,63 @@
+import type { TypedDocumentNode } from '@graphql-typed-document-node/core';
+
+import { parse } from 'graphql';
+
+import { exploreCardNodeFields, exploreCardPropertyFragment } from './explore-card-selection';
+
+const FRAGMENT = 'ExploreBestFragment';
+
+/**
+ * The "Best" sort — Phase A ranked feed, backed by `entities_ranked_for_feed`.
+ *
+ * Deliberately passes no `filter`. The other two sorts build one with
+ * `buildFeedFilter`, but every clause of it is enforced inside the function now:
+ *
+ *   * name presence            -> migration 0075
+ *   * system entities          -> 0076, keyed on the unforgeable System Type relation
+ *   * data/text/ranking blocks -> `entity_type_exclusions` config
+ *   * space and type scoping   -> 0077, as function arguments rather than filter clauses
+ *   * the recency window       -> `createdAfter`, the function's own argument
+ *
+ * Sending them again would be redundant, and not free: `filter` combined with
+ * `totalCount` and `edges` on this connection exceeds the statement timeout, because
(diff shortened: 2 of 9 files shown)


ohohoreilly:
Rebased onto `master` — this had gone 13 commits behind since 2026-08-13 and was sitting `BEHIND`. Clean rebase, no conflicts, CI re-running on the new head.

## Checked against #2171 before pushing

`master` picked up **"show claims on Explore by default" (#2171)** while this sat, which adds `Claim` to the Explore type filter. No files overlap, but a file-level check wouldn't catch the interesting question: this PR deliberately sends **no `filter`** and relies on scoping being enforced server-side, so if the ranking function couldn't take a type filter, making Best the default would have silently broken the new Claim filter for the default view.

It's fine. Migration `0077_feed_space_and_type_scoping.sql` adds `space_ids` and `type_ids` as function arguments, and the **live** API exposes them:

```
entitiesRankedForFeedConnection -> [minRankingScore, createdAfter, createdBefore, spaceIds, typeIds, ...]
```

So Best passes the type filter as a function argument and Claim works with it. (Migration 0076 is the one without those parameters — worth noting if anyone else reads the SQL and reaches for the wrong file, as I did first.)

## Dependency is deployed, not just merged

gaia#892 is merged **and live** — confirmed by introspecting the running API schema rather than the merge state, since the v2 cluster is `workflow_dispatch`-only and merging to `staging` deploys nothing there.

## State

Green before the rebase (Build/Test/Lint/Vercel), no review requested in five days, only the Vercel bot comment. Looks forgotten rather than blocked — it was opened at 21:59 the same evening gaia#892 merged at 21:49.

@jwalkingjew flagging you since this makes **Best** the default Explore sort and it lands on top of your claims work.

PR #2178, opened 2026-08-20
Restore the page that opened a debate flow

jwalkingjew:
## Summary
- remember the exact page that launched a debate flow, including query parameters and hashes
- preserve that destination across recording and debate-again/rematch transitions
- keep unrelated browser tabs out of shared debate-again room transitions
- return forced exits, cancellations, connection timeouts, and room-ownership conflicts to the captured page while retaining safe legacy fallbacks

## Validation
- 258 focused debate, rematch, and room-ownership tests pass
- TypeScript (`bunx tsc --noEmit`) passes
- ESLint passes with no errors
- Full Next build reached production compilation but was stopped after the documented Turbopack workspace-root stall

### DIFF
--- a/apps/web/app/entry.tsx
+++ b/apps/web/app/entry.tsx
@@ -28,6 +28,7 @@ import { FeatureFlagsDialog } from '~/partials/feature-flags/feature-flags-dialo
 import { GovernanceReopenEditLoadingBar } from '~/partials/governance/governance-reopen-edit-loading-bar';
 import { Main } from '~/partials/main';
 import { Navbar } from '~/partials/navbar/navbar';
+import { PendingActionsRunner } from '~/partials/pending-actions-runner';
 import { FlowBar } from '~/partials/review/flow-bar';
 import { StatusBar } from '~/partials/review/status-bar';
 import { SearchDialog } from '~/partials/search';
@@ -136,6 +137,7 @@ export function App({ children }: { children: React.ReactNode }) {
         <ClientOnly>
           <OnboardingDialog />
           <PendingPersonalSpaceRunner />
+          <PendingActionsRunner />
           <CreateSpaceDialog />
           <PendingCreatedSpaceRunner />
           <PendingCreatedSpaceStatus />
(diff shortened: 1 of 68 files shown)


ohohoreilly:
Preston — I owe you an apology, and it's a bit absurd given you handed me the ticket list: I picked GEO-2605 off it about a day after you opened this PR, didn't check for existing PRs, built my own narrower version, and merged it as #2186. That broke this PR (10 conflicts) and shipped the worse fix.

**#2194 backs mine out and lands yours.** Yours is better on every point they overlap:

- **`/space/{id}/debates` as an origin.** Mine used `/\/debates(\/|$|\?)/`, which matches the hub — so entering a debate from the hub recorded nothing and fell back to `router.back()`. That's the most common entry point and exactly where your "weird screen" report comes from, so #2186 didn't fix the main case. Yours handles it.
- Hash preserved; mine dropped it.
- `safeInternalHref` validation; mine had none.
- A 6-hour max age. I'd documented staleness as a won't-fix — expiring it is just better.
- `leaveConflictingRoom` and the rematch/recording transitions, neither of which I touched. The "already open in another tab" screen shows up in your GEO-2604 recording, so that one's earning its keep.

Your coordinator change also supersedes my #2187: I narrowed the guard so it wouldn't push into a *finished* debate, you removed the source-room push entirely because only the tab already in the room owns that transition. That kills the whole loop class plus the multi-tab bug. I verified the premise rather than assuming it — `finishLiveDebate` does compute `rematchDestination(session)` and replace there itself, so the coordinator's push was redundant.

Two changes I made to your work in #2194:

1. **Rebased onto master** — this branch predates #2183, so `useClaimPickerPage` became `useClaimEntitiesByIds`; the other two conflicts were import unions.
2. **Added a test for the `//host` guard**, which mutation-testing showed nothing covered. Being precise about severity: it isn't a live open redirect — `new URL('//evil.com/x', base)` reconstructs to the bare path `/x`, so without the guard you'd land on the wrong *local* page. The test pins reject-over-rewrite.

One gap worth its own PR, which your change widens rather than causes: with the coordinator no longer routing for a rematch that has a source debate, nothing surfaces an active rematch outside the room tab. `hasActiveDebateFlow()` greys out every Debate control, the rejoin bar is suppressed while `activity.rematch` is set, and no push happens — so losing the room tab mid-rematch strands you until it expires.

Happy to close this in favour of #2194, or you can rebase this one yourself and I'll close mine — it's your design either way.

PR #2182, opened 2026-08-20
fix(debates): stop the ready prompt flickering over the rematch handoff

jwalkingjew:
## The bug

Requester side of the rematch flow: send a claim request from the picker, the opponent accepts, and "Your debate is ready" flashes on screen for a beat before the app redirects to the get-ready screen. No interaction, nothing to decide — it just appears and leaves.

## Why

Accepting fires one `debate.rematch_changed`. The gateway turns that into **two separate refetches** — the account's activity and the rematch session — and either can land first.

The picker only redirects once *its* session query reports `converted`. So when activity landed first, there was a window where the requester had a new `ready` debate in activity, the rematch session already cleared from it, and a picker that had not yet learned to redirect. `DebateCoordinator` reads exactly that shape as "this person has not been told their debate is ready" and opens the dialog — which the picker then navigates out from under them.

The accepting tab was already covered. Its mutation has the debate in hand, so it calls `markEnteringDebate` before pushing ([hooks.ts:634](apps/web/core/debates/hooks.ts)). The requester has no mutation to hang that on. All they have is the page they are already sitting on.

## The fix

The coordinator now treats the rematch picker the way it already treats the debate room: a place that owns its own way in and must not be interrupted.

The rejoin bar is suppressed there too. Letting it through instead would be the same flicker in a smaller hat — and the existing accepting-tab test already guards that pairing, so it seemed worth matching.

Scoped to being *on* the picker, not to having been in a rematch at all: someone who walked away still has nothing else to tell them, which is the entire point of the prompt (GEO-2514).

## Testing

`debate-coordinator.test.tsx` has the right harness for this, but **none of it was running**. `pending-personal-space` reads localStorage at module scope (`atomWithStorage` with `getOnInit`), and the storage jsdom hands back here has no `getItem` — the throw took the file down during collection, so all 40 tests were dead on master and in CI alike. The coordinator reaches that module through `SpaceChip`.

Mocked locally so the suite runs. That is 40 pre-existing tests brought back, all green, plus two new ones:

- the requester is not prompted (nor offered the rejoin bar) while the picker is handing them off
- they *are* prompted once they have left the picker, so the guard hasn't swallowed the case the prompt exists for

I checked the new test fails without the fix rather than trusting it passes with it.

- `bun run build` passes.
- Full `vitest run`: 21 failed files / 39 failed tests vs. a baseline of 22 / 39 — one fewer broken file, same failures, 42 more tests passing.

## Not covered, and one thing worth a look

I traced this from the code and the gateway's invalidation path, not from a live repro — it needs two signed-in accounts running a rematch to completion. The reasoning rests on activity dropping `rematch` at conversion, which is what the reported symptom implies (the existing `!activity?.rematch` guard would otherwise have suppressed the dialog already). Worth a confirm on a real run before this leaves draft.

The root localStorage problem is worth its own PR: the same throw is why ~20 other test files never collect, and fixing `pending-personal-space` to tolerate a storage without `getItem` would bring all of them back at once. Out of scope here, but happy to pick it up.

### DIFF
--- a/apps/web/core/debates/debate-coordinator.test.tsx
+++ b/apps/web/core/debates/debate-coordinator.test.tsx
@@ -105,6 +105,19 @@ vi.mock('./claim-response-indexed-notifier', () => ({
 
 vi.mock('~/core/state/feature-flags', () => ({}));
 
+// `pending-personal-space` reads localStorage at module scope (`atomWithStorage` with
+// `getOnInit`), and the storage jsdom hands back here has no `getItem`. That throws while the
+// module graph is still being built, which took the whole suite down at collection — every test
+// in this file, on master and in CI alike. The coordinator reaches it through `SpaceChip`.
+vi.mock('~/core/state/pending-personal-space', () => ({
+  PENDING_PERSONAL_SPACE_PREFIX: 'pending:',
+  pendingPersonalSpaceAtom: { toString: () => 'pendingPersonalSpaceAtom' },
+  pendingPersonalSpaceId: (topicId: string) => `pending:${topicId}`,
+  isPendingPersonalSpaceId: (spaceId: string | null | undefined) =>
+    typeof spaceId === 'string' && spaceId.startsWith('pending:'),
+  usePendingPersonalSpace: () => ({ isPending: false }),
+}));
+
 beforeEach(() => {
   sessionStorage.clear();
   mocks.push.mockReset();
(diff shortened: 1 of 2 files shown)


ohohoreilly:
Preston — same apology as on #2178: I built and merged #2189 for this bug a day after you opened this, without checking for an existing PR. **Your fix and mine are the same fix** — suppress the coordinator's prompt *and* the rejoin bar while the viewer is on the picker path — so #2189 already covers it and this is superseded rather than lost.

Your account of the mechanism is better than mine and I'd like to fold it into the code comment: one `debate.rematch_changed` becomes two refetches (activity and the rematch session) and either can land first. I'd written it as "activity reports the converted debate before the session query catches up", which is the same race described one level shallower. Yours says *why* they can land out of order.

**One claim doesn't reproduce, and it matters because it would send you after the wrong thing.** You wrote that `debate-coordinator.test.tsx` never collects — "all 40 tests were dead on master and in CI alike" — because `pending-personal-space` reads localStorage at module scope, and offered a follow-up to bring ~20 files back that way.

On current master it collects fine: **44 tests pass** (your 40, +2 from #2187, +2 from #2189).

I think your "22 failed files / 39 failed tests" baseline is an **unbuilt `@geogenesis/auth` workspace package**, not localStorage. I hit exactly that tonight: 8 phantom `tsc` errors and 7 phantom test failures, all from `packages/auth/dist` being a month stale while `src` exported the symbol. `bun turbo run build --filter=@geogenesis/auth` cleared every one. CI runs `turbo run test` which `dependsOn: ["build"]`, so it builds first — meaning those tests were never dead in CI, and the localStorage follow-up would be chasing a problem that isn't there.

Worth confirming on your machine before you spend time on it. If you build that package and still see the collection failure, I'm wrong and I'd want to know.

ohohoreilly:
Closing as superseded — #2189 shipped the same fix (suppress the coordinator's prompt and the rejoin bar while on the picker path) and it's deployed. Reopen freely if you disagree; the reasoning is in my comment above.

Your description of the mechanism is the better one and I don't want it lost: one `debate.rematch_changed` becomes two refetches and either can land first. I'll fold that into the code comment on master.

Also worth re-checking before you spend time on the test-collection follow-up — `debate-coordinator.test.tsx` collects fine on master (44 passing), so the localStorage theory doesn't hold here; `bun turbo run build --filter=@geogenesis/auth` cleared the same phantom failures for me.

PR #2189, opened 2026-08-20
fix(debates): keep app-wide debate prompts off the rematch page

ohohoreilly:
Addresses the **second half** of GEO-2604. The first half needs a product call — see the bottom.

## The reported bug

> I sent someone else a request. They accepted. Then this popup came up for a second, required no interaction from me, and then without me clicking redirected me to the I'm ready screen. This popup should not appear here.

The screenshot is `DebateReadyPrompt`.

## Why it appeared

The redirect itself is correct — `rematch-page-client.tsx` walks the requester into the converted debate:

```js
if (session.status === 'converted' && session.converted_debate_id) {
  markEnteringDebate(session.converted_debate_id);
  router.replace(`/space/${session.source_space_id}/debates/${session.converted_debate_id}`);
}
```

The popup is the gap in front of it. When the session converts, **activity** gains the new `ready` debate and drops the rematch — but the page's **session query** hasn't caught up, so `markEnteringDebate` hasn't run and the page hasn't navigated. For that beat the coordinator sees:

- a `ready` debate, ✅
- describable (both participants), ✅
- `!atDebate` — pathname is still the rematches page and no entry intent is held, ✅
- `!activity?.rematch` — already gone, ✅

which is precisely the "this person has not been told their debate is ready" shape. So it opened the dialog on top of a page that was already navigating, and unmounted it when the navigation landed.

## The fix

The rematch page owns routing for its own session, and what it is about to do is more current than activity is — so nothing app-wide belongs over it:

```js
const atRematchPage = pathname.includes('/debates/rematches/');
```

Applied to the ready prompt and to the rejoin bar, which would flash in the same window.

This is the same principle as the existing test *"does not route stale debate activity over an active rematch page"* — that guard covered the coordinator's navigation but not its dialogs.

## Verification

- 57 debate suites / 827 tests pass.
- `tsc --noEmit`: 8 errors, identical to the `master` baseline (all pre-existing, unrelated files).
- eslint and prettier clean.
- **Mutation-tested:**
  | mutation | killed by |
  |---|---|
  | drop the guard from the prompt | `does not prompt over the rematch page for a debate that page is about to open` |
  | drop the guard from the rejoin bar | same test's second assertion |
  | `atRematchPage = true` (suppress everywhere) | 8 tests, including the new `still prompts for a ready debate away from the rematch page` |

  The third mutation matters: a guard that suppresses too much would have looked like a pass without it.

## The first half of the ticket

> When I am entering a debate there are various loading indicators that dont need to be there.

I stepped through that `.mov` frame by frame. It is not a stray indicator — it is real waiting, in three stages:

| what is on screen | roughly |
|---|---|
| `Opening debate-again choices…` (the route's `loading.tsx`) | 2s |
| grey skeleton rows under `Recommended` | ~10s |
| `Connecting` / `Waiting for video` placeholders in the room | several seconds |

The 10-second skeleton is the picker's own comment describing itself: *"The opponent's claims arrive in one round trip; the curated lookup is three, in sequence."* Removing the indicators would just show an empty page for the same 10 seconds, so this needs either the serial round trips collapsed or a decision about what to show while they run — same underlying work as GEO-2599. Not folded in here.

The tail of that recording also shows *"This debate is already open in another tab"* repeatedly, which looks like a separate issue and may be related to GEO-2602.

### DIFF
--- a/apps/web/core/debates/debate-coordinator.test.tsx
+++ b/apps/web/core/debates/debate-coordinator.test.tsx
@@ -436,6 +436,42 @@ describe('DebateCoordinator', () => {
     expect(screen.queryByRole('button', { name: /Your debate is/ })).not.toBeInTheDocument();
   });
 
+  // GEO-2604. The window this closes: the rematch session has converted, so activity reports the
+  // new `ready` debate and no longer reports a rematch, but the page's own session query has not
+  // caught up and so has not navigated yet. That is exactly the shape the ready prompt exists for —
+  // a ready debate the viewer has not been told about — so it opened over a page that was already
+  // on its way into the room, then vanished when the navigation landed. Preston reported a popup
+  // that "required no interaction" and redirected him anyway.
+  it('does not prompt over the rematch page for a debate that page is about to open', async () => {
+    mocks.pathname = '/space/space-1/debates/rematches/rematch-1';
+    mocks.activity = {
+      ...activityWithDebate(),
+      rematch: null,
+      debate: { ...activityWithDebate().debate!, status: 'ready', participants: bothParticipants() },
+    };
+
+    render(<DebateCoordinator />);
+
+    await waitFor(() => expect(mocks.push).not.toHaveBeenCalled());
+    expect(screen.queryByText('Your debate is ready')).not.toBeInTheDocument();
+    // Nor the fallback bar: it would flash in exactly the same window.
+    expect(screen.queryByRole('button', { name: /Your debate is/ })).not.toBeInTheDocument();
+  });
+
+  // Still prompted anywhere else, so suppressing it above cannot swallow a real one.
+  it('still prompts for a ready debate away from the rematch page', async () => {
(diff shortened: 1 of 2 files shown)


ohohoreilly:
**Integration check** — merged all four of #2186, #2187, #2188, #2189 onto `master` in a scratch branch: 57 debate suites / **843 tests pass**, `tsc` still at the 8-error baseline. Only textual conflicts in `debate-coordinator.tsx`/`.test.tsx`, auto-resolved.

**One caveat worth naming**, since this guard hides the rejoin bar too: a viewer sitting on a rematch page who *also* has a separate live `ready` debate now has neither way in while on that page. In practice `activeFlow` means those two shouldn't coexist, the `deciding` routing effect walks them into a live source debate anyway, and navigating off the full-screen picker restores the bar — so I left it rather than adding machinery for it. Flagging it in case you'd rather the bar stayed.

PR #2194, opened 2026-08-21
Adopt #2178's return-navigation over my narrower #2186/#2187

ohohoreilly:
Supersedes #2178 (adopting its work) and reverts my #2186 and #2187.

**I duplicated work.** Preston opened #2178 on 08-20 00:41 and #2182 at 01:30. I then picked GEO-2605 and GEO-2604 off the list he handed me, about a day later, without checking for open PRs — built my own, and merged them — which broke both of his (10 and 2 conflicts). Preston's #2178 is better than what I shipped, so this backs mine out and lands his.

## Where his is better

| | mine (#2186) | his (#2178) |
|---|---|---|
| `/space/{id}/debates` as an origin | **treated as in-flow → not recorded** | recorded |
| URL hash | dropped | preserved |
| destination validation | none | `safeInternalHref`, rejects `//host` |
| stale origin | documented as won't-fix | 6-hour max age |
| room-ownership conflict exit | not handled | `leaveConflictingRoom` |
| rematch/recording transitions | not handled | destination survives them |

The first row is a real bug in what I merged. My predicate was `/\/debates(\/|$|\?)/`, which matches the debates **hub** — so entering a debate from the hub recorded no origin at all and fell straight back to `router.back()`. The hub is the most common entry point and precisely where Preston's "weird screen" report lives, so **#2186 did not fix the main case.** Verified both predicates side by side against real paths.

His coordinator change also supersedes my #2187. I narrowed the guard so the coordinator wouldn't push into a *finished* debate; he removes the source-room push entirely, on the grounds that only the tab already in the room owns that transition. That kills the whole loop class *and* a multi-tab bug — routing from an app-wide coordinator sent every other open tab into a room that ownership then rejected.

I checked that claim rather than taking it: `finishLiveDebate` in the room does compute `rematchDestination(session)` and `router.replace`es there itself, so the coordinator's push was redundant. ✅

## What I changed in his work

- **Rebased onto current master.** His branch predates #2183, so its `useClaimPickerPage` import had to become `useClaimEntitiesByIds`; two other conflicts were import-block unions.
- **Added a test for the `//host` guard**, which mutation-testing showed was unverified. Worth being precise about severity: this is *not* a live open redirect — `new URL('//evil.com/x', base)` reconstructs to the bare path `/x`, so without the guard you'd be sent to the wrong **local** page, not off-origin. The test pins reject-over-rewrite.

## One claim of his that doesn't reproduce

#2182 says `debate-coordinator.test.tsx` doesn't collect — "all 40 tests were dead on master and in CI alike" — because `pending-personal-space` reads localStorage at module scope, and offers a follow-up to bring ~20 files back.

It collects fine on master: **44 tests pass** (40 original, +2 from #2187, +2 from #2189). His full-suite baseline of "22 failed files" is almost certainly an **unbuilt `@geogenesis/auth` workspace package** — I hit the same thing tonight, saw 8 phantom `tsc` errors and 7 phantom test failures from it, and `bun turbo run build --filter=@geogenesis/auth` cleared all of them. CI builds before testing, so those tests were never dead there. The offered follow-up would chase the wrong cause.

## Kept

#2189 stays — it's the same fix as his #2182 (suppress the coordinator's prompt and rejoin bar on the picker path). His description of the mechanism is better than mine, though: one `debate.rematch_changed` becomes two refetches and either can land first. Same race, more precise account.

## Known gap, not introduced here but widened

With the coordinator no longer routing for a rematch that has a source debate, nothing surfaces an active rematch outside the room tab — `hasActiveDebateFlow()` greys out every Debate control, the rejoin bar is suppressed while `activity.rematch` is set, and no push happens. Lose the room tab mid-rematch and you're stranded until it expires. Previously the coordinator would at least push you *somewhere* (into the loop). Worth its own fix; flagging rather than folding in.

## Verification

- 58 debate suites / **836 tests** pass.
- `tsc --noEmit`: 1 pre-existing error (`use-entity-vote.ts`), same on master.
- eslint and prettier clean.
- **Mutation-tested his three guards:** treating the hub as in-flow, accepting `//host`, and never expiring each fail a test.

### DIFF
--- a/apps/web/app/space/[id]/(space)/debates/[debateId]/debate-room-page-client.test.tsx
+++ b/apps/web/app/space/[id]/(space)/debates/[debateId]/debate-room-page-client.test.tsx
@@ -1398,7 +1398,8 @@ describe('DebateRoomPageClient', () => {
       await screen.findByRole('heading', { name: 'This debate is already open in another tab.' })
     ).toBeInTheDocument();
     expect(screen.getByText('Close this tab and continue your debate in the original tab.')).toBeInTheDocument();
-    expect(screen.getByRole('link', { name: 'Go to debates' })).toHaveAttribute('href', '/space/space-1/debates');
+    fireEvent.click(screen.getByRole('button', { name: 'Go back' }));
+    expect(mocks.replace).toHaveBeenCalledWith('/space/space-1/debates');
     expect(screen.queryByRole('button', { name: 'Continue here' })).not.toBeInTheDocument();
     expect(screen.queryByRole('button', { name: 'Back to debates' })).not.toBeInTheDocument();
     expect(screen.queryByText('Debate room')).not.toBeInTheDocument();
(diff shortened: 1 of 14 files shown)


ohohoreilly:
Redundant — #2178 merged as `ad7cb8f32` with the same content, attributed to Preston, which is the right outcome. This PR only existed because I'd built a duplicate; the reverts, the #2183 rebase and the `//host` guard test all went onto his branch instead.

Branch left in place for a day in case anything needs referencing, then deleted.

PR #2200, opened 2026-08-21
fix: let a viewer change who they voted for as winner

webmagic123:
Feature:
change your debate winner vote after casting it.

Problems:
Publishing a second vote leaves two on the tally, which reads them back
in no fixed order — so a change could silently revert.
A vote could land invisible to the tally, permanently.
A failed vote could delete a newer one.
A per-component flag left one surface's buttons live while the other
was publishing.

Fixed:
1. Edit the existing vote instead of publishing a new one.
2. Attach the two relations that make a vote findable, on every path.
3. Roll back only your own row on failure.
4. Share the in-flight flag across surfaces, so both disable together.


apps/web/core/debates/use-debate-votes.tsx
```diff
@@ -198,22 +205,38 @@ export function useDebateVotes(debate: Debate): DebateVotesResult {
...
+                id: previousWinnerRelationId,
+                isDeleted: true,
+              },
+              newWinnerRelation,
+            ]
+          : [
+              relate(TYPES_PROPERTY_ID, 'Types', VOTE_TYPE_ID, 'Vote'),
+              relate(VOTE_DEBATES_PROPERTY_ID, 'Debates', debateEntityId, debateName),
+              newWinnerRelation,
+            ];
```
b-d055: This branch publishes only `[delete old winner, create new winner]`, assuming the Vote entity already carries `Types` and `Debates`. That holds for a `previousVote` parsed from the indexer, and for the optimistic row when the first publish succeeded — the case the comment just below covers, and what the "switching twice before the indexer catches up" test exercises.

It does not hold when the first publish fails while a second vote is already in flight from the other mounted instance of this hook (see the comment on `:161`):

1. Vote 1 writes the optimistic row — `winnerRelationId` non-null by construction (`:233`) — into `votesQueryKey(debateEntityId)`, the same cache entry `previousVote` reads from (`:128`, `:163`). It then awaits its receipt, up to `RECEIPT_DEADLINE_MS` (90s, `use-smart-account.ts:27`).
2. Vote 2 fires from the other surface, reads that row as `previousVote`, reuses `voteEntityId` (`:179`), and takes this branch — no `Types`, no `Debates`.
3. Vote 1 fails; its rollback (`:278-281`) clears the row. Vote 2 then succeeds, creating the entity on-chain with a Name value and a winner relation only.
4. `getDebateVoteEntities` filters `Debates` backlinks to the Vote type, so it never sees the entity. The poll at `:304-308` runs all 45 attempts, invalidates, and the pills fall back to "Winner?".

Recoverable: `previousVote` is null afterwards, so the next vote mints a fresh entity id and publishes the full relation set, and `tallyDebateVotes` collapses per voter space, so counts are unaffected. The cost is a lost second pick and an orphan entity in the voter's personal space.

Suggestion: emit `Types` / `Debates` with deterministic ids (as `ID.createValueId` does for the Name value at `:185`) so a duplicate emit is a no-op and this branch cannot produce an orphan regardless of how it is reached.

apps/web/core/debates/use-debate-votes.tsx
```diff
@@ -156,7 +158,11 @@ export function useDebateVotes(debate: Debate): DebateVotesResult {
 
   const castVote = React.useCallback(
     async (participant: DebateParticipant) => {
-      if (hasVoted || isVoting) return;
+      if (isVoting) return;
```
b-d055: `isVoting` is `React.useState` inside the hook (`:121`), so every caller gets its own copy and this guard is per-instance.

Two instances are mounted at once for the same debate: `debate-claims-panel.tsx:23` and `debate-feed.tsx:307`. Both render in the same tree (`debate-feed.tsx:262` for `sidePanel`, `:273` for `feed`) — per the comment at `:269-270`, the feed stays mounted so toggling the panel doesn't restart playback. Both pass `disabled={votes.isVoting}` to `WinnerVoteButton`, so publishing from the player leaves the panel's pills enabled, and vice versa.

Pre-existing for a first vote. New in this PR for a switch, since `winner-vote-button.tsx:60-77` makes the non-picked post-vote pill a `<button>`.

Scope: the sends themselves are not concurrent. `smartAccount.sendUserOperation` is the wrapper at `use-smart-account.ts:88-97`, which serializes per-EOA via `enqueueFor` and awaits `confirmInclusion`, so a sequential A→B→C switch chain converges correctly on-chain. The consequence is the orphan path noted on `:210-224`.

Suggestion: derive `isVoting` from `useIsMutating` on a mutation key, or from a module-level in-flight set keyed by `debateEntityId`, instead of local `useState`.

apps/web/core/debates/use-debate-votes.tsx
```diff
@@ -118,6 +119,7 @@ export function useDebateVotes(debate: Debate): DebateVotesResult {
   const [, setToast] = useToast();
   const reportError = useReportError();
   const [isVoting, setIsVoting] = React.useState(false);
+  const pollGenerationRef = React.useRef(0);
```
b-d055, line 154: `pollGenerationRef` is per-instance for the same reason as `isVoting`, so the claims panel's poll is not cancelled by a vote cast from the player and can fire `invalidateQueries` (`:322`) against the other surface's newer optimistic pick. Folds into whatever shared-state fix lands for `:161`.

b-d055 (commented):
@webmagic123 Two findings on the switch path, plus one related note. Details inline.

webmagic123:
- Feature: change your debate winner vote after casting it.
- Problems:
1. The tally reads them back in no fixed order and vote change can revert.
2. A vote can land invisible to the tally
3. A failed vote deletes a newer one.
4. A per-component flag left one side's buttons live.
- Fixed:
1. Edited the existing vote instead of publishing a new one.
2. Attached the two relations that make a vote findable, on every path.
3. Modified to revert the user's row in case of an error.
4. Fixed an issue where only one button was enabled due to component-specific flags.

b-d055 (approved):

PR #2203, opened 2026-08-24
fix(debates): stop the rematch picker emptying its curated and opponent tabs

ohohoreilly:
Fixes [GEO-2637](https://linear.app/geobrowser/issue/GEO-2637/debate-again-flow-bugs). Three reports off the rematch picker; two of them turned out to be the same bug.

## The allowlist was applied to every tab (reports 1 and 2)

`useClaimSpaceAllowlist` is *featured spaces + the spaces the viewer belongs to*. It exists for the **All** tab, which browses the whole published corpus and would otherwise offer claims from spaces the viewer has nothing to do with. It was applied to the whole pool instead — and a debater's claims live in their **personal space**, which nobody else is a member of.

Checked against the curator page in the report (`123ed70f…`):

- all **136** curated claims live in `8a4955bc…`, Adam's personal space — `type: PERSONAL`, zero members, zero editors
- so it enters the allowlist only via `if (personalSpaceId) allowed.add(...)`

That makes it Adam's space and no one else's. The Recommended tab drew for him and rendered empty for Dovile — precisely the asymmetry reported. The same filter dropped every position he took on his own claims out of Dovile's *"Adam's positions"* tab and zeroed its count, which is what read as **"the tab isn't updating."**

The allowlist now narrows the All tab only. The other two tabs are each bounded by an explicit source — one person's own responses, or one page published in a curator space this build trusts by id — so neither can fan out the way browsing can, and the viewer's own memberships say nothing about whether the source is worth showing. They also no longer wait on a lookup they don't use.

## The live path had a second hole (report 1)

`debate.claims_changed` is delivered per space, and the picker retained a gateway scope only on the spaces of the tab **in front of the viewer**. Two consequences: switching tabs dropped the scopes the other lists depend on, and the opponent's *first* position could never arrive live — the tab is empty at that point, so there is no claim to derive a scope from.

Now scoped on every list's spaces plus both participants' personal spaces, where a debater's own claims live. The 20-second poll goes back to being the backstop it was meant to be.

Scope retention moved out of `useClaimReadinessByClaimId` (it was incidental there) and up into the page, where the union is available. It only ever *invalidates* keys with `refetchType: 'active'`, so the two extra profile-space scopes cost no requests — the picker mounts neither `['debates','claims',…]` nor `['debates','space',…]` for them.

## The missing section was its own bug (report 3)

The blocks' collection items were fetched as one page capped at 100 ids, with the cap falling across **all** blocks at once rather than per block. Measured against the live entity, in block order:

| Block | Claims | Actually resolved |
|---|---|---|
| AI | 31 | 31 |
| Relationships | 32 | 32 |
| Health | 21 | 21 |
| Politics | 43 | **21** |
| Current Affairs | 14 | **0** |

A block holding no claims is dropped rather than headed as an empty section — hence a curated group that simply never appeared, and Politics silently showing half of what the curator picked.

Routed through the picker's own `useClaimEntitiesByIds`, which batches the ids and asks the graph for claims specifically. That also drops the client-side type filter it makes redundant, and swaps a full-entity fetch for the narrow projection built to avoid exactly this cost.

## Verification

- `tsc` clean on all four files. Three errors remain in `sort-open-proposals.test.ts` and `use-entity-vote.ts` — pre-existing on master, untouched here.
- Lint clean (two warnings in the test file are pre-existing, on untouched lines).
- **858 tests / 60 files** green across `core/debates` and `app/space/[id]/(space)/debates`, re-run after rebasing onto current master.
- Two tests that encoded the old allowlist behaviour are replaced by ones asserting the opposite, with the reasoning; new regression test covers the item-fetch cap.

## Two things for review

1. **Needs a gateway-side check.** I verified the client now holds a scope on the participants' personal spaces, but not whether geo-chat actually *broadcasts* `claims_changed` to a subscriber who isn't a member of that space. If the server gates on membership too, report 1's live path needs a geo-chat change as well and the poll stays the only route. Worth confirming with whoever owns the gateway.
2. **Report 2's symptom is "the tab renders but draws nothing"**, not "the tab is absent" — the tab button isn't allowlist-gated, so it appears either way. If Dovile saw *no tab at all*, there is a third thing going on that this doesn't cover.

### DIFF
--- a/apps/web/app/space/[id]/(space)/debates/rematches/[sessionId]/rematch-page-client.test.tsx
+++ b/apps/web/app/space/[id]/(space)/debates/rematches/[sessionId]/rematch-page-client.test.tsx
@@ -1023,29 +1023,46 @@ describe('DebateRematchPageClient', () => {
     await waitFor(() => expect(screen.queryByText('A newly published claim')).toBeNull());
   });
 
-  // Applied to the pool, not to the All tab alone, so the tab a viewer lands on describes the same
-  // set of claims as every other one.
-  it('drops disallowed claims from the opponent tab too', async () => {
+  // Applied to the All tab alone. The opponent's tab is bounded by one person's own responses, and
+  // a debater's claims live in their personal space, which nobody else is a member of — narrowing
+  // it by the viewer's own memberships emptied the tab, and zeroed its count, for everyone but the
+  // debater who published the claims.
+  it('keeps the opponent’s claims from spaces outside the viewer’s allowed set', () => {
     mocks.claims = [sharedClaim()];
     mocks.spaceAllowlist = new Set([SPACE_2.replace(/-/g, '')]);
     render(<DebateRematchPageClient sessionId="rematch-1" />);
 
-    await waitFor(() => expect(screen.queryByText('A claim both participants chose')).toBeNull());
-    // The count follows the same pool, so it can't advertise a position the tab no longer lists.
+    expect(screen.getByText('A claim both participants chose')).toBeInTheDocument();
     const tab = screen.getByRole('button', { name: /Salina’s positions/ });
-    expect(within(tab).getByText('0')).toBeInTheDocument();
+    expect(within(tab).getByText('1')).toBeInTheDocument();
+  });
+
+  // Same reasoning, and the case that made it visible: a curator's page is trusted by the space it
+  // was published in, so the claims on it are vouched for wherever they live. The other debater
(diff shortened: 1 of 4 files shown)


ohohoreilly:
Closed by a branch rename — reopened as #2204 on `ohohoreilly/geo-2637-debate-again-flow-bugs`. Same commit (`554d24d`), no content change.

PR #2213, opened 2026-08-25
Browse-mode dropdowns for data tables

TheHoneySeller:
## Summary

Any user browsing a data table can now customize their own view through per-property dropdowns, without editing the table's filters for everyone.

- **Config**: a `Dropdowns` relation (`720dec9123044054b044c49a03cee273`) on the block's BLOCKS-relation entity, parallel to `Properties`. Each to-entity is a relation property offered as a dropdown. Set from the edit-mode properties menu ("Browse dropdowns" section).
- **Behavior**: one checkbox menu per property; multiple checks act as **OR**; the table updates on every toggle — **no Done button** (GEO-2592 principle).
- **Personal view**: selections live in localStorage keyed per block relation entity. They never modify the block's filters and never affect other users.
- **Defaulting**: a dropdown whose property is already in the block's filter is preset to the filter's values. Only user overrides are stored; an override that returns to the filter's values (or empties) is dropped, so the dropdown keeps following later filter edits.
- **Scope**: relation properties only. Other property types in the config are ignored.

## Stacked on #2167

Based on `agent/data-block-property-filter-mode` and reviewed as a stack: OR-within-a-property is the same semantics that PR lands, so the overlay reuses its `filterStateToWhere` transformer instead of duplicating one (GEO-2593 cross-check). GitHub will retarget this PR to `master` when #2167 merges.

## How it fits together

```
block filters ──► temporary filters (browse) ──► applyDropdownSelectionsToFilters ──► filterStateToWhere ──► query
                                                 (query only — pills/edit flows never see it)
```

| Layer | Files |
|---|---|
| Config read/write | `use-block-dropdowns.ts`, `block-ontology-ids.ts`, `shown-column-relations.ts`, `diff.ts`, `table-block-properties-menu.tsx` |
| Personal selections | `table-dropdown-selections.ts` (pure, tested), `use-table-dropdown-selections.ts` (localStorage, Explore type-filter pattern, tested) |
| Query overlay | `use-data-block.tsx` (browse mode, after hydration) |
| UI | `table-block-dropdowns.tsx` (`Menu` + `CheckboxVisual`, as Explore/Bounties), `use-dropdown-options.ts`, `table-block.tsx` mount |

## Judgment calls (please push back)

1. Unchecking every option **removes the override** (dropdown snaps back to the table's filter) rather than meaning "match nothing" — same snap-back as the Bounties filters.
2. Options are **target-type based** (entities of the property's relation value types, space-scoped, capped at 100), not derived from the current result set — so a user can widen beyond the block's filter, as "preset … unless the user changes it" implies.
3. Dropdowns render **under the block header, always visible in browse mode**, rather than inside the filter panel, for discoverability.

## Testing

- 20 new tests: pure selection/overlay logic (15) and the persistence hook (5: hydration, no-write-on-hydrate, persist/clear, per-block scoping).
- Full web suite 2,888 / 285 files green; strict `tsc` clean; eslint 0 errors.
- Manual path (needs a block with the new relation — none exist yet): preview → edit mode → properties menu → "Browse dropdowns" → check a relation property → leave edit mode → use the dropdown.

## Follow-ups (not in this PR)

- Chat/agent write path: a `set dropdowns` op parallel to `applySetDataBlockShownColumns`.
- Other property types (boolean/date/integer) per the ticket's future scope.

https://claude.ai/code/session_01KSsRb6kYhhrKJf5ZEsTWp3

---

## Update: collection blocks + per-option match counts (`e60686bb5`)

- **Dropdowns now cover COLLECTION source blocks** (e.g. template-generated profile tabs). Population becomes an explicit union: query blocks walk `entitiesConnection` as before; collection blocks feed their enumerated, Position-ordered item ids through the same page/chunk pipeline, with the block's `where` checked per id-slice via one indexed query. Row filtering reuses `useCollection({ where })` — the gate (`sourceSupportsDropdowns`) was the only change needed there, on the table block and Power Tools alike. RELATIONS blocks stay excluded (selector projections, no population).
- **Every option shows its exact match count**, faceted like the bounty board: the block's filter plus the OTHER dropdowns' selections apply, the property's own constraint removed. Only exact numbers are ever shown — the walk's free tally once the scope is fully read (collections, small tables), otherwise a lazy per-option server `totalCount` for just the revealed options (0.7–4.6s each, a screenful of 25 in ~4s at a 6-way concurrency gate, cached per population+option). No lower bounds; a badge appears when its number is known. Definite zeros render inert unless selected.


TheHoneySeller:
## Scope note: Power Tools included (intent-based extension)

The ticket doesn't name Power Tools, but its premise is user-centric — *"a user browsing a table has no way to adjust what they personally see"* — and Power Tools is that same table expanded: it opens with the block's `relationId` and already reads its column config (`Properties`) off the **same relation entity** the ticket puts `Dropdowns` on. Leaving it out would make a personal view vanish the moment the user expands the table, and #2167 already treats Power Tools as a data-block surface for filter modes.

Power Tools runs its own filter state and data hook, so it needed the overlay applied to its own effective filters (through `usePowerToolsData`'s existing `filterStateOverride` / `modesByColumnOverride`) plus the dropdown row under the browse-mode banner. Selections are shared with the page's table (same per-block storage key). Without a `relationId` there is no relation entity, so it's a clean no-op.

If the narrower reading is preferred, this is one isolated commit to drop.

PR #2256, opened 2026-08-27
fix(explore): cap consecutive same-type items in the Best feed

ohohoreilly:
Closes GEO-2690.

Preston asked to "lower their rank slightly so we get a better mix of content types". The measurement says that lever can't do it, so this does it at read time instead. He approved the approach: *"I'm cool with a diversity cap solution right now."*

## The measurement

Production, Best sort, paging via `nextCursor`:

|  | first screen (22) | 110 items / 5 pages |
| -- | -- | -- |
| **Claim** | **22 — 100%** | 86 — 78.2% |
| News story | 0 | 24 — 21.8% |
| everything else | 0 | **0** |

Longest unbroken claim run: **30**. Tweet and Episode each return 22 items when queried alone, so the content exists — it's outranked, not missing.

## Why the type weight can't fix it

`entity_type_weights` is the intended lever, applied as `ln(weight)`. Claims are **older** than news stories (median 168.6 h vs 118.1 h), so the recency term already favours news by 4.04 score units — and claims still win. The gap on the participation/comment terms is therefore >4 units, needing `weight ≈ 0.018`.

The schema forbids it:

```sql
CONSTRAINT "entity_type_weights_range" CHECK ("weight" > 0.1 AND "weight" <= 10.0)
```

and that floor is deliberate, per the migration's own comment: *"A weight can demote but never effectively exclude."* Maximum demotion is `ln(0.1) ≈ -2.30`, about 29 hours of freshness. The gap is over 50.

## What this does

At most 3 consecutive items share a type; the gap is filled by the highest-ranked item of another type. On the measured mix the first screen goes **22/22 claims → 17/22**. 77% is the floor a run cap alone reaches — the ticket's option 2 (normalising the participation terms, which claims accrue as a byproduct of every debate and news stories never do) is still the principled fix.

Only **Best** reorders. New is reverse-chronological and an activity log that shuffles is wrong; Top is an explicit "rank by score" request.

## Two things worth reviewing closely

**1. Paging, which is most of the diff.** Diversifying a page requires seeing past it, which leaves the over-fetched remainder homeless. Two options rejected:

- *Advance the server cursor past the whole window.* This is what the existing code already did in miniature — scan 30, serve 22, advance 30, **silently dropping 8 items per page**. At a 66-item window it drops 44. (That pre-existing bug is fixed here as a side effect.)
- *Query each type separately and merge.* Preston flagged the failure himself: independent cursors per type make a thin type loop, so *"users may see the same entity over and over"* on reload.

So the window start stays put and the cursor carries an offset into it (`w1:<offset>:<serverCursor>`). `applyDiversityCap` is a pure function of the ranked list, so re-fetching the window reproduces the same order and slicing deeper neither repeats nor skips. **The cost is that a window is fetched once per page it serves.** That's the tradeoff I'd most like a second opinion on — the alternative is returning 66-item pages, which is less total work but triples the first response.

**2. Type classification.** `types` comes from relations whose order isn't meaningful, so the key walks `EXPLORE_ENTITY_TYPES` in declared order. Claim is declared last and loses every tie, which is the useful direction: an entity that is also an Episode breaks a claim run instead of extending one.

## Sizing

3× the page size, from the measurement rather than picked round. A full screen needs `floor(22 / (3+1)) = 5` off-type items; news only starts around rank 31 and runs ~30% of ranks 30–110, so 2× supplies ~4 (short) and 3× supplies ~11. A type filter narrowed to one type keeps the old narrow scan — nothing to diversify against.

An earlier version of this used `ceil` and a test caught it: the last run on a screen has nothing after it to separate from, so it's `floor`.

## Verification

- 3182 tests / 304 files pass; `tsc` shows only the 5 pre-existing errors.
- Cap tested against the measured mix positionally, not as a ratio — the problem is *where* the off-type items sit.
- Both halves mutation-tested: disabling the cap fails 3 tests, dropping the window tail fails 3 others.
- Cursor covers the awkward cases: bare cursors from clients mid-scroll, `hasNextPage` with a null `endCursor` (would otherwise scroll the first screen forever), and a window that comes back shorter than the offset.

### DIFF
--- a/apps/web/core/explore/explore-diversity.test.ts
+++ b/apps/web/core/explore/explore-diversity.test.ts
@@ -1,0 +1,291 @@
+import { describe, expect, it } from 'vitest';
+
+import { CLAIM_TYPE_ID } from '~/core/claims/ontology';
+
+import {
+  EPISODE_TYPE_ID,
+  EXPLORE_PAGE_SIZE,
+  NEWS_STORY_TYPE_ID,
+  TWEET_TYPE_ID,
+} from './explore-constants';
+import {
+  EXPLORE_DIVERSITY_MAX_RUN,
+  EXPLORE_DIVERSITY_WINDOW_SIZE,
+  applyDiversityCap,
+  exploreItemTypeKey,
+  longestTypeRun,
+} from './explore-diversity';
+import {
+  decodeExploreWindowCursor,
+  encodeExploreWindowCursor,
+  nextExploreWindowCursor,
+} from './explore-window-cursor';
+
+type Item = { id: string; types: { id: string }[] };
+
+const item = (id: string, ...typeIds: string[]): Item => ({ id, types: typeIds.map(t => ({ id: t })) });
+
+const key = exploreItemTypeKey;
+const ids = (items: readonly Item[]) => items.map(i => i.id);
+
+/**
+ * The measured production mix, Best sort (GEO-2690): claims own ranks 1-30 outright, then
+ * news stories run about 30% of the way down. Reproduced positionally rather than as a
+ * ratio, because the whole problem is *where* the off-type items sit, not how many exist.
+ */
+function measuredProductionWindow(size: number): Item[] {
+  const items: Item[] = [];
+  for (let rank = 0; rank < size; rank++) {
+    const isNews = rank >= 30 && rank % 3 === 0;
+    items.push(item(`${isNews ? 'news' : 'claim'}-${rank}`, isNews ? NEWS_STORY_TYPE_ID : CLAIM_TYPE_ID));
+  }
+  return items;
+}
+
+describe('exploreItemTypeKey', () => {
+  it('is stable across relation order', () => {
+    // `types` comes from relations, whose order is not meaningful. Two identical entities
+    // must not land in different diversity buckets because the rows came back swapped.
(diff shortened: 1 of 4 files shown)


ohohoreilly:
Added a second commit for **GEO-2695**, a pre-existing bug this work uncovered.

Every sort scanned 30 and served `EXPLORE_PAGE_SIZE` = 22, then advanced the cursor by 30. Ranks 23–30 of every page were fetched, decoded, and discarded — not a tail, a repeating band, roughly **27% of the feed unreachable on all three sorts**. The activity feed rides the New path, so it was listing 22 of every 30 entities in a space.

It's in this PR rather than a follow-up because the fix is the cursor the first commit already introduces, and applying it to New and Top **collapses the two paging models into one**. Best now differs from New and Top in exactly three ways — which connection it queries, how wide it scans, and whether it reorders — instead of also having its own pagination. Two PRs serially rewriting the same function would be worse to review than one.

The over-scan itself stays: `buildItems` drops entities with no displayable space, so scanning 30 to fill 22 absorbs that. Advancing past the remainder was the bug, and it can't be fixed by advancing less — these cursors are opaque offsets over the server's ordering, so there's no "advance by 22".

Still 5 pre-existing `tsc` errors and a green suite.

PR #2259, opened 2026-08-27
fix(debates): read the avatar stack from present_count

ohohoreilly:
Client half of GEO-2691's third attempt. Pairs with **geo-chat#74**, which adds the `present_count` field this reads.

## The overflow has now been computed from three populations

| basis | why it was wrong |
| -- | -- |
| `total_count` | counted offline holders under a control whose whole meaning is availability — a side with nobody available rendered a bare "+2" and no faces |
| `available_now_count` | **viewer-relative**: excludes the viewer and anyone they have already debated on that claim, so a claim you had actually argued showed an empty stack — to you and nobody else |
| **`present_count`** | the population geo-chat draws `participants` from |

Faces and count now describe the same people, and the same people for every viewer of the claim.

## `withViewerPosition` loses its unconditional `+1`

That bump existed because `available_now` excluded the viewer even after the server knew their position, so their face took a slot the overflow couldn't see. `present` includes them, so:

- `present_count` and `total_count` are adjusted **only while the server hasn't caught up**,
- `available_now_count` is **never** adjusted — it means "people this viewer could request", which the viewer is not and never becomes.

## A bug found on the way

The **rematch picker** was collateral damage from the earlier fix. It reports `available_now_count: 0` because a rematch is a fixed pair — and while the stack was gated on that number, its avatar faces were **silently never rendered**. It now reports `present_count: participants.length`, the people that surface actually knows about.

## Two changed expectations, both of which encoded a bug

1. **"+2"** for a side where the viewer holds the position and four others are available. One short — it came from a count that excludes the viewer whose face is on screen. Now +3.
2. **"+1"** left on the side the viewer had just left, where the viewer had been that side's *only* present person. The count now follows the faces off the side rather than claiming somebody is still standing there.

Neither was a deliberate assertion about intent; both fell out of `available_now_count` being the basis.

## Verification

- 3256 tests / 308 files pass; `tsc` at its 5-error baseline.
- New test covers the regression directly — a pair-blocked opponent is drawn, with no overflow claiming anyone else is present.
- **Mutation-tested**: pointing the overflow and the render gate back at `available_now_count` fails 2 tests, including the new one.

### DIFF
--- a/apps/web/app/space/[id]/(space)/debates/rematches/[sessionId]/rematch-page-client.tsx
+++ b/apps/web/app/space/[id]/(space)/debates/rematches/[sessionId]/rematch-page-client.tsx
@@ -1597,8 +1597,13 @@ function rematchPositionSummaries(
       position_label:
         holders.find(holder => holder.position_label)?.position_label ?? responsePositionLabel(responseKind, position),
       total_count: holders.length,
-      // Only meaningful for the hub's "available now" counts; a rematch is already a fixed pair.
+      // Only meaningful for the hub's "available now" counts; a rematch is already a fixed pair,
+      // so there is nobody here the viewer would send a request to.
       available_now_count: 0,
+      // The people this surface knows are on the side, which is what the stack draws. It has to be
+      // its own number rather than reusing `available_now_count`: that is 0 here, and while the
+      // stack was gated on it these faces were silently never rendered.
+      present_count: participants.length,
       participants,
     };
   });
--- a/apps/web/core/debates/api.ts
+++ b/apps/web/core/debates/api.ts
@@ -397,13 +397,24 @@ export type DebateClaimPositionSummary = {
    * without standing ready to debate it is excluded. Not shown on the claim card.
    */
   total_count: number;
-  /** Online, available, not in a debate, not blocked either way. */
+  /**
+   * Eligible for *this viewer* to send a request to right now: excludes the viewer themselves and
+   * anyone they are already pair-blocked with on this claim. Drives the "Debate now" filter.
+   *
+   * Viewer-relative, so it is the wrong basis for anything presented as a fact about the claim.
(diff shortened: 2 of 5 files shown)


ohohoreilly:
**This is the fix for what Preston is seeing right now** — "The images still arent there", after geo-chat#73 deployed. Pushed a second commit that also removes the deploy-ordering constraint.

## Why the revert didn't fix it

The revert fixed geo-chat's half and left production in a state **worse than either end** of the change:

- geo-chat again sends every holder in `participants` ✅
- geogenesis still gates the whole stack on `available_now_count > 0` ❌

That count excludes the viewer *and* anyone they've already debated. On a claim whose only opponent is pair-blocked it's **0**, so `PositionAvatars` never renders and **no faces appear at all** — on precisely the claims Preston was testing. Verified: the running API pods are on `a7b62d72`, matching the `pin cfbe8b325371` commit, so the revert is genuinely live and this is the client half.

## The ordering hazard, which I nearly shipped in mirror image

`present_count` comes from geo-chat#74, and the two halves deploy independently. A client shipping first would gate on `undefined > 0` and **again draw nothing** — the same bug, arrived at from the other side. I'd told Patrick to merge this one first, which was wrong.

So the field is optional and read through `presentCount()`, falling back to the number of faces geo-chat actually sent. In the rollout window that draws every face and claims no hidden extras — strictly better than the old gate *or* a bare undefined. **The two PRs can now land in either order**, and it covers a geo-chat rollback too.

The two viewer adjustments deliberately leave `present_count` undefined when the server sent none rather than inventing a number, so the fallback keeps tracking the participant list they've already adjusted.

## Verification

- 3257 tests / 308 files; `tsc` at its 5-error baseline.
- New test pins the exact production state: no `present_count`, pair-blocked opponent, faces still drawn.
- Mutation-tested — reading `present_count` directly instead of through the fallback fails that test.

PR #2263, opened 2026-08-27
feat(assistant): add a geo-query sub-agent, and close preflight #1–#4

0xTaneja:
Adds graph queries, spreadsheet imports, and image attachments to the assistant, with a preview → correction → confirmation flow for curator imports and better recovery across long conversations.

For example, a curator can attach a workbook with People and Organizations tabs, map the rows and properties, correct one column or exclude a tab, and confirm the import. The assistant stages the rows with links between tabs pointing to the same entities. Missing ontology can be proposed, created with approval, and used in a new preview before publishing.

- **Graph queries and skills:** Adds a dedicated `geoQuery` agent using the embedded geo-query skill v0.2.8 for filters, counts, relations, and data-block/table inspection. Includes schema and pagination guidance, bounded query results, time budgets, and query error recovery. A separate `countEntities` tool returns exact totals without summing overlapping type counts; type listings now distinguish complete results from partial pages.
- **CSV/TSV/XLSX imports:** Parses files in a browser worker and handles every workbook tab containing data. Stores parsed sheets and mappings in IndexedDB so a refresh does not lose the import. Sends bounded column samples to the mapping agent, shows mappings and skipped columns per tab, and stages the included sheets after confirmation.
- **Ontology and mapping corrections:** Validates selected properties, data types, and relation targets against retrieved or pending local schema. Suggests missing properties for curator approval and makes approved schema available immediately. Restores previously selected ontology when correcting a mapping, preserves untouched columns, and keeps mappings fixed when including or excluding sheets.
- **Import correctness:** Resolves rows by name and type across the whole workbook before creating links, including forward references and cycles. Reuses pending entities, blocks duplicate pending imports, and validates the workbook before staging. Rejects conflicting rows, incompatible target types, and invalid conversions; preserves decimal precision and literal TEXT values such as `N/A`. Fixes browser CSV compatibility, delimiter/encoding handling, extra-cell preservation, and row/link statistics.
- **Image attachments:** Adds image selection and previews in the composer and lets the assistant use the attached file for entity image properties. A failed replacement preserves the existing image.
- **Responses and conversation state:** The Sonnet executor now writes the final answer directly, removing the separate opener/closer rewrite stages. Follow-ups reflect available tools and actual outcomes. Restates the current space on each question, fixes chat restoration after reload, and preserves active attachments and mappings through compaction. Adds a context meter and manual summarization, with guards against repeated compaction failures and stale results after changing chats. Stop/new-chat handling cancels pending import work and removes unfinished tool calls.
- **Search, identity, and membership:** Improves schema discovery across Root and trusted spaces, resolves missing type names, and returns actual space container IDs for navigation. Supplies the user's profile entity in context and adds explicit membership requests through the existing smart-account flow, reporting requests separately from granted access.

Validation from the local hardening pass:

- 843 targeted tests passed across 44 files; TypeScript, targeted ESLint, and whitespace checks passed.
- Fresh real-model API conversations exercised counts, mapping corrections, approved local ontology creation, workbook imports after compaction, and retry handling. Successful staging used an isolated store with fixture authorization; no graph edits were published.
- Browser checks using the production attachment components, parser worker, and real IndexedDB passed XLSX/UTF-16 parsing, image previews, reload persistence, loading state, and cancellation.
- Signed-in Vercel preview smoke testing remains pending. The full Next page crashed during local browser testing, so the component checks do not establish full application coverage.

Supported spreadsheet formats are CSV, TSV, and XLSX, up to 10 MB. Legacy XLS needs conversion to XLSX. Media-upload columns are explicitly reported as unsupported by the assistant spreadsheet importer; image attachments use the separate upload flow.


0xTaneja:
All Copilot Reviews are taken note of and resolved.

apps/web/core/chat/import/parse.worker.ts
```diff
@@ -0,0 +1,85 @@
...
+import { MAX_FILE_SIZE_BYTES, MAX_FILE_SIZE_MB, type ParseResult } from './types';
+
+export type ParseRequest = {
+  file: File;
+  /** Force a delimiter, overriding the sniff. */
+  delimiter?: string;
+};
+
+async function readSpreadsheet(file: File): Promise<ParseResult> {
+  const workbook = await readXlsxFile(file);
```
jwalkingjew: Blocking: XLSX numbers lose precision before the string-based coercion layer sees them. read-excel-file returns numeric cells as JavaScript numbers by default, and cellToString later calls String(value), so values such as 9007199254740993 or high-precision decimals are already rounded. The decimal-preservation test starts from a string and does not exercise this boundary. Please pass a precision-preserving parseNumber implementation and add an end-to-end XLSX fixture for a large integer and a high-precision decimal. The library documents parseNumber specifically for this case: https://github.com/catamphetamine/read-excel-file#numbers

apps/web/core/chat/geo-query-dispatcher.ts
```diff
@@ -0,0 +1,114 @@
...
+        if (!isToolUIPart(part)) continue;
+        if (part.type !== GEO_QUERY_TOOL_PART) continue;
+        if (part.state !== 'input-available') continue;
+        if (dispatchedRef.current.has(part.toolCallId)) continue;
+        dispatchedRef.current.add(part.toolCallId);
+
+        const input = ((part as { input?: unknown }).input ?? {}) as GeoQueryInput;
+        const toolCallId = part.toolCallId;
+
+        enqueue(async () => {
```
jwalkingjew: Stop, New Chat, and chat switching do not abort this request. This controller lives for the whole component mount, while those actions only scrub the tool part and the widget remains mounted. A stopped geoQuery can therefore run for up to 45 seconds, occupy the shared serialized tool queue, and later submit a result through the ref for a different transcript. Please use a per-tool-call controller and abort it when the corresponding pending part disappears, matching the import, edit, and join dispatchers.

apps/web/core/chat/import/use-file-attachment.ts
```diff
@@ -0,0 +1,242 @@
...
+   * later, after the user has looked at the mapping and said yes.
+   *
+   * Dropping the chip is also what stops the next message re-announcing the
+   * same file and inviting a second import of it.
+   */
+  const dismiss = React.useCallback(() => {
+    generationRef.current++;
+    parseController.current?.abort();
+    setAttachment(current => {
+      // The chip goes; the file stays, because `setEntityImage` may not run
```
jwalkingjew: An unconsumed image becomes unreachable here but is not released: dismiss clears the React state and revokes only the preview URL, while the File remains in the module-level ImageAttachments map. Later New Chat or remove cannot recover its id, and clearAll has no production caller. Repeating attach, send, and new-chat can retain up to 10 MB per image for the lifetime of the tab and keeps prior-chat bytes available. Please retain a lifecycle handle and clear unused files on replacement, chat reset or switch, history clear, and unmount while preserving the current-chat retry case.

apps/web/app/api/chat/chat-system-prompt.ts
```diff
@@ -331,17 +404,19 @@ You can edit the graph on the user's behalf in spaces where they're a member.
...
+- **Joining a space.** When the user explicitly asks to join / become a member of / request access to a space, call \`joinSpace\` with a spaceId from a tool result this turn (\`listSpaces\` first when they named it). This is the one transaction you sign for them, and it goes through with no further confirmation step — so call it ONLY on an explicit request. Wanting to read, search, browse or navigate a space is NOT a request to join it; navigate instead. It submits a **request** the space's editors vote on: say you've requested membership, never that they have joined or now have access. On \`{ ok: false }\` **nothing was sent** — report the specific reason and never describe it as a request you just made: \`already_member\` (they're already in), \`already_requested\` (a request they made earlier is still up for a vote — tell them to wait on that one, you did not send another), \`not_joinable\` (a personal space; only public DAO spaces have a membership flow), \`no_personal_space\` (their account is still finishing setup — suggest retrying shortly), \`space_not_found\`, \`request_failed\`.
```
jwalkingjew: This makes an LLM instruction the only authorization boundary for an on-chain transaction. The client signs immediately with no confirmation of the exact target, so indirect prompt injection or a model or tool-selection error can submit a membership proposal the user did not intend. Please add a client confirmation step bound to the target space, or get explicit product and security sign-off for treating natural-language intent as sufficient transaction authorization.

apps/web/app/api/chat/geo-query/route.ts
```diff
@@ -0,0 +1,367 @@
...
+      if (query.trim().length === 0) {
+        return {
+          error:
+            'Your query arrived empty because the document you were writing ran past the output limit and was cut off mid-call. Send a shorter one — fewer aliases, fewer fields.',
+        };
+      }
+      // A query that outruns the loop budget takes the answer turn down with
+      // it, so it may only have whatever is left of that budget.
+      const remaining = TOOL_LOOP_BUDGET_MS - (Date.now() - startedAt);
+      const result = await runGeoGraphql(
```
jwalkingjew: The skill itself documents a generated query shape that has OOM-killed API pods, but runQuery forwards any model-generated GraphQL document here. Prompt guidance and a client timeout do not protect server memory once execution starts. Please enforce query complexity in code before forwarding, at least rejecting large root pages combined with unfiltered nested relations, or use an AST or query-cost limit.

PR #2300, opened 2026-08-28
feat(debates): full-screen matchmaking hub at /matchmaking

jwalkingjew:
Closes [GEO-2726](https://linear.app/geobrowser/issue/GEO-2726/debates-full-screen-matchmaking-hub-at-a-directly-linkable-route)

## ⚠️ First: the blocker is stale, and that changed the shape of this

The ticket says to settle this before building, so I did — and it's already settled.

It was written against a geo-chat that returned `topics: []`, ignored `topic_id`, and took a **single** `space_id`, which would have made a permanent facet rail a filter over the few hundred rows in memory. **[GEO-2659](https://linear.app/geobrowser/issue/GEO-2659/geo-chat-serve-space-and-topic-facets-with-counts-for-the-matchmaking) shipped on 26 Aug** (geo-chat #60). In `master` today:

- `api.ts` sends `space_ids` **and** `topic_ids` (both repeatable) — multi-select, server-side.
- `MatchmakingFacets` carries `space_facets` and `topic_facets` with **counts**, computed server-side over the whole filtered corpus, each dimension narrowed by the other.
- `claims-tab.tsx` already reads `facets?.topic_facets` — real counts, not client derivation.

So the fallback the ticket describes — *"ships without a topic rail, topics as best-effort narrowing"* — **isn't needed**. The rail is here with honest counts. That's the single biggest difference from what the ticket anticipated.

## What's built

| Zone | |
|---|---|
| **Facet rail** (left) | Show / Spaces / Topics open rather than menued, with server counts. Topic search appears past 12 rows — a 400-claim sample carries 329 distinct topics. |
| **Claims** (centre) | Same `MatchmakingClaimCard`, 1→2→3 column grid reflowing on the **centre column's** width. Search pins to the centre only. |
| **Live rail** (right) | Requests → Matches → Available now. Urgency order, not tab order. |

Plus: `/matchmaking` route, expand control in the panel header, filter state carried across the hop, signed-out shape, container-query collapse (facet rail first, then live rail).

## Decisions I made

**Route: `/matchmaking`**, per the ticket's recommendation — `/debate` vs the existing `/debates` differ by one character and mean different surfaces. One-line change if you'd rather have `/debate`.

**One `ClaimsTab` behind a `layout` prop**, not a fork. The panel keeps its menu row; the workspace swaps it for the rail and the grid. Same data, same effects, same tests.

**Signed-out live rail keeps People and explains the rest.** Presence is a fact about other people, not the viewer, so it's the one list that still answers — the rail leads with what works rather than being wholly a prompt.

## Two things lifting the filter state taught me — both caught by tests

**The setters must preserve `useState`'s identity bail-out.** Two effects hand back the *same* array when nothing changed and rely on React bailing out. Spreading into a fresh object regardless made each a real state change that re-triggered the effect that caused it — an infinite loop that took the vitest worker down rather than just spinning. The setters now return `current` untouched when the value is unchanged.

**`close` inside `DebatesHubSurface` resolves to `window.close`.** My first pass passed `onNavigate={close}` in the panel header — but that component has no hub `close` in scope, and `window.close` has the same signature, so **tsc stayed quiet** and the panel simply never shut. The control now reads `useDebatesHub()` itself.

## Not in this PR

**The detail pane.** Selecting a claim to open it beside the list is the one Layout item deferred. `ClaimPageView` is a genuine embed candidate as the ticket says, but it needs selection state, a third column at ≥1280px, and its own collapse rule — enough to be worth reviewing separately rather than doubling this diff. Everything else in Requirements and Layout is here.

Also untouched, both open questions rather than omissions: **no confirmation on the availability toggle** (the panel has none; adding one is a product call), and the ticket's *"is the ask really the Claims tab?"* is assumed yes, as the design assumes.

## Testing

- `hub-live-rail.test.tsx` — new: stack order, signed-out shape, Privy routing, and nothing drawn until Privy resolves.
- `claims-tab.test.tsx` — rail vs menus per layout, and the panel→workspace filter carry-over.
- `debates-hub-panel.test.tsx` — the expand control's href and that it closes the panel.
- Full suite: **338 files, 3654 tests, all passing.** `tsc` clean; `eslint` clean (3 pre-existing warnings).

Not exercised in a browser — the grid reflow and the collapse breakpoints are what I'd want your eyes on.

### DIFF
--- a/apps/web/app/matchmaking/page.tsx
+++ b/apps/web/app/matchmaking/page.tsx
@@ -1,0 +1,27 @@
+import type { Metadata } from 'next';
+
+import { DebatesHubWorkspace } from '~/core/debates/matchmaking/hub-workspace';
+
+/**
+ * The full-screen matchmaking hub.
+ *
+ * Top-level and cross-space on purpose: the hub spans spaces by design, so a `/space/[id]/` URL
+ * would read as a bug the first time someone shared the link — and `/space/[id]/debates` is already
+ * the *watch* feed, which is a different product.
+ *
+ * `/matchmaking` rather than `/debate` because `/debate` and the existing `/debates` differ by one
+ * character while meaning different surfaces, which is a durable footgun for shared links and for
+ * reading logs. The path is jargon; the page is still titled Debates.
+ *
+ * No server-side fetching: every list behind this is geo-chat or the knowledge graph through
+ * react-query on the client, and the page has to resolve for signed-out visitors, so there is
+ * nothing to await here.
+ */
+export const metadata: Metadata = {
+  title: 'Debates',
+  description: 'Find a claim, take a side, and get paired with someone who disagrees.',
+};
+
+export default function MatchmakingRoutePage() {
+  return <DebatesHubWorkspace />;
+}
--- a/apps/web/atoms/index.ts
+++ b/apps/web/atoms/index.ts
@@ -36,6 +36,37 @@ export type DebatesHubTab = 'requests' | 'matches' | 'claims' | 'people';
 /** `null` while the debates matchmaking hub is closed. */
 export const debatesHubAtom = atom<{ tab: DebatesHubTab } | null>(null);
 
+/** Featured is the tab's own source rather than one of geo-chat's filters — see `claims-tab.tsx`. */
+export type DebatesHubClaimsFilter = 'featured' | 'all' | 'mine' | 'debate_now';
+
+/**
+ * What the viewer has narrowed the claims corpus to, held outside the tab that renders it.
+ *
+ * GEO-2726. The panel and the full-screen workspace are two layouts over one session, so expanding
+ * has to continue the search someone was in the middle of rather than restart it. Local state in
+ * `ClaimsTab` could not survive that hop: the panel unmounts as the route mounts.
+ *
+ * Deliberately not persisted. It describes a sitting, not a preference, and a filter set restored
+ * days later would open the hub onto a sparse page nobody asked for.
+ */
+export type DebatesHubFilters = {
+  search: string;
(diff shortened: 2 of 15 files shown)


ohohoreilly:
Closing this in favour of #2397, which is building the same surface and is the live track.

The two overlap on 85% of their files (11 of 13), and both *create* the same four: `app/matchmaking/page.tsx`, `hub-workspace.tsx`, `hub-facet-rail.tsx` and `hub-live-rail.tsx`. Neither has landed, so this is two implementations of one feature rather than a base and a follow-up — #2397 has 13 commits through 15 Sep and is out of draft; this has four from 28 Aug.

I checked what is here that #2397 does not touch, so that nothing is lost by closing:

- **`debatesHubFiltersAtom` / `DEBATES_HUB_DEFAULT_FILTERS`** (the GEO-2726 idea: hold the narrowing outside the tab so expanding the panel continues the search instead of restarting it). **Already on master, in a more developed form** — per-list search/space/topic atoms, `debatesHubFiltersOwnerAtom` so filters are never handed to a different viewer across a sign-in, the `*SeedSpentAtom` flags, and a single `resetDebatesHubFiltersAtom`. The design in this branch is the early version of what shipped.
- **`matches-tab.tsx`** — that file has since been **deleted** on master; `matches-list.tsx` replaced it, and #2397 works against the replacement.

So the reasoning survives in the code and the ticket, and nothing here needs carrying across. #2397 also adds what this did not: `hub-workspace.test.tsx`, `claims-filter-params.ts` (the narrowing in the URL, so a filtered workspace is linkable), and `lobby-tab.tsx`.

Reopen if the intent was that this branch is the one to finish — the only thing driving the choice is that #2397 is the one still moving.

PR #2310, opened 2026-08-31
feat: open a share sheet instead of blocking on video prep

webmagic123:
Feat:
  Replaced the old "Preparing video for sharing…" behavior on the debate Share button with a proper share dialog — a 316×142 view with one-tap social sharing.

Implemented:
  - use-debate-share-action.ts slimmed to a tiny open/close hook.
  - debate-interaction-bar.tsx — Share button now just takes an onShare callback (dropped the old preparing/disabled/tooltip state).
  - Dialog rendered from both surfaces that show the bar: debate-feed.tsx (full-screen feed) and debate-explore-feed-card.tsx (explore card).


apps/web/core/debates/browse/share-dialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+      setRequested(false);
+      autoDownloadRef.current = false;
+    }
+  }, [enabled]);
+
+  React.useEffect(() => {
+    if (!autoDownloadRef.current) return;
+    if (prepared.status === 'ready' && prepared.downloadUrl && prepared.file) {
+      autoDownloadRef.current = false;
+      downloadPreparedVideo(prepared.downloadUrl, prepared.file.name);
```
b-d055: I think this is why Download doesn't work.

`download()` below just sets `autoDownloadRef` and `setRequested(true)`. The anchor click actually happens here, in an effect, seconds after the person pressed the button: there's a Privy token exchange, then the artifact POST, then a full streamed fetch of the video into memory. The 30s stall timeout in `downloadSocialVideo` tells you multi-second latency is the expected case, not the edge case.

By the time this runs, transient user activation is long gone. Safari on macOS and iOS refuses a programmatic `download` outside a gesture, and Chrome's automatic-downloads gate behaves the same way. Both fail quietly, nothing in the console and nothing on screen. I confirmed it by patching `HTMLAnchorElement.prototype.click`: `navigator.userActivation.isActive` is false when the anchor fires.

`debate-coordinator.tsx` already has the version that works. It prepares eagerly at mount (line 336), shows "Preparing video..." until ready (416-420), and only hands off inside the click handler once the blob exists (380-407). If the sheet prepared on open rather than on press, the click handler would have something ready to hand over and this goes away.

apps/web/core/debates/browse/share-dialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+                ariaLabel="Copy link"
+                onClick={onCopy}
+                tile={<GlyphTile icon={<Link />} />}
+              />
+              <ShareAction
+                label="Download"
+                ariaLabel="Download debate video"
+                onClick={download.download}
+                disabled={download.status === 'preparing'}
+                tile={<GlyphTile icon={download.status === 'preparing' ? <Spinner /> : <Download />} />}
```
b-d055: `idle`, `ready` and `error` all render identically here: enabled button, Download glyph, no toast, no message. `prepared.error` never gets read.

So every failure in `social-video-share.ts` vanishes: the artifact `onError` at 150-162, a fetch or CORS `TypeError` at 321-322, `assertCompleteVideoDownload` at 389-393 (which throws any time `receivedBytes` doesn't exactly match `content-length`), and the 30s stall. The only trace left is a PostHog event.

I drove the real dialog with the real `usePreparedSocialVideo` and forced each of those: artifact 404, `TypeError('Failed to fetch')`, content-length mismatch. Every one ended with the button enabled and `setToast` never called.

`setToast` is already in scope on line 34 and `onCopy` uses it. Surfacing `prepared.error` plus a different glyph on error would make this and the two comments below diagnosable instead of invisible.

apps/web/core/debates/browse/share-dialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+              <ShareAction label="Linkedin" ariaLabel="Share on LinkedIn" onClick={onLinkedIn} tile={<LinkedIn />} />
+              <ShareAction
+                label="Copy link"
+                ariaLabel="Copy link"
+                onClick={onCopy}
+                tile={<GlyphTile icon={<Link />} />}
+              />
+              <ShareAction
+                label="Download"
+                ariaLabel="Download debate video"
```
b-d055: Worth checking that the debate actually has a `social_video` rendition before showing this tile.

The gating upstream is `hasProcessedVideo` (`playback-utils.ts:21-23`), which only proves `final_video` exists. `social_video` is a separate kind, and the debug page treats a missing one as a perfectly normal state: it renders "Share video rendition is missing." (`debug-debates-page-client.tsx:315`).

So for any debate rendered before the social rendition joined the media job, the tile appears, the POST 404s, and the comment above means nobody ever finds out.

Easy to see for yourself: turn on the "Debates debug tab per space" flag with Cmd/Ctrl+Shift+F, open `/space/<id>/debug-debates`, find a debate whose Share video card says the rendition is missing, then press Download on that same debate.

apps/web/core/debates/browse/share-dialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+/**
+ * The Share sheet for a debate: one-tap social hand-offs (Reddit, X, LinkedIn), a copy-link button,
+ * and a download of the prepared debate video.
+ */
+export function DebateShareDialog({ open, onOpenChange, debate, spaceId }: Props) {
+  const download = useDebateVideoDownload(debate.id, open);
+  const [, setToast] = useToast();
+
+  const shareUrl = () => `${window.location.origin}${NavUtils.toEntity(spaceId, debate.id)}`;
+  const shareText = 'Watch the debate on Geo!';
```
b-d055: I don't think this piece landed. The ask was `"<Claim text>. Watch the debate on Geo! <Debate link>"` and what went in is the static half only. `d33b6087e` also dropped the `entityType` prop on the way through, and `ed5bd25af` only changed the LinkedIn separator to `\n`.

`debate.claim.claim` is right there on the prop and non-nullable (`core/debates/api.ts:20-26`).

Two things I'd fold into the same change:

Reddit's `title` param is the submission title, not body text, since link posts have no body field. So right now every debate shared from here gets the identical headline, which reads like spam both to human mods and to the AutoModerator title rules a lot of subs run.

And once a claim is prepended the length caps start to bite: Reddit titles stop at 300 characters, X at 280 including the URL. Without truncation you'd be trading a copy problem for a rejected-post problem.

apps/web/core/debates/browse/share-dialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+ * Download.
+ */
+function useDebateVideoDownload(debateId: string, enabled: boolean) {
+  const [requested, setRequested] = React.useState(false);
+  const autoDownloadRef = React.useRef(false);
+  const prepared = usePreparedSocialVideo(debateId, { enabled: enabled && requested, includePreview: false });
+
+  React.useEffect(() => {
+    if (!enabled) {
+      setRequested(false);
```
b-d055: Closing the sheet throws away whatever preparation was in flight, and also whatever had already finished.

`enabled` is `open`, so closing flips `usePreparedSocialVideo`'s `enabled` to false, and its cleanup at `social-video-share.ts:166-170` aborts the controller and revokes the blob while `:63-68` resets state.

I tested both directions. Close mid-prepare and the fetch signal aborts, so reopening and clicking fires a second `artifactMutate`; the person waited through the render and the multi-MB fetch for nothing. Close after it's ready and the finished result gets discarded too, so reopening refetches the whole video from zero. Nothing survives an open/close cycle.

There's a narrower version of the same problem: `downloadPreparedVideo` points an `<a download>` at the blob URL, and closing right after revokes it (line 169) while the browser may still be reading a multi-megabyte blob. Moving the revoke to unmount rather than tying it to `open` would cover all three cases.

apps/web/core/debates/browse/share-dialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+      await navigator.clipboard.writeText(shareUrl());
+      setToast(<span>Link copied!</span>);
+    } catch {
+      setToast(<span>Could not copy link.</span>);
+    }
+  };
+
+  return (
+    <Root open={open} onOpenChange={onOpenChange}>
+      <Portal>
```
b-d055, line 110: Small structural thing with a real accessibility cost: there's no `Dialog.Trigger` here, so `context.triggerRef` never gets populated. The Share buttons are plain buttons wired to `share.onOpen` (`debate-interaction-bar.tsx:52` and `:67`, `debate-explore-feed-card.tsx:191-199`).

Radix's modal content overrides focus restoration like this:

```js
onCloseAutoFocus: composeEventHandlers(props.onCloseAutoFocus, (event) => {
  event.preventDefault();
  context.triggerRef.current?.focus();
}),
```

The `preventDefault()` suppresses `FocusScope`'s own restore, and then `triggerRef.current` is null, so nothing gets focused at all. Tab to Share, press Enter, press Escape, and `document.activeElement` is `<body>`; the next Tab starts over at the top of the page. On the fullscreen scroll-snap feed that's a long way back.

The same cause means the Share buttons never get `aria-haspopup="dialog"`, `aria-expanded`, `aria-controls` or `data-state`, so a screen reader just announces "Share debate, button" with no hint that it opens anything.

Either wrapping the button in `<Trigger asChild>`, or passing `onCloseAutoFocus={e => { e.preventDefault(); openerRef.current?.focus(); }}` on `Content`, fixes both.

apps/web/core/debates/browse/share-dialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+                type="button"
+                aria-label="Close"
+                onClick={() => onOpenChange(false)}
+                className="absolute right-0 grid size-4 place-items-center text-grey-04 transition-colors hover:text-text"
+              >
+                <Close />
+              </button>
+            </div>
+
+            <div className="mt-5 flex justify-center gap-4">
```
b-d055, line 135: Against the Figma on GEO-2390 (node 73469-13348) this is missing WhatsApp and Email, and the read-only URL field with the inline copy button became a "Copy link" tile instead. Download isn't in that design at all, which seems reasonable as a debate-specific addition, but worth confirming which frame is actually governing since Linear doesn't have another one attached.

the panel is a fixed `h-[142px] w-[316px]`, so 268px of content width. Today's five tiles measure 264px and just fit. Six tiles comes to 320px and overflows by about 52px, and the URL field still needs height the fixed panel doesn't have. "WhatsApp" also measures 44.4px in a 40px column, so its label will spill the way "Download" already does.

apps/web/core/debates/browse/share-dialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const [, setToast] = useToast();
+
+  const shareUrl = () => `${window.location.origin}${NavUtils.toEntity(spaceId, debate.id)}`;
+  const shareText = 'Watch the debate on Geo!';
+
+  const openComposer = (href: string) => {
+    window.open(href, '_blank', 'noopener,noreferrer');
+  };
+
+  const onReddit = () => {
```
b-d055, line 75: None of the four hand-offs here emit any analytics, and neither does the download.

Before this PR a successful share went through `handoffPreparedSocialVideo` and fired `captureSocialVideoEvent('debate_social_video_handoff_resolved', { debate_id, method })` (`social-video-share.ts:230`).

The awkward part is the event doesn't drop to zero, because `debate-coordinator.tsx:392` still fires it. So any dashboard built on it will show a step change at deploy that looks like a product regression and isn't, while the four buttons this PR is actually about have no success metric at all. After merge there's no way to answer "is anyone using the Reddit button".

Error telemetry for the download path does survive via `debate_social_video_preparation_failed`, for what that's worth.

apps/web/core/debates/browse/share-dialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * The Share sheet for a debate: one-tap social hand-offs (Reddit, X, LinkedIn), a copy-link button,
+ * and a download of the prepared debate video.
+ */
+export function DebateShareDialog({ open, onOpenChange, debate, spaceId }: Props) {
+  const download = useDebateVideoDownload(debate.id, open);
+  const [, setToast] = useToast();
+
+  const shareUrl = () => `${window.location.origin}${NavUtils.toEntity(spaceId, debate.id)}`;
```
b-d055: Two smaller things on this line.

`window.location.origin` means a share from a preview deploy posts `geogenesis-git-*.vercel.app` out to Reddit or X or LinkedIn. There is a canonical origin at `app/layout.tsx:52`, but it's `ENV_URL` without a `NEXT_PUBLIC_` prefix so it isn't reachable from a client component as-is.

Separately, `debate.id` is the hyphenated geo-chat UUID, while the explore card links the same entity as dashless hex (`debate-explore-feed-card.tsx:150`). Both resolve fine, `IdUtils.isValid` matches either and the OG tags render both ways, so nothing is broken here. But there's no `alternates.canonical` in `generateMetadata`, so they're two canonical URLs as far as any scraper or analytics is concerned: split OG caches, split per-entity numbers, and a copied link that visibly doesn't match the address bar. `ID.uuidToHex(debate.id)` would line it up with the rest of the site.

b-d055 (commented):
@webmagic123  see comments

apps/web/core/debates/browse/share-dialog.tsx
webmagic123: Updated.

apps/web/core/debates/browse/share-dialog.tsx
webmagic123: Updated.

apps/web/core/debates/browse/share-dialog.tsx
webmagic123: Updated.

apps/web/core/debates/browse/share-dialog.tsx
webmagic123: Updated.

apps/web/core/debates/browse/share-dialog.tsx
webmagic123: Updated.

apps/web/core/debates/browse/share-dialog.tsx
webmagic123: Updated.

apps/web/core/debates/browse/share-dialog.tsx
b-d055: disregard this one

apps/web/core/debates/browse/share-dialog.tsx
webmagic123: Updated.

apps/web/core/debates/browse/share-dialog.tsx
webmagic123: Updated.

apps/web/core/debates/browse/share-dialog.tsx
```diff
@@ -0,0 +1,252 @@
...
+  const socialVideoReady = hasSocialVideo(media.data);
+
+  // Prepare once the sheet has opened and keep it alive for the item's lifetime instead of re-doing
+  // it on every open.
+  const [everOpened, setEverOpened] = React.useState(false);
+  React.useEffect(() => {
+    if (open) setEverOpened(true);
+  }, [open]);
+
+  const download = useDebateVideoDownload(debate.id, everOpened && socialVideoReady);
```
b-d055: `everOpened` only ever goes false to true, so once the sheet has been opened once `enabled` stays true for the rest of the item's life. Two consequences.

First, `usePreparedSocialVideo` starts streaming the full MP4 into a Blob as soon as it's enabled, and its effect deps are `[debateId, enabled, includePreview, retryCount]`. Since `enabled` never flips back the effect never re-runs, so the `AbortController` in the cleanup never fires. Closing the sheet doesn't stop the download. Someone who taps Share just to copy a link still pulls the whole video in the background.

Second, and this is the part I'd worry about more, the blob is retained. `debate-feed.tsx:234` renders `debates.slice(0, visibleCount)` with `visibleCount` only ever growing, so it's append-only infinite scroll with no windowing. Feed items never unmount as you scroll past them. Every card that has had its sheet opened holds its own `File` plus object URL until the page navigates away. Open Share on ten debates over a session and that's roughly half a gig parked in the tab, which is realistic crash territory.

The old dwell timer on master was worse than this. But gating on `open` instead of `everOpened`, or only kicking off prep once Download is actually hovered or pressed, would avoid both.

b-d055 (approved):

PR #2316, opened 2026-08-31
feat(debates): debate share card, claim size ladder, and a preview route

ohohoreilly:
GEO-2755, **first slice**. Draft: the card and the claim fitter, plus a route to look at them. **Video stills and publishing onto the entity are not here.**

## Try it

```
/api/debate-og/preview?claim=Waking%20up%20early%20improves%20health%20and%20productivity
/api/debate-og/preview?claim=Crypto%20will%20become%20the%20primary%20way%20people%20transfer%20money%20across%20borders%20within%20the%20next%20decade
```

Optional `name1` / `stance1` / `name2` / `stance2`. The published card is generated once at publish time from a real debate's stills, so it can't be exercised until a debate has finished processing — this route exists so the design and the type ladder can be reviewed before any of that is wired up.

## The claim ladder has a real trap in it

Each rung starts where the previous one bottomed out, and that's load-bearing rather than tidy. The obvious *"try two lines from 64, else three lines from 64"* version is **not monotonic**: a ~104-character claim renders *larger* than a ~101-character one, because it happens to find a two-line fit the shorter one just misses. There's a test sweeping 30–210 characters asserting size never increases with length.

`fitWrappedText` is reused rather than reimplemented, as the ticket asks, so measurement matches what the ranking card already renders with instead of a second estimator that can drift. It needed exporting, plus one guard:

**It never fails — it truncates at its own minimum and returns that**, which is indistinguishable from a real fit by size or line count. A 600-character claim came back as two truncated lines at 40px and passed every geometric check while having dropped most of the claim. Content loss is the only honest discriminator, so the ladder compares text in against text out.

## The angled divider

Satori has no `clip-path` and no `mask`, so the diagonal can't be cut out of a panel directly. It's drawn as rotated opaque slabs sharing one pivot, over panels that deliberately overrun the seam. `transform: rotate` and `overflow: hidden` are both already proven in the ranking card, which is why this was chosen over pre-compositing upstream.

Worth noting the overrun-and-cover approach isn't an approximation: a 32px band cannot cover a vertical seam across a 40px lean at *any* straight position, which is precisely why the slabs must rotate as one rigid unit.

## Two deviations from the design, flagged not absorbed

**Calibre.** Specified for every piece of UI text, but it ships here as `.woff2` alone and Satori can't load that format. The existing ranking card registers no Calibre for the same reason. Geist Medium stands in until a `.ttf`/`.otf` cut exists — worth a decision.

**The Geo mark** in the footer is a placeholder square, not the real lockup.

## What's verified, and what isn't

Verified: 6 ladder tests, typecheck clean, and **Satori renders the card to SVG** — so the whole card sits inside its CSS subset, which was the main risk.

**Not verified: the final PNG and visual fidelity.** `next dev` won't start in my environment without `NEXT_PUBLIC_PRIVY_APP_ID` — the same class of local gap that kept geo-chat unreachable on GEO-2764. Someone with a working env should hit the preview route before this leaves draft.

## Still to do for the ticket

- Video stills, seeked to 10s into each speaker's first speaking segment (with the <10s → 50% rule), minding the per-slot offsets in `use-debate-playback.ts`.
- Real avatars via `AVATAR_PROPERTY_ID`, with the initials-on-gradient fallback.
- Publishing onto the debate entity under `7abfeb2d147a464e8e26efdd26441189`, mirroring the keyframe block — and per the ticket's decision, generated once at publish time, never regenerated, and never allowed to fail the debate publish.


ohohoreilly:
Preston's three points from Slack are addressed in 025910b8d.

**Corners of the video stills.** The radius was already in the code and already drawing nothing — it sat on the panel's wrapper, which clips nothing, so the still (an absolutely sized child) simply overran the curve and left a square corner. A radius only rounds what it clips. It now sits on the still itself and on the placeholder field. Verified on SVG, PNG and JPEG stills: the block's corner pixel is page-white and the image does not start until ~16px in.

**Divider not centred above.** It was centred at the block's *mid-height*, so the 40px lean carried the top crossing 20px right of centre. It is now centred where it crosses the top edge — the crossing the eye actually checks, sitting under the claim with white on both sides, and roughly where the artboard puts it (the design's band crosses the top at 572–604 against a card centre of 600).

One consequence worth a decision: the lean now falls entirely left of centre, so the left panel averages 40px narrower than the right. The design carries the same asymmetry and more of it, so this moves toward the artboard — but it gives up the exactly-equal panels an earlier review round asked for. If we want both, the lean has to shrink. Flagging rather than deciding.

**White line centred on the VS circle.** The badge sat on the black bar's centre, 8.5px right of the white bar's edge, which is why the line looked flush entering the top of the badge and ~15px adrift leaving the bottom. It is now on the white bar's right edge, so the lean splits evenly — about 6px either side instead of 2 and 15.

The geometry tests asserted the two rules this overturns (equal panels, badge on the black bar), so they are rewritten around the new ones rather than deleted. Every candidate position here is within 20px of every other, so a render never tells you which number was intended — that is why they are asserted at all.

PR #2318, opened 2026-09-01
fix: keep the mobile feed below the iOS URL bar

webmagic123:


apps/web/core/debates/browse/debate-feed.tsx
```diff
@@ -331,8 +356,10 @@ export function DebatesBrowseFeed({
 
   // Keep the feed in the same tree position whether or not a side panel is open, so
   // toggling the claims/join panel doesn't remount the players and restart playback.
+  // Mobile drops the fixed overlay: the document scrolls so Chrome's toolbar offset becomes a
+  // real, correctable scroll position instead of an invisible displacement of the fixed layer.
   return (
-    <div className="flex h-[calc(100dvh-2.75rem)] items-stretch md:fixed md:inset-0 md:z-[70] md:h-dvh md:bg-white">
+    <div className="flex h-[calc(100dvh-2.75rem)] items-stretch md:h-auto md:min-h-svh md:items-start md:bg-white">
```
b-d055: **640–767px gets a half-applied takeover: navbar hidden, sidebar still rendered.**

The new rule (`apps/web/styles/styles.css:288-295`) hides `[data-app-navbar]` up to **767px**. But `apps/web/app/entry.tsx:102` builds `fullscreenActive` from the ranking and community atoms only — `debateFullscreenActiveAtom` isn't in it — and the sidebar wrapper at `entry.tsx:131` is `sm:hidden`, i.e. **<=639px**. So from 640–767px (iPad mini portrait 744, iPhone SE landscape 667, tablet split view) `BrowseSidebar` paints its 200px `sticky h-dvh` column with no navbar above it. `browseSidebarOpenAtom` defaults to open, so that's the default path.

With `md:fixed md:inset-0 md:z-[70]` dropped here, the feed is in flow beside that column while the debate column is still `md:w-[calc(100vw-1rem)]` (line 449) — full viewport width, ignoring the 200px. `body { overflow-x-clip }` clips rather than scrolls, so ~100px of media is cut off each side.

`ranking-compose-fullscreen.tsx` does the same atom + data-attribute pairing, but its atom *is* in `fullscreenActive`, so its sidebar half works. Simplest fix: hide the sidebar in the same <=767 band.

apps/web/core/debates/browse/debate-feed.tsx
```diff
@@ -231,6 +235,25 @@ export function DebatesBrowseFeed({
     return () => setDebateFullscreenActive(false);
   }, [rendersFeed, setDebateFullscreenActive]);
 
+  // Mobile-only: document scroll — snap on root, hide navbar via data attribute, scroll to top.
+  React.useLayoutEffect(() => {
+    if (!rendersFeed || !isMdLayout) return;
```
b-d055: The navbar hide is gated on `isMdLayout`, which is `false` until after hydration (`getServerSnapshot` returns `false`). Items already carry `md:min-h-svh` from first paint, so on mobile the navbar paints once and then disappears — a visible shift on every load of the feed.

`styles.css:288` already wraps the rule in `@media (max-width: 767px)`, so CSS can own the breakpoint entirely: set `DEBATE_FEED_MOBILE_ATTR` gated only on `rendersFeed` and drop `isMdLayout` from this effect. Only `root.style.scrollSnapType` genuinely needs the JS gate — and it could move into the same media block as `html[data-debate-feed-mobile] { scroll-snap-type: y mandatory }`.

apps/web/core/debates/browse/debate-feed.test.tsx
```diff
@@ -1155,3 +1154,57 @@ describe('DebatesBrowseFeed ordering', () => {
...
+ *
+ * Two halves. `position: fixed` on iOS resolves against the layout viewport, which runs underneath
+ * the browser UI — and Chrome puts its bar at the top where Safari puts it at the bottom. And a
+ * nested scroll container means the document never scrolls, so Chrome never collapses that bar and
+ * the page stays in the state where a fixed element is misplaced.
+ *
+ * So on mobile the document scrolls instead: nothing is positioned against the layout viewport, and
+ * the bar collapses the way it does on any ordinary page.
+ */
+describe('DebatesBrowseFeed mobile document scrolling', () => {
```
b-d055: `window.matchMedia` is `undefined` under this vitest/jsdom setup (nothing in `vitest.setup.ts` mocks it, and `setupTests.ts` isn't wired in), so the new `typeof window.matchMedia !== 'function'` guards make `useIsMdLayout()` permanently `false` here. These three tests assert only static Tailwind class strings, which render identically either way — so none of the new JS-gated behavior is exercised:

- the `md:` layout effect never runs: `data-debate-feed-mobile`, `documentElement.style.scrollSnapType`, `scrollTo(0, 0)`
- `observerRoot` is always `scrollEl`; the mobile `null` path is never taken
- `MockIntersectionObserver` drops its options arg, so the root can't be asserted at all

`debate-room-page-client.test.tsx` already has a `setMobileLayout()` matchMedia stub — that plus recording the observer options would cover the actual regression.

apps/web/core/debates/browse/debate-claims-panel.tsx
```diff
@@ -122,7 +122,7 @@ export function DebateClaimsPanel({ debate, onClose }: { debate: Debate; onClose
   }, [onClose]);
 
   return (
-    <aside className="flex w-[360px] shrink-0 flex-col border-l border-divider bg-white md:w-full">
+    <aside className="md:shadow-2xl flex w-[360px] shrink-0 flex-col border-l border-divider bg-white md:fixed md:inset-x-0 md:top-auto md:bottom-0 md:z-[80] md:h-[85dvh] md:w-full md:rounded-t-[16px] md:border-t md:border-l-0">
```
b-d055: `md:shadow-2xl` compiles to nothing — `styles.css:46` resets `--shadow-*: initial` and defines no `--shadow-2xl`, so the sheet ships shadowless. Use `md:shadow-big` (`0 1px 21px`): its near-zero Y offset reads above the sheet's top edge, unlike the downward-offset `shadow-panel`/`shadow-card`. Note ~9 other `shadow-2xl` uses in the repo are already dead for the same reason.

apps/web/core/debates/browse/debate-feed.tsx
```diff
@@ -231,6 +235,25 @@ export function DebatesBrowseFeed({
     return () => setDebateFullscreenActive(false);
   }, [rendersFeed, setDebateFullscreenActive]);
 
+  // Mobile-only: document scroll — snap on root, hide navbar via data attribute, scroll to top.
```
b-d055: `data-debate-feed-mobile` is the fifth copy of this html+body takeover effect — see `ranking-compose-fullscreen.tsx:32-55`, `ranking-compose-create-entity-panel.tsx:146-167`, `entity-side-panel.tsx:410-423`, `power-tools-screen.tsx:102-120` — and `styles/styles.css:290-291` clones the navbar-hide rule at `styles/styles.css:640-641`: same selector pair, same `display: none !important`, only the attribute and breakpoint differ.

Full reuse isn't available here — `RankingComposeFullscreen` is `fixed inset-0`, which is exactly what this PR removes. The honest extraction is a small `setTakeoverAttribute(name)` util beside `core/utils/hide-main-scrollbars.ts`, leaving just the snap/`scrollTo` logic here.

PR #2330, opened 2026-09-01
[DO NOT MERGE] experiment: hardcode featured spaces to isolate the GEO-2777 cold path

jwalkingjew:
**Experiment only — do not merge, do not review.**

Isolation test for [GEO-2777](https://linear.app/geobrowser/issue/GEO-2777/speed-up-explore-load-replace-the-featured-spaces-tree-walk-fix-the). Opened solely to get a Vercel preview URL to measure against.

`fetchFeaturedSpaces` returns a frozen list of the five spaces the live traversal actually produces, so the discovery step makes zero queries. Measuring a cold instance of this build against prod answers one question: is the featured traversal what costs ~11s on the signed-out explore feed?

Established so far:
- cold 11.5s / warm 0.37s, same code
- cold *boot* is only 0.22s (control: same route, early return, no outbound request)
- the backend API is never cold (0.39s first hit after 8 min idle)
- DNS is clean, no IPv6 fallback trap
- the traversal's module-level cache is the only per-instance cache on this path

Will be deleted once measured.


ohohoreilly:
Closing this as its own body asks — "experiment only, do not merge, will be deleted once measured".

The measurements are safe. The round-by-round traversal numbers are already in GEO-2777's description, and the cold-start findings that existed only here (cold 11.5s / warm 0.37s with discovery stubbed, 0.22s cold boot, backend never cold, clean DNS, module-level cache the only per-instance cache) are now copied into a comment on GEO-2777.

Worth restating the conclusion, because it is the opposite of what the experiment set out to confirm: with `fetchFeaturedSpaces` stubbed to a frozen list, a cold instance still took 11.5s. The featured traversal is about 2.95s of the signed-out explore load, not the ~11s. Roughly 11s remains unaccounted for and no fix on GEO-2777 addresses it — that is noted on the ticket as an open question.

Fix 1 lives on in #2327, which stays open.

PR #2362, opened 2026-09-04
Feat/remove individual edits from governance slide up

webmagic123:
Feat:
Select which edits go in a publish, and discard the ones you don't want.

- a checkbox (checked by default) to exclude an edit from this publish while keeping it locally
- a Discard control to throw that edit away and revert to the published value, with an undo toast
- a change count beside the title
- collapse-on-deselect, so an excluded row shrinks to its header instead of leaving a tall diff panel in the way

Implemented: (Block — deselection is disabled with a tooltip explaining the dependency.)
A module, core/review/publish-selection.ts 

 The rows I see and the ops that publish are not the same set. groupBlocksUnderParents folds block entities into their parent and removes them from the top level, so filtering the publish arrays by the visible ids would publish a page while silently dropping the table on it.

- buildOwnershipIndex maps every real entity id -> the row that governs it, following both the folded blocks[] and the BLOCKS relations for nesting.
- selectOpsForPublish splits the publish arrays by that index.  Ops it can't attribute are kept and reported, never dropped.
- collectOpsForEntities is the mirror, for discard - a test holds the two to being exact complements.

buildIsNewEntity distinguishes created-here from already-on-graph, reading the store only for collectCandidateEntityIds and deliberately not the publish arrays, which are pending-only and would mark everything new.


apps/web/partials/review/review-changes.tsx
```diff
@@ -557,6 +742,16 @@ export const ReviewChanges = () => {
   const handleSubmit = React.useCallback(async () => {
     if (!activeSpace) return;
     if (!isReadyToPublish) return;
+
+    const dangling = findDanglingDependencies(ownershipIndex, selectedEntityIds, relationsFromSpace, isNewEntity);
```
b-d055, line 746: Uploading an image, cover, avatar or video blocks publishing, even with every row selected. I reproduced this locally.

**Repro:**
1. Set a cover image on an entity and open review.
2. Leave everything selected, name the proposal and click Publish.
3. The status bar shows "Cannot publish: an edit still links to something excluded from this publish." Nothing publishes, and there's no row to include.

**Cause:**
- `createAndLinkImage` writes the new image entity (its values plus a Types relation) and a relation from the entity to the image with `renderableType: 'IMAGE'`.
- `fromLocal` puts that target in `imageEntityIds` and removes it from the rows, so `buildOwnershipIndex` never gives it an owner.
- All of the image's records are local and unpublished, so `buildIsNewEntity` reports it as new.
- `findDanglingDependencies` doesn't skip a target with no owner, so the relation to the image counts as dangling and `handleSubmit` returns early.

**Suggested fix:** in `buildOwnershipIndex`, assign IMAGE/VIDEO relation targets that aren't rows to the owner of the relation's `fromEntity`, the same way blocks fold into their parent. That fixes the publish block. It also means excluding the parent row excludes the image's own ops; right now they have no owner, so `selectOpsForPublish` keeps them and an orphan image publishes.

Could you also add a case to `publish-selection.test.ts` for a new media entity linked from a row? There's no media coverage there yet.

b-d055 (changes requested):
@webmagic123  see comment

b-d055 (approved):

PR #2371, opened 2026-09-04
revert: the Request debate position gate on ClaimEndSlot (#2354)

jwalkingjew:
Reverts #2354. Reported from the browser: remove your position, refresh, and the hub still shows a **rendered but permanently disabled "Request debate"** with nothing saying why.

## It is the frontend, not geo-chat

The button renders only when `match` exists, and `match` is what says the request is viable. #2354 then *additionally* required `readiness.viewer_response` to hold a position. When those two disagree you get exactly the reported state: a live match, an empty `viewer_response`, and a disabled button whose request would have succeeded.

Refreshing could not help, because it is not staleness — the gate disables on `viewer_response` regardless of what `match` says.

## Master already said not to do this — twice

`claim-end-slot.tsx`, added in #2353, one commit before #2354:

> A match is derived from the same `debate_claim_readiness` rows `create_debate_request_as` reads, **so no additional position check belongs here** — one against the graph would only be slower.

`request-gate.ts`, also predating #2354:

> The debates hub does not use this gate, **deliberately**. Its match data and `create_debate_request_as` both read `debate_claim_readiness`, which the in-flight response notification writes, so a graph-backed position check there would hold a button the server would have accepted.

Both were in files #2354 edited. Neither was engaged with. They describe the observed bug in advance.

## Why the gate is right for the picker and wrong here

The picker's opponent half is `opposing`, computed client-side from two positions — no server guarantee, which is why #2352 was a real fix and stays untouched. The hub, claim page and explore card use `match`, which is derived from the rows the request endpoint itself reads, so the position question is already answered before the button renders.

## Nothing worth keeping

Every line #2354 added exists to serve the gate: the required prop, the four wirings, the `requestPosition` composition, the `DebateRequestPosition` type, the three mock fields, the hub tests. The sizer change is moot once there is no longer a pending label to size against — the button is back to "Request debate" / "Requesting…", which that grid was already built for.

Checked before recommending the revert over a hand-written fix: a targeted fix would be the same diff with more chance of leaving a dead type or a stale mock field behind.

## Safety

- **No commit since #2354 touches any of the five files.**
- The only external reference to its API (`requestPosition`) is a test mock #2354 itself added.
- `debateRequestGate` keeps its three call sites in `rematch-page-client.tsx`, so #2352's picker fix is intact.
- **384 test files, 4292 tests passing.** Real `bun run build`, not just `tsc`. ESLint clean. `tsc --noEmit` at the same 5 pre-existing errors as master.

## The one addition

A note beside those two warnings recording that this was tried in #2354 and reverted, and why — cheaper than the next person rediscovering it in a browser. Plus a clause dropped that described the hub as a caller, which the revert makes untrue.

## Left open

The hub, claim page and explore card go back to offering a debate the server may still refuse — the pre-#2354 behaviour. That refusal is rendered rather than swallowed, per the `claim-end-slot.tsx` comment, so it is visible when it happens. Whether that is good enough is a separate question from this regression.

Draft on purpose: worth exercising the hub panel, claim page and explore feed against the Vercel preview before it lands.

### DIFF
--- a/apps/web/core/claims/browse/claim-end-slot.tsx
+++ b/apps/web/core/claims/browse/claim-end-slot.tsx
@@ -12,7 +12,6 @@ import Link from 'next/link';
 import type { Debate } from '~/core/debates/api';
 import { debatePath } from '~/core/debates/debate-routes';
 
-import { type DebateRequestPosition, debateRequestGate } from '~/core/debates/request-gate';
 import { useClaimMatchup } from './use-claim-matchup';
 
 /**
@@ -44,22 +43,10 @@ export function ClaimEndSlot({
   activeDebate,
   enabled = true,
   variant = 'inline',
-  position,
   className,
 }: {
   claimId: string;
   spaceId: string;
-  /**
-   * The viewer's position on this claim, from both clocks, so the offer only appears once geo-chat
-   * will honour it (GEO-2808).
-   *
-   * geo-chat validates a request against its *own* copy of the position and rejects an early one
-   * with `claim_response_required`. Every host of this slot already runs `useClaimPositionControl`,
-   * which holds both readings — so this is required rather than optional: a host that could not
-   * answer would be offering a debate it has no way to know is valid, which is what the hub and the
-   * feed cards were doing.
-   */
-  position: DebateRequestPosition;
   /**
    * The live debate on this claim.
    *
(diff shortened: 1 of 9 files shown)


jwalkingjew:
Closing — handed to another dev. The branch `preston/revert-2354-claim-end-slot-gate` is left in place if it's useful as a starting point.

PR #2397, opened 2026-09-09
Feat/full screen debates hub

webmagic123:
Feat:
A full-screen cross-space workspace for browsing claims and getting paired, alongside the existing 400px panel. 

Implemented:
A second surface for the debates hub. 400px panel for glance, this is for a session. Same parts, same data, but laid out for a screen, not a column.

Corpus browsing. Claims are presented as a grid that flows into 1, 2 or 3 columns based on the width of the center column itself — so when the rails collapse, the grid takes the space back without a second set of breakpoints. Cards get enough space to accommodate their topics, not just their titles.

Filtering in the open Scope (Featured / All claims / My positions / Debate now) is at the top of a left rail. Space and topic are below as multi-select lists with server-computed counts. The topic list has its own search box - a 400-claim sample has 329 distinct topics, which is a list to search, not a dropdown to skim. The rail is pinned, so scrolling the corpus does not carry your filters along.

Tab row live lists The right rail fills up with Requests, Matches and Available now, prioritized by urgency -- a debate request expires in about 25 minutes, so it can’t sit behind a tab you’re not looking at. On the panel it was just a number on a tab, obscured by whatever you were mostly doing.

When usersare not signed in, the route is resolved without an account. Claims and People work; the live rail tells you what Requests and Matches would provide, rather than displaying two empty headings.

When it is collapsing in, first comes the filter rail, then the live rail. The narrow layout goes back to the filter rail, but every narrowing in the filter rail is still reachable from the menus in the panel. Both measured against a named container, so they can't disagree on a width.


ohohoreilly:
Closed #2300 in favour of this one — it was a second, earlier implementation of the same surface (85% file overlap, and both created `app/matchmaking/page.tsx`, `hub-workspace.tsx`, `hub-facet-rail.tsx` and `hub-live-rail.tsx`). This is the live track, so the hub now has one.

Nothing from it needs carrying over — I checked the two files it touched that this does not. Its `debatesHubFiltersAtom` is already on master in a more developed form, and its `matches-tab.tsx` changes are moot because master deleted that file in favour of `matches-list.tsx`, which this branch already works against.

One thing worth knowing: this is currently conflicted against master, and `matches-tab.tsx` is part of why.

PR #2423, opened 2026-09-12
test-only: replay #1806 (smooth drag-and-drop reordering of relation chips) on master

jwalkingjew:
Replay of #1806 on current `master`, opened so a Vercel preview builds and the drag behavior can be tested. #1806 is 378 commits behind master and conflicting, so its preview is unusable. **Not for merge — test target only.** Review and merge the original.

## What

Makes drag-and-drop reordering of relation chips feel smooth, including wrapped multi-row chip lists.

## How

Migrates this one component to the next-gen dnd-kit:

- **`@dnd-kit/react` + `@dnd-kit/helpers` (0.5.0)** — measures the real DOM and FLIP-animates optimistic reordering, so variable-width wrapped chips reflow smoothly and land correctly. Reordering is optimistic during the drag; positions are committed on drop by reassigning the existing position set (same scheme as before).
- `chip.tsx`: `LinkableRelationChip` gains an optional `sortableDragHandleRef` (the new API marks drag handles via callback ref instead of spread listeners). The dots button remains the handle; grab UX unchanged.
- `editable-entity-page.tsx`: the add-entity button moves into the chip list via the new `afterChips` slot; IMAGE/VIDEO branching normalized to `renderableTypeStrict` (behavior-equivalent).

The classic `@dnd-kit/core`/`@dnd-kit/sortable` packages remain — other surfaces (table rows, tabs, etc.) still use them; the libraries coexist.

## Differences from #1806

Content is identical apart from one conflict that master's movement forced:

- `vitest.setup.ts` did not exist when #1806 branched. Master has since added it (the unmocked-network guard) and wired `setupFiles` in `vite.config.js`. So the `ResizeObserver` stub — needed because `@dnd-kit/dom` references `ResizeObserver` at module scope and jsdom does not implement it — is **appended** to master's setup file instead of creating one, and #1806's duplicate `setupFiles` key in `vite.config.js` is dropped (`vite.config.js` is now unchanged from master).

`reorderable-relation-chips-dnd.tsx` is byte-identical to #1806. Original authorship preserved on the commit.

## Verification on this branch

- `bun install` leaves `bun.lock` unchanged — lockfile is consistent with master's dependency set
- `tsc --noEmit`: only the 3 pre-existing failures already on master; none in the touched files
- `bun run build`: passes
- Full vitest suite: **462 files / 5122 tests passing**

### DIFF
--- a/apps/web/design-system/chip.tsx
+++ b/apps/web/design-system/chip.tsx
@@ -87,6 +87,8 @@ type LinkableRelationChipProps = {
   className?: string;
   disableLink?: boolean;
   sortableDragHandleListeners?: DraggableSyntheticListeners;
+  /** Callback ref marking the dots button as the drag handle (next-gen @dnd-kit/react). */
+  sortableDragHandleRef?: (element: HTMLButtonElement | null) => void;
 
   truncateLabel?: boolean;
   children: React.ReactNode;
@@ -199,6 +201,7 @@ export function LinkableRelationChip({
   className = '',
   disableLink = false,
   sortableDragHandleListeners,
+  sortableDragHandleRef,
   truncateLabel = false,
   children,
 }: LinkableRelationChipProps) {
@@ -216,6 +219,16 @@ export function LinkableRelationChip({
 
   const triggerRef = React.useRef<HTMLButtonElement>(null);
   const contentRef = React.useRef<HTMLDivElement>(null);
+
+  // Stable merged ref: an inline callback would re-run on every render (null then node),
+  // detaching/re-attaching the drag handle during hover/popover state updates.
+  const dotsButtonRef = React.useCallback(
+    (node: HTMLButtonElement | null) => {
+      triggerRef.current = node;
+      sortableDragHandleRef?.(node);
+    },
+    [sortableDragHandleRef]
+  );
 
   const shouldClamp = !truncateLabel && typeof children === 'string' && children.length >= 42;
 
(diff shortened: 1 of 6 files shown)


ohohoreilly:
Checked this while auditing the PR queue, and it is worth un-parking.

**The bug it replays is still live on master.** `reorderable-relation-chips-dnd.tsx` there still imports `@dnd-kit/core` / `@dnd-kit/sortable` with `CSS.Transform`, and `@dnd-kit/react` is not in `package.json`. Nothing else has landed the migration since May, so wrapped multi-row chip lists still reorder the way #1806 set out to fix.

I briefly closed #1806 in a sweep of stale conflicting PRs and have reopened it — that was my error, and this PR is the evidence I should have looked at first.

**This looks like the landable artifact, not #1806.** That branch is 378 commits behind for the sake of one component; this one is on master, `reorderable-relation-chips-dnd.tsx` is byte-identical to it, original authorship is preserved on the commit, and it reports `tsc` clean, `build` passing, `bun.lock` unchanged, and 462 files / 5122 tests green.

So "not for merge — review and merge the original" is the more expensive path by a wide margin. Un-drafting this and landing it looks strictly better, unless preserving #1806 as the merged artifact matters for reasons outside the code.

One thing to confirm before it lands, since it is the part most likely to bite: the `ResizeObserver` stub appended to `vitest.setup.ts`. `@dnd-kit/dom` touches `ResizeObserver` at module scope and jsdom does not implement it, so that stub is load-bearing for every test file the setup touches — not just these. Worth a look that it cannot mask a real missing-observer failure elsewhere.

PR #2442, opened 2026-09-15
feat(debates): surface claims over the video as they are said

jwalkingjew:
**Stacked on #2439** — review that one first; this targets its branch, so the diff here is only the during-playback layer.

Split out on its own because this is the part that risks being a distraction rather than an addition, and it should be judged on that alone. #2439 keeps the uncontested half: knowing when each claim was said, markers on the scrubber, and the card at the end that asks for positions once there is no video left to interrupt.

## What this adds

Claims appear while the debate is still playing. Each is a line of white text on a soft scrim — the same visual language as the subtitle, deliberately not a card — rising in the speaker's own corner and fading out with its own window. Newest at the bottom, nearest their name; earlier ones ride up above it.

- **Two at a time at most.** The stack grows upward and the subtitle sits a little higher on the same tile; a third reaches into it.
- **Capped to the debater's name row**, so a line never runs past their position chip. Measured with a `ResizeObserver` — names vary, and there is no CSS way to say "no wider than that absolutely-positioned sibling".
- **Two small icons under the text**, recessive until hovered, unmistakable once a side is taken. Thumbs for a stance claim, chevrons for a factual one, matching the split the rest of the app draws. The label survives as the accessible name and the tooltip.
- **No crowd split and no error text on the line.** Either would resize it while the reader is part-way through; the end card is where the numbers live.

## Two decisions worth a look

**Only confidently-placed claims appear here.** A line asserts *they are saying this now*, and being wrong about that misquotes a real person. Anything the resolver placed loosely stays in the panel and on the end card, where a list makes no claim about when.

**Opacity comes from the playhead, not a CSS animation.** Scrubbing into the middle of a claim finds the line at full strength rather than part-way through an animation that started when the element mounted.

## Testing

- `claim-ticker.test.ts` — stack ordering and capping, overlap, linger, fade in and out.
- `debate-claim-ticker.test.tsx` — icons name themselves, selected side, fade, answer reported once with its direction, no play/pause leak through the line.
- Full `apps/web` suite green (483 files, 5674 tests), `tsc --noEmit` clean apart from the three files already failing on master.

## Still unknown

How the fade reads in motion — whether the linger is too long or too short, and whether two stacked lines is one too many. That is the thing to watch rather than anything in the diff.


jwalkingjew:
Folded back into #2439 so both halves can be worked on together — this branch merged into that one, which is why GitHub marked this as merged. Nothing has shipped to `master`.

The during-playback layer now lives in #2439 alongside the timing resolver and the end-of-debate scorecard. Easy to split out again later: the commits are intact, and the removal that created this branch is a single revert away.

PR #2446, opened 2026-09-17
Feat/community tab leaderboard

webmagic123:
feat: The curator leaderboard on the Community tab now pages through every curator, opens full screen on its own page, and shows debate participation as a column and an overview box.

Implemented:

1. Removed the row cap so the board comes back whole - buildRows (core/community/fetch-curator-leaderboard.ts). The server already computed every curator and then cut it to five, so deleting the slice was the entire server change.
2. Paged on the client using the app's existing controls.
3. Gave the board its own page — (app/space/[id]/(community-new)/community/leaderboard/page.tsx). It renders the same section with an expanded flag: all rows, no pager. In (community-new) rather than (space) so it doesn't inherit the tab strip it came from. 
4. Added a fifth count source beside the existing four — fetchDebateCounts (core/community/fetch-curator-leaderboard.ts). Counts through the Participants relation so both sides count equally, on the same time window as the others. One count serves both readers: the column reads row.debates, the box reads metrics.debates.


apps/web/core/utils/utils.ts
```diff
@@ -41,6 +41,7 @@ export const NavUtils = {
   // A bounty is a regular entity; its detail page is the entity page.
   toBounty: (spaceId: string, bountyId: string) => `/space/${spaceId}/${bountyId}`,
   toCommunity: (spaceId: string) => `${NavUtils.toSpace(spaceId)}/community`,
+  toCommunityLeaderboard: (spaceId: string) => `${NavUtils.toCommunity(spaceId)}/leaderboard`,
```
b-d055: This breaks for the root space. `toSpace` maps the root space to `/root`, so this becomes `/root/community/leaderboard`, and nothing serves that route (only `app/root/page.tsx` exists). The root space still shows the Community tab, since `space-tabs.tsx` links it as `/space/${spaceId}/community`, so clicking expand there leads to a 404. The "Community" back link in the new `leaderboard/page.tsx` has the same problem because it uses `toCommunity`. Could you build both as `/space/${spaceId}/community/...` the way `space-tabs` does?

apps/web/partials/community-tab/curator-leaderboard-section.tsx
```diff
@@ -223,20 +280,45 @@ export function CuratorLeaderboardSection({ spaceId, initialData }: Props) {
...
+        <div className="flex items-center gap-3">
+          <SingleSelectPill
+            value={period}
+            options={CURATOR_LEADERBOARD_PERIOD_OPTIONS}
+            onChange={setPeriod}
+            contentClassName="max-w-[180px]"
+          />
+          {!expanded && rows.length > 0 ? (
+            <Link
+              href={NavUtils.toCommunityLeaderboard(spaceId)}
```
b-d055: Agree with Copilot here. If I pick "All time" and then expand, the full page opens on "Last week", so it's not the board I expanded. Could you pass the period through, e.g. `?period=all`, check it against `CURATOR_LEADERBOARD_PERIOD_OPTIONS` in the page, and use it as the section's starting period? That probably means an `initialPeriod` prop, since today the starting value only comes from `initialData?.period`.

b-d055 (changes requested):
@webmagic123  please fix this + the changes co-pilot suggested

apps/web/core/utils/utils.ts
webmagic123: @b-d055 
Updated.

apps/web/partials/community-tab/curator-leaderboard-section.tsx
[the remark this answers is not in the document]
webmagic123: @b-d055 
Updated.

PR #2449, opened 2026-09-17
fix(debates): videos stuck on Loading…, plus player control updates

o-p-o-p-o:
Three changes to the fullscreen debate feed.

## 1. Debate videos sat on "Loading…" forever

Both video panels on a debate page never left the placeholder, even though every request to the debates API returned 200.

**Cause.** `useDebatePlayback` marked a debate's recordings as "already fetched" as soon as it *started* the signed-URL request. If the effect was cleaned up before that request settled, the re-run saw the mark and skipped, while the cancelled run threw its URLs away. Nothing ran again.

Two ways to hit it:
- **Dev:** StrictMode's mount → cleanup → re-mount does it on every page load.
- **Prod:** scrolling a card out of view and back before its URLs arrive.

It came in with the GEO-2895 re-activation guard (#2429, #2431).

**Fix.** The mark is set only once URLs are committed. A cancelled run leaves nothing behind, so the next run fetches. The GEO-2895 behaviour holds: re-activating a card that already has its URLs does not refetch or blank it.

**Tests.** Two new cases in `use-debate-playback.test.tsx`, both failing before the fix:
- loads URLs under `React.StrictMode`
- loads URLs when the card is deactivated and reactivated before the first fetch settles

## 2. Player controls

| | Before | After |
|---|---|---|
| Play/pause, desktop | Big centre glyph, only while paused | Always-visible button beside mute, top-left |
| Clicking the video, desktop | Toggles playback silently | Toggles, and flashes a centre play/pause glyph for ~600ms. Feedback only: `pointer-events-none`, `aria-hidden` |
| Corner controls | 32px | 42px (+30%), icons scaled to match |
| Mute after unmuting, mobile | Invisible for good (hover-only, and touch has no hover) | Stays visible |

Mobile otherwise keeps its current behaviour: no corner play button, and the centre glyph while paused.

The larger size applies to every corner control, including replay and mobile mute, since they share `ControlCircle`.

One overlap to be aware of: when a debate ends unvoted, the mute spot becomes Replay, and the new play button next to it also replays. Easy to hide the play button in that state if we'd rather not have both.

## 3. Claims icon matches Explore

The fullscreen feed's Claims button drew `InfoSmall`, a question mark in a ring. Explore's debate card uses `ExploreClaimsIcon`, a thin ring with an "i". The interaction bar now uses `ExploreClaimsIcon` in both its layouts (the desktop vertical rail and the mobile pill bar), so the claims icon is the same everywhere. No new icon: it reuses Explore's component.

## Testing

- `vitest` over `core/debates` and `partials/explore`: all pass (1,631 tests before the last control tweaks; the 194 feed/explore tests re-run after them).
- ESLint and Prettier clean on changed files.
- Checked in Chrome against testnet, on a debate page:
  - videos load and play
  - desktop play/pause and mute are both 42×42 and visible while playing and paused
  - the flash appears on click and has faded by 1.35s
  - on an emulated iPhone 13, mute stays visible after tapping unmute
  - the Claims button renders the same SVG as Explore's, on fullscreen desktop, fullscreen mobile and the explore card
- Not checked on a real phone.


apps/web/core/debates/use-debate-playback.ts
```diff
@@ -207,17 +209,14 @@ export function useDebatePlayback(debate: Debate, enabled: boolean) {
...
-        // Scrolled away mid-flight: nothing is committed, so release the key or the card
-        // would hold a claim on URLs it never received and never fetch again.
-        if (cancelled) {
-          releaseKey();
-          return;
-        }
+        // Cancelled mid-flight: commit nothing and leave the key unclaimed so the next run
+        // fetches again.
+        if (cancelled) return;
+        fetchedForRef.current = recordingsKey;
```
ohohoreilly, line 215: **The key is claimed before the commit, so a throw in between recreates the permanent "Loading…" this PR exists to remove.**

The claim happens on the line above `setUrls`, and the `.catch` no longer releases it on the grounds that "the key was never claimed". That holds for a rejected request, but not for anything that throws *after* line 215 and before the commit lands.

Concretely: `geoChatRequest` returns `undefined` on a 204 (`core/debates/api.ts:1642`), so `slot1Result.url` throws a `TypeError` once the key is already claimed. The card shows an error, `urls` stays null, and every later re-activation hits the `fetchedForRef.current === recordingsKey` early return — so it sits on "Loading…" forever with no retry, which is precisely the failure class this PR is fixing.

The old unconditional `releaseKey()` in `.catch` covered this case.

Either move the claim below `setUrls(...)`, or keep a release in the catch.

apps/web/core/debates/use-debate-playback.ts
```diff
@@ -195,10 +200,7 @@ export function useDebatePlayback(debate: Debate, enabled: boolean) {
     if (fetchedForRef.current === recordingsKey) return;
 
     let cancelled = false;
-    fetchedForRef.current = recordingsKey;
-    const releaseKey = () => {
-      if (fetchedForRef.current === recordingsKey) fetchedForRef.current = null;
-    };
+    fetchedForRef.current = null;
```
ohohoreilly, line 203: **Dropping the up-front claim also drops in-flight de-duplication.**

The comment just above notes that the explore feed activates on a bare `intersectionRatio >= 0.6` with no hysteresis, so a single drag can flip `enabled` several times. Each flip landing inside the first request's window now starts two more signed-URL POSTs, and `useRecordingUrl` is a mutation so nothing caches them — 2N round trips on a slow connection where it used to be 2.

Worth noting the amplification lands hardest exactly where the bug is worst: a slow connection widens the in-flight window that causes both.

Keeping a promise per `recordingsKey` and having later runs await it would fix the hang without the amplification.

apps/web/core/debates/browse/debate-feed-player.tsx
```diff
@@ -109,31 +128,45 @@ export function DebateFeedPlayer({ debate, active, preload = false, votes }: Deb
...
-              }
-            >
-              {mutedByUser ? <SpeakerMuted /> : <Speaker />}
-            </ControlCircle>
+          ready ? (
+            <div className="flex items-center gap-2">
+              {/* Desktop: a persistent play/pause beside the mute control. Mobile keeps the
+                  centred paused glyph and tap-to-toggle instead. */}
+              <ControlCircle
+                ariaLabel={playing ? 'Pause debate' : playbackEnded ? 'Replay debate' : 'Play debate'}
```
ohohoreilly, line 139: **Two adjacent buttons with the same accessible name.**

On desktop at the end of an unvoted debate, `showReplay` is true, so the second control is the `RetrySmall` "Replay debate" button — and this label resolves to "Replay debate" too. Both call through to `playFromStart`.

A screen reader announces "Replay debate button" twice with nothing to distinguish them. The PR body flags the visual overlap here; the duplicated accessible name is the sharper half of the same problem.

apps/web/core/debates/browse/debate-interaction-bar.tsx
```diff
@@ -50,7 +50,7 @@ export function DebateInteractionBar({
           presentation="debate-vertical"
         />
         <CircleAction label={String(commentCount)} onClick={onComment} icon={<Comment />} ariaLabel="Comments" />
-        <CircleAction label={String(claimsCount)} onClick={onClaims} icon={<InfoSmall />} ariaLabel="Claims" />
+        <CircleAction label={String(claimsCount)} onClick={onClaims} icon={<ExploreClaimsIcon />} ariaLabel="Claims" />
```
ohohoreilly, line 53: **Claims renders 25% smaller than the icons beside it.**

`ExploreClaimsIcon` hard-codes `width="12" height="12"`, while `Comment` and `Share` next to it are 16x16 — in both the vertical rail and the pill bar (also at :68).

Since the stated purpose of the change is icon consistency, this is worth a sizing `className`, which the component already accepts.

ohohoreilly (commented):
Reviewed at `d78a0e0`, 4 files / +124 -42. One medium, four low.

**The fix itself is right, and it matters more than the PR title suggests.** Not claiming `fetchedForRef` until the URLs commit is the correct shape, and both new tests genuinely fail against master's hook, so they are real regression tests rather than restatements. Worth knowing: this appears to be the cause of a live user report tracked on **GEO-2950** — grey panels that never resolve, reported from a 28-space Explore feed where cards activate and deactivate constantly. "Scrolling away and back before the fetch settles" fits that exactly. It would be worth confirming with the reporter once this lands.

It does **not** subsume GEO-2965, which is the asymmetric case — both panels holding URLs but painting at different times, downstream of this fetch. Neither should be closed on the strength of the other.

## On the overlap with #2453

Both PRs touch `use-debate-playback.ts` and `debate-feed-player.tsx`, and I flagged a likely conflict earlier. Having actually run the merge of the two heads rather than comparing hunk ranges: **only `use-debate-playback.test.tsx` conflicts.** Both source files auto-merge cleanly, with `isResuming` threading and `lastRunningPlayheadRef.current = null` surviving alongside this PR's changes. So the collision is real but much smaller than I first said, and it is confined to the test file.

## Verified, and deliberately not raised as findings

- The `md:` breakpoints are correct. This repo redefines `md` as `@media (max-width: 767px)` (`styles/styles.css:277`), so `md:hidden` is desktop-only and `hidden md:grid` is mobile-only — matching the stated intent rather than inverting it.
- `size-10.5` and `scale-[1.3]` compile under Tailwind v4.3.2 (`2.625rem` and `scale: 1.3`).
- All 13 `use-debate-playback` tests and all 194 `core/debates/browse` plus `partials/explore` tests pass on the PR head.

apps/web/core/debates/browse/debate-feed-player.tsx
```diff
@@ -98,6 +98,25 @@ export function DebateFeedPlayer({ debate, active, preload = false, votes }: Deb
...
+    visible: false,
+  });
+  const flashTimeoutRef = React.useRef<ReturnType<typeof setTimeout> | null>(null);
+  React.useEffect(
+    () => () => {
+      if (flashTimeoutRef.current) clearTimeout(flashTimeoutRef.current);
+    },
+    []
+  );
+  const toggleFromVideo = () => {
```
ohohoreilly, line 113: **The click flash fires on the "Loading…" placeholder.**

`toggleFromVideo` is not gated on `ready`, so clicking the placeholder on desktop flashes a play glyph even though no `<video>` exists yet and `resumeBoth` returns immediately.

That is feedback for an action that did not happen, shown on exactly the state this PR is fixing. Gating `setFlash` on `ready` is enough.

ohohoreilly:
Follow-up to my review: **#2453 has now merged**, so this needs a rebase before it can land.

I ran the merge rather than guessing at it. The damage is small and confined:

- `core/debates/use-debate-playback.ts` — merges cleanly
- `core/debates/browse/debate-feed-player.tsx` — merges cleanly
- `core/debates/use-debate-playback.test.tsx` — **conflicts**

So it is one test file to resolve, not the two source files. `isResuming` threading and `lastRunningPlayheadRef.current = null` from #2453 both survive alongside your changes.

Worth doing soon: GEO-2895 is sitting in review waiting on this PR, and the fix here looks like the cause of a live user report on GEO-2950.

ohohoreilly:
I merged `master` into this branch (`b2869d63`) so it is mergeable again — it was 13 commits behind and conflicting. **Merged rather than rebased**, deliberately, so nothing on your branch is rewritten and any local work you have still pulls cleanly.

Two conflicts, and one of them is a decision you should ratify rather than just inherit.

**`use-debate-playback.test.tsx`** — import lists only. Took the union: master added `afterEach`, this branch added `React`, both are used.

**`debate-interaction-bar.tsx` — I took master's side, which drops this branch's Claims icon change.** Not because it was wrong, but because **#2462 ("unify claim action icon") already landed the same intent** and deleted `partials/explore/explore-claims-icon` on the way. Keeping this branch's version would mean restoring a component master deliberately removed. The Claims icon is now `Warning`, matching its neighbours — which is what the change here was reaching for.

Worth knowing *how* that surfaced, because it is a trap: git merged around the deletion **silently**. A removed file and an unchanged import do not textually conflict, so the merge looked clean and the tests were what caught it — `Failed to resolve import "~/partials/explore/explore-claims-icon"`. If you had rebased this yourself without running the suite, it would have looked fine.

That also retires one of my earlier review findings here: the note about `ExploreClaimsIcon` hard-coding 12x12 against 16x16 neighbours is moot now, since #2462 unified them.

Verified on the merged result: 137 tests across 9 files in `core/debates` pass.

**Two things still pending on this PR:**

1. The medium finding from my review is unaddressed — the key is claimed on the line *before* `setUrls`, and the `.catch` no longer releases it, so anything throwing between the claim and the commit recreates the permanent "Loading…" this PR exists to remove. `geoChatRequest` returning `undefined` on a 204 is a concrete path to it.
2. **#2467 will now conflict with this branch** in `use-debate-playback.test.tsx`. Both touch it, and both are open.

This fix looks like the cause of a live user report on GEO-2950, and GEO-2895 is sitting in review waiting on it, so it is worth getting over the line.

PR #2452, opened 2026-09-17
feat(explore): offer account creation from the email capture confirmation (GEO-2948)

jwalkingjew:
Follow-on from #2444 (GEO-2925). Linear: [GEO-2948](https://linear.app/geobrowser/issue/GEO-2948/explore-offer-account-creation-from-the-email-capture-success-state)

The capture's confirmation was a dead end — "You are on the list" and a close button — at the one moment someone is most willing, having just handed over an address on purpose. It now offers **Create account** or **Skip**.

Skip is exactly the old behaviour, kept as an equal option rather than a smaller one: the newsletter signup they already completed is a real outcome, and nothing here should read as undoing it.

## Not asking for the email twice

Privy's headless email flow (`useLoginWithEmail`) takes the address we already hold, so the modal's first step never happens. `sendCode` fires the moment they press, so the mail is in flight while the card swaps to the code field.

**One correction to the original idea:** there is no code waiting from the newsletter signup — MailerLite sends none. The only code that exists is the one this requests. The saving is *not retyping an email*, not *skipping a step*.

After `loginWithCode` resolves, nothing in this component runs: `authenticated` flips, the existing guard unmounts the card, and the onboarding dialog picks up a new account with no profile on its own. That hand-off needed no wiring.

## Details worth a look in review

- **State comes from Privy's own `OtpFlowState`** (`sending-code` / `awaiting-code-input` / `submitting-code` / `error` / `done`) rather than a second copy of the same machine kept alongside it.
- **One code field, not six boxes.** Paste works without per-box splitting, screen readers get one labelled control, and the code arrives by mail — pasting is what people actually do. Non-digits are stripped rather than rejected, since a pasted code brings whitespace with it. `autoComplete="one-time-code"` lets iOS and Chrome offer it from the message.
- **`sendCode` failing falls back to the normal Privy dialog.** Captcha, an outage, an address it will not take — signing up is the point, and not retyping an email is a convenience that must not become the reason nobody can sign up.
- **Resend is always reachable**, not only after a failure: Privy retires a code after five wrong attempts, and mail sometimes just does not arrive. Neither is an error state, so the way out cannot live only inside one.
- **`packages/auth` now re-exports `useLoginWithEmail`** (one line plus a rebuild). Nothing used a headless OTP flow before this.

## Two things sabotage testing caught

Sending the live input instead of the address that was actually subscribed **passed at first**, because the test used tidy input. It is padded now — Privy will not take an address with spaces, and a trailing space is an ordinary thing to type.

The `explore-page` suite broke on the popup reaching wagmi through `usePrivySignIn`. The capture is stubbed there alongside the siblings it already stubs; I confirmed the provider tree wraps it properly in the app rather than assuming.

## Open, for you

Both need a look at the Privy dashboard, and I could not check either from here:

- **Captcha** — if enabled, headless `sendCode` has to satisfy it. The fallback covers it gracefully, but the shortcut would never work.
- **MFA** — Privy's modal handles enrollment and challenges; headless would need `useAuthMfa`. No MFA config in `core/wallet/privy.tsx` today.

Also unhandled by design: someone whose email **already has an account** gets a login, not a signup, so "Create account" is the wrong label for them. Worth deciding the copy before this leaves draft.

---

491 files, 5801 tests passing. Typecheck, lint and production build clean.


apps/web/core/hooks/use-prepare-onboarding.ts
```diff
@@ -0,0 +1,54 @@
...
+    (redirectTo?: string | null) => {
+      const search = searchParams?.toString();
+      setPostOnboardingRedirect(
+        redirectTo === null ? null : (redirectTo ?? `${pathname}${search ? `?${search}` : ''}`)
+      );
+      setName('');
+      setTopicId('');
+      setAvatar('');
+      setSpaceId('');
+      setStep('start');
```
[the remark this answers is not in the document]
jwalkingjew: Confirmed and fixed — and it is worse than a gap in my hook. `selectedTopicIdsAtom` is persisted, and **no sign-in path has ever cleared it**: not the hand-written resets I replaced, not the hook I extracted from them. Only sign-*out* did.

So on a browser where somebody abandoned onboarding after picking interests, the next account inherits those picks and `PendingPersonalSpaceRunner` submits them as membership proposals on its behalf. Cleared in the shared hook, which fixes all four entry points at once.

Worth flagging: my first test for this was vacuous. The atom defaults to `[]`, so asserting `[]` after the reset passed with the reset deleted. It seeds `["topic-a","topic-b"]` first now, and removing the clear fails it.

`14ac73296`

packages/auth/src/use-ensure-embedded-wallet.tsx
```diff
@@ -0,0 +1,77 @@
...
+  // -- it is an error, and an error here is the same dead end as having no wallet.
+  const requestedWalletRef = useRef(false);
+  useEffect(() => {
+    if (!authenticated || embeddedWallet || requestedWalletRef.current) return;
+
+    requestedWalletRef.current = true;
+    void createWalletRef.current().catch(() => {
+      // Already created elsewhere, or Privy refused. The effect below activates whatever turns up,
+      // and retrying here would only loop.
+    });
```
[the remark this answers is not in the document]
jwalkingjew: Right, and this one was a one-way door. `requestedWalletRef` was set before `createWallet` resolved and never released, so a single dropped request left the session without a wallet until a reload — every smart-account path behaving as though nobody had signed in, which is precisely the bug this hook exists to prevent.

It tracks in-flight rather than ever-attempted now, releases in a `finally`, and stops after three attempts. Bounded because an account Privy genuinely will not make a wallet for should not sit in a loop asking.

`14ac73296`

packages/auth/src/use-ensure-embedded-wallet.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const activatedAddressRef = useRef<string | null>(null);
+  const walletToActivate =
+    embeddedWallet ?? (linkedWalletAddress ? wallets.find(w => w.address === linkedWalletAddress) : undefined);
+
+  useEffect(() => {
+    if (!authenticated || !walletToActivate) return;
+    if (activatedAddressRef.current === walletToActivate.address) return;
+
+    activatedAddressRef.current = walletToActivate.address;
+    void setActiveWalletRef.current(walletToActivate);
```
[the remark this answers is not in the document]
jwalkingjew: Same shape, same fix. Recorded up front, a failed activation was indistinguishable from a successful one to every later render, and wagmi stayed empty for the rest of the session.

The address is recorded in `.then()` now, with a separate in-flight flag so a pending activation is not started twice, and the rejection is handled rather than floating.

`14ac73296`

apps/web/partials/explore/email-capture-account-step.tsx
```diff
@@ -0,0 +1,159 @@
...
+  // Mounted only while someone is signing up, which is the point of it living here: Privy's login
+  // hooks register on a shared emitter, and one of these sitting on every Explore visit would be
+  // registering callbacks beside the navbar's own login for every reader who never presses the
+  // button that leads here — which is what broke that button.
+  //
+  // The embedded wallet this login needs is not created here. It cannot be: this component is
+  // unmounted by the card's own visibility rule the instant `authenticated` turns true, which is
+  // the exact render in which the wallet becomes creatable. `useEnsureEmbeddedWallet`, mounted for
+  // the life of the app in `core/providers.tsx`, does it instead.
+  const { sendCode, loginWithCode, state: otpState } = useLoginWithEmail();
```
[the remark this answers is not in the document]
jwalkingjew: Good catch — the flow reported nothing at all. The three other manual entry points call `trackPrivyAuth`, and `AnalyticsUserIdentifier` covers session restores rather than the moment someone signs in, so this path was invisible in both directions.

Now reported via `useLoginWithEmail`'s `onComplete`, as `auth_flow: "manual_login"` with `link_source: "explore_email_capture"` so it is separable from the modal logins — the whole point of this entry point is that it starts somewhere they already were.

`14ac73296`

apps/web/partials/explore/email-capture-account-step.tsx
```diff
@@ -0,0 +1,159 @@
...
+  // Guarded as well as keyed on a stable callback: one code is what mounting means, and the guard
+  // holds even under StrictMode's deliberate double-invoke, where the dependency array alone does
+  // not. A second send mails a second code and silently retires the first, so the one the reader
+  // is looking at stops working.
+  const hasRequestedRef = React.useRef(false);
+  React.useEffect(() => {
+    if (hasRequestedRef.current) return;
+    hasRequestedRef.current = true;
+    void requestCode();
+  }, [requestCode]);
```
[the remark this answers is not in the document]
jwalkingjew: Fixed. A dismissal while the request was open could be answered by throwing up a login dialog, which is the opposite of what the reader asked for. A mounted ref guards the catch now.

`14ac73296`

apps/web/partials/explore/email-capture-popup.tsx
```diff
@@ -266,11 +298,46 @@ function EmailCapturePopup() {
...
             <p className="text-[28px] leading-[17px] font-medium tracking-[-0.84px] text-[#151515] max-[382px]:leading-[30px]">
-              You are on the list.
-            </p>
-            <p className="mt-[8px] text-[16px] leading-[19px] tracking-[-0.48px] text-[rgba(21,21,21,0.7)]">
-              We will be in touch about features, points, and the path to mainnet.
+              {wantsAccount ? 'Check your email.' : 'You are on the list.'}
             </p>
+
+            {wantsAccount ? (
+              <AccountStep
```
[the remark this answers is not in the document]
jwalkingjew: This was the sharpest of the six. Mounting the step is what sends a code, and the guard returns `null` for the whole card — so opening search or the chat panel mid-sign-up and closing it again mailed a second code and silently retired the one being typed. The reader would be holding a dead code with no indication why.

Hidden rather than unmounted while an attempt is live. `hidden` gives up the screen completely — layout, hit-testing and the accessibility tree — so it still yields to the overlay; it just does not forget where the reader was. Two tests: no second code and the typed digits survive, and the card is genuinely unreachable while the overlay is up.

Worth noting this fix had **no test at all** until I sabotaged it and nothing failed. It does now.

`14ac73296`

jwalkingjew:
On the three suppressed comments. `14ac73296`

**No coverage for `usePrepareOnboarding`'s redirect modes.** Fair, and more pressing than when it was one caller — it is four now. Added tests for all three modes (current URL, explicit destination, `null`) plus every persisted field it clears.

The `null` case is the one worth having: it is easy to write as `redirectTo ?? default`, which turns "clear it" into "use the current page", and the navbar's button depends on the difference. Reverting to `??` fails that test specifically.

**`EmbeddedWalletSync` only tests the wrapper, not the hook.** Correct, and I tried to fix it properly rather than leave it. It is not reachable from here: `packages/auth` resolves **its own copy** of `@privy-io/wagmi`, so `vi.mock` from the app does not intercept it — I wrote the full suite (create, activate, retry-after-failure, bounded attempts, failed-activation-not-recorded) and every case died on `WagmiProviderNotFoundError` because the real hook was still in play.

Making it work needs vitest, jsdom, RTL and a config in a package that has never had tests. That is test infrastructure rather than this feature's work, and moving the hook into the app to make it testable would break the rule that all Privy access goes through `@geogenesis/auth`. So I deleted the suite rather than ship one that does not exercise what it claims to.

The gap is real and I would rather name it than paper over it: the call site is guarded, the hook is not. Happy to file a ticket for a `packages/auth` harness — it is the right place for it and this hook is the argument for having one.

**"Create account" copy.** Declining this one, as a product call rather than a disagreement. The flow is already correct for an existing account: `loginWithCode` logs them in, `useEnsureEmbeddedWallet` no-ops because they have a wallet, and onboarding's own `!isRegistered` gate correctly skips it. So it is a label that is wrong for one case, not a behaviour that is wrong — and detecting which case applies *before* sending would mean an account-enumeration oracle on an anonymous public endpoint, which Privy deliberately does not expose. Preston has opted to leave the copy as is.

---

One thing from this round worth stating plainly: sabotage caught **two of my own tests being vacuous** — the topics reset (the atom already defaults to the asserted value) and the overlay fix (no test at all until reverting it changed nothing). That is the third time on this branch, same cause each time: a fixture already in the state being asserted. Both are real now.

493 files, 5840 tests passing. Typecheck, lint and build clean.

apps/web/core/hooks/use-prepare-onboarding.ts
```diff
@@ -0,0 +1,79 @@
...
+ * rest, but until now only sign-*out* cleared it — every sign-in path, including the hand-written
+ * resets this hook replaced, left it alone. `PendingPersonalSpaceRunner` reads it and turns it into
+ * membership proposals for the new personal space, so a browser where somebody abandoned onboarding
+ * after picking interests would hand those picks to the next account and submit them on its behalf.
+ * The redirect is taken at the press rather than on completion, because completion can arrive
+ * minutes later on a different URL.
+ *
+ * Extracted from `usePrivySignIn` when a second way into onboarding appeared (GEO-2948). Any path
+ * that signs someone in has to do this, and the headless email flow skipping it is not a difference
+ * anyone would notice until they resumed a stranger's onboarding.
```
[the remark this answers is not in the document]
jwalkingjew: You are right and my fix was half done — which is worse than not having started, because I described it as closed. `sign-in-prompt.tsx` and `use-ranking-compose-access.ts` both still hand-reset the old fields, so both still leak `selectedTopicIdsAtom` into the next account through `PendingPersonalSpaceRunner`.

I left them alone last round because their redirect behaviour differed, and I weighed that as a refactor question. It was a correctness question: the difference was in the redirect, and the bug was in the fields.

The hook takes `keepReturnTo` now, which is exactly what both want — `use-ranking-compose-access` holds its destination in a ref and `sign-in-prompt` has never set one, so writing the shared atom would have been a behaviour change smuggled in under a refactor. Neither changes behaviour; both stop leaking.

No hand-written reset remains outside sign-out and the onboarding dialog itself.

`e8b174f9a`

packages/auth/src/use-ensure-embedded-wallet.tsx
```diff
@@ -0,0 +1,115 @@
...
+    void createWalletRef.current()
+      .catch(() => {
+        // Either something else created it first -- in which case `embeddedWallet` is about to
+        // appear and the guard above will stop us -- or the request genuinely failed and the next
+        // render is welcome to try.
+      })
+      .finally(() => {
+        walletAttemptRef.current.inFlight = false;
+      });
+  }, [authenticated, embeddedWallet]);
```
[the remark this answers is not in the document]
jwalkingjew: Correct, and this one is worse than a missing retry — the code claimed a behaviour it did not have. Clearing a ref in `.finally()` schedules nothing, and neither dependency changes when creation fails, so "three bounded attempts" was one attempt with a comment describing three.

It only looked right because my tests called `rerender()` themselves, supplying the render that production never supplies. Retries are driven by state now, with a 2s delay and a real ceiling.

That also changed my mind about the harness I declined last round. This hook has been wrong four times — mounted where it unmounted itself, two refs recorded before their promise resolved, and now a retry that never retried — always with the same symptom of onboarding silently not appearing. `packages/auth` has vitest now and nine tests, and **none of them call `rerender`**, because that is exactly what hid this. Reverting to the ref version fails two.

`e8b174f9a`

packages/auth/src/use-ensure-embedded-wallet.tsx
```diff
@@ -0,0 +1,115 @@
...
+        activatedAddressRef.current = address;
+      })
+      .catch(() => {
+        // Left unrecorded on purpose, so the next render tries again rather than believing a wallet
+        // is active when it is not.
+      })
+      .finally(() => {
+        activatingRef.current = false;
+      });
+  }, [authenticated, walletToActivate]);
```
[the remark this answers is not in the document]
jwalkingjew: Same fix, same reasoning — activation reschedules through state now, bounded the same way.

The failure mode here was the nastier of the two: `wallets` can settle and then never change again, so a single rejected `setActiveWallet` left wagmi empty for the rest of the session with nothing to prompt another try. Covered by a test that advances timers rather than forcing renders, and sabotage confirms it: recording the address before the promise resolves fails two tests.

`e8b174f9a`

jwalkingjew:
The five suppressed comments. `e8b174f9a`

**`maxLength` truncating pasted codes.** Real, and it defeats the one thing this field is designed for. The browser enforces it on raw input before `onChange` runs, so "123 456" from a mail client became "123 45" and only then had its spaces stripped — five digits in a step that cannot be completed, looking to the reader like they pasted the wrong thing. Removed; the normalized `.slice(0, CODE_LENGTH)` was already doing the job in the only place it is correct.

The test pins the attribute rather than the behaviour, and I want to be explicit about why: `fireEvent.change` assigns `.value` directly, so jsdom never performs native truncation and a behavioural test would pass with `maxLength` restored. Pinning the absence does fail when it comes back.

**Both resend controls racing.** Also real. They stayed live during `submitting-code`, so pressing resend while a verification was in flight retired the very code being checked. A single `busy` flag now covers sending, verifying and the local in-flight request, and disables both controls.

**Test seeding only `name`.** Correct, and the third version of this mistake on this branch. The other four atoms default to exactly the values the reset produces, so they were asserted against a fresh store and would pass with their resets deleted. All five seeded now — and dropping any individual reset fails it, which I checked one at a time rather than assuming.

**`rerender` not exercising a double mount.** Correct — it preserves the instance and never replays setup/cleanup/setup, so the test claimed StrictMode coverage it did not have. It renders under `React.StrictMode` now.

**"Create account" copy.** Declining, as Preston's call rather than a disagreement. The flow is correct for an existing account: `loginWithCode` logs them in, the wallet hook no-ops, and onboarding's `!isRegistered` gate skips it. Detecting the case before sending would mean an account-enumeration oracle on an anonymous endpoint, which Privy deliberately does not expose.

---

Two things worth stating about this round.

`packages/auth` now has a test harness. I declined that last round as infrastructure rather than feature work, and that was wrong — the hook had been wrong four times by then, and this round found a fifth. Nine tests, and **none of them call `rerender`**, because supplying a render by hand is exactly what hid the retry bug.

And three of my tests were vacuous again, same cause each time: a fixture already in the state being asserted. That is now the reliable failure mode on this branch, and sabotage is the only thing catching it — including twice in this round, after the fix was already written.

493 files, 5843 tests in the app plus 9 in `packages/auth`. Typecheck, lint and both builds clean.

packages/auth/src/use-ensure-embedded-wallet.tsx
```diff
@@ -0,0 +1,124 @@
...
+    return () => {
+      cancelled = true;
+      if (timer) clearTimeout(timer);
+    };
+  }, [authenticated, embeddedWallet, createAttempts]);
+
+  useEffect(() => {
+    if (!authenticated || !walletToActivate) return;
+    if (activatedAddressRef.current === walletToActivate.address) return;
+    if (activateAttempts >= MAX_ATTEMPTS) return;
```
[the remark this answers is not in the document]
jwalkingjew: Correct. A session-wide budget assumes a fixed target, and the target is exactly what changes here — a linked wallet burning three attempts left the embedded wallet inheriting an exhausted count, so `useSmartAccount` never got the one wallet it can use.

Counted per address now, and the retry timer is discarded when the target changes because the effect is keyed on the address rather than the object.

Test drives the sequence you described: a linked wallet fails its three, the embedded wallet appears, and activation still happens for it. Restoring a session-wide count fails it.

`89b16f114`

packages/auth/src/use-ensure-embedded-wallet.tsx
```diff
@@ -0,0 +1,124 @@
...
+      .catch(() => {
+        if (cancelled) return;
+        timer = setTimeout(() => setActivateAttempts(attempts => attempts + 1), RETRY_DELAY_MS);
+      });
+
+    return () => {
+      cancelled = true;
+      if (timer) clearTimeout(timer);
+    };
+  }, [authenticated, walletToActivate, activateAttempts]);
```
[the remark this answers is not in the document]
jwalkingjew: Right, and this was the subtler of the two. A new object for the same address re-ran the effect, whose cleanup marked the in-flight attempt cancelled — so a **successful** activation was thrown away as stale while a second one ran for the same wallet.

Keyed by address now, with the wallet itself read through a ref so a fresh object cannot re-trigger it.

My first test for this proved nothing, which is worth flagging: it swapped the object *after* activation resolved, where the address guard short-circuits regardless. It swaps mid-flight now — a deferred `setActiveWallet`, object changed while pending — and reverting the fix fails it.

`89b16f114`

jwalkingjew:
The two suppressed comments. `89b16f114`

**Fallback losing its completion callback.** Declining, having checked rather than assumed. The sign-in is still tracked: `navbar-actions.tsx` renders `GeoConnectButton` for every logged-out reader (`if (!address)`), and its `onComplete` calls `trackPrivyAuth` unconditionally. So the fallback does emit a `manual_login` event — what is actually lost is one `link_source` field, on a branch that only runs once `sendCode` has already failed.

I built the keep-mounted version before deciding that. It needs a small state machine watching the modal open and close again, plus handling for a dismissed-without-signing-in modal leaving a stale code step on screen — real complexity in auth code to reclaim one attribution field on an error path. I took it back out and left the reasoning in a comment where the next reader will find it.

**"Create account" copy.** Declining again, as Preston's call. The flow is correct for an existing account: `loginWithCode` signs them in, the wallet hook no-ops, and onboarding's `!isRegistered` gate skips. Detecting the case before sending would require an account-enumeration oracle on an anonymous endpoint.

---

Worth reporting separately, because this round caused it and finding it was the most useful thing in the round.

The StrictMode test I added last round started failing **unrelated debates suites** at random — different tests each run, all passing in isolation. The easy call was "flake, unrelated". I bisected instead: clean at the previous commit, clean with that one test removed. So it was mine.

The cause is in `useAnyModalOpen`, which shipped in the merged capture PR. Teardown removes the container from `body` — itself a mutation — and the observer can be handed it before `disconnect()` lands, scheduling a frame that fires *after* cleanup, reads the document, and sets state on an unmounted component once the owning test has finished. StrictMode's double mount only made it frequent enough to notice.

That is precisely the failure mode `vitest.setup.ts` already documents: work escaping a test's lifetime and blaming whichever file happens to be running. A `cancelled` flag closes it, and two full runs are clean with the test back in.

493 files, 5843 tests in the app plus 11 in `packages/auth`. Typecheck, lint and both builds clean.

ohohoreilly (commented):
Review of the diff at `02024589`, 23 files, +1445/-118. Seven findings below, ordered by severity; finding 1 is the one I would fix before this leaves draft.

Checked and cleared:

- `subscribedEmail` can never be empty when `AccountStep` mounts — `'subscribed'` is the only result that sets `status: 'done'`, and it sets the email on the same line.
- The `hidden={hiddenByOverlay}` swap really does hide the card: the div sets no `display` utility, so the UA `[hidden] { display: none }` applies. The tests pass for the accessibility-tree reason, but the visual behaviour is correct too.
- `mountedRef`, `hasRequestedRef` and the `sendCodeRef`/`giveUpRef`/`openPrivyModalRef` indirections hold under StrictMode's setup -> cleanup -> setup, and the create/activate retry chains are correctly bounded at 3 (per-address for activation).
- The four `usePrepareOnboarding` call sites preserve each caller's prior redirect behaviour exactly (`returnTo: null` for the navbar, `keepReturnTo` for `use-ranking-compose-access` and `sign-in-prompt`, default for the vote buttons).
- `EmbeddedWalletSync` sits inside both `LazyPrivyProvider` and `LazyWalletProvider`, so `useSetActiveWallet` has its wagmi context; being outside `JotaiProvider` is fine since it touches no atoms.
- The new `useSearchParams()` call sites (`SignInPrompt`, `PrivyConnectButton`) do not create a prerender bailout — `LazyWalletProvider` is `ssr: false`, so the subtree never server-renders.

packages/auth/src/use-ensure-embedded-wallet.tsx
```diff
@@ -0,0 +1,152 @@
...
+  // one wallet it actually needs.
+  //
+  // Keying on the object is wrong because Privy can hand back a new object for the same address.
+  // That re-runs the effect, whose cleanup marks the in-flight attempt cancelled, so a *successful*
+  // activation is discarded as stale while a second one is already running for the same wallet.
+  const [activateAttempts, setActivateAttempts] = useState<Record<string, number>>({});
+  const addressToActivate = walletToActivate?.address;
+  const attemptsForAddress = addressToActivate ? (activateAttempts[addressToActivate] ?? 0) : 0;
+
+  useEffect(() => {
```
ohohoreilly, line 101: **The create effect ignores `useWallets()`'s `ready` flag, so it calls `createWallet()` for users who already have one.**

`UseWalletsInterface` is `{ wallets, ready }` (confirmed in `@privy-io/react-auth@3.32.2`), and this reads only `wallets`.

`EmbeddedWalletSync` is mounted for the life of the app, so on **every** page load of an already-authenticated session Privy sets `authenticated: true` from the restored session before `wallets` has hydrated. At that moment `embeddedWallet` is `undefined`, the guard passes, and `createWallet()` fires for a user who already has one — which Privy's own docs say "will error if the user already has an embedded wallet".

The rejection is swallowed, a 2s retry is scheduled, `createAttempts` bumps and the effect re-runs. If hydration takes longer than the retry delay — slow network, blocked Privy iframe — this burns all three attempts.

Gate on `ready`, or check `user.linkedAccounts` for an embedded wallet, before calling `createWallet`.

packages/auth/src/use-ensure-embedded-wallet.tsx
```diff
@@ -0,0 +1,152 @@
...
+    if (!authenticated || embeddedWallet || createAttempts >= MAX_ATTEMPTS) return;
+
+    let cancelled = false;
+    let timer: ReturnType<typeof setTimeout> | undefined;
+
+    void createWalletRef.current().catch(() => {
```
ohohoreilly: **Both failure paths are silent, which reproduces the exact symptom this file exists to prevent.**

This `catch {}` and the one at line 120 have no `console.error` and no Sentry. After `MAX_ATTEMPTS` the hook gives up permanently, with nothing emitted.

The file's own doc comment says the symptom of this hook being wrong is "onboarding simply never appeared, with nothing in the app pointing at the cause" — the exhausted-budget path reproduces that precisely.

`useSmartAccount` logs its init failures for the same reason (`apps/web/core/hooks/use-smart-account.ts`, `[SMART-ACCOUNT] initialization failed`). This should too.

packages/auth/package.json
```diff
@@ -19,7 +19,7 @@
   "scripts": {
     "build": "tsc",
     "format": "npx prettier . --write",
-    "test": "vitest"
+    "test": "vitest run"
```
ohohoreilly, line 22: **The new test suite never runs in CI.**

The 210-line `use-ensure-embedded-wallet.test.tsx` and this whole vitest setup are not reachable from the pipeline: `.github/workflows/ci.yml:64` is `bun turbo run --filter=web test` and `:88` is `bun turbo run --filter=web lint`, so only the `web` package is tested and linted.

The stated safety net for the hook that "decides whether an authenticated session ends up with a wallet" is wired to nothing. Widen the filter, or add a second job.

(Worth noting the `"test": "vitest"` → `"vitest run"` change here is a genuine fix — watch mode would have hung any runner that did pick it up.)

apps/web/partials/explore/email-capture-account-step.tsx
```diff
@@ -0,0 +1,213 @@
...
+  // Reported like every other way into sign-in. Without this the flow emits nothing at all:
+  // `AnalyticsUserIdentifier` covers session restores rather than the moment somebody signs in, and
+  // the three manual entry points each report their own. `link_source` separates this one from the
+  // modal logins, since the whole point of it is that it starts somewhere they were already.
+  const {
+    sendCode,
+    loginWithCode,
+    state: otpState,
+  } = useLoginWithEmail({
+    onComplete: args => trackPrivyAuth(args, { auth_flow: 'manual_login', link_source: 'explore_email_capture' }),
```
ohohoreilly, line 44: **Sign-up analytics are double-counted on the happy path.**

`useLoginWithEmail`'s `onComplete` and `useLogin`'s `onComplete` both subscribe to Privy's single `login` event.

The navbar renders `GeoConnectButton` for every logged-out reader (`apps/web/partials/navbar/navbar-actions.tsx:82`, `if (!address)`), and its `useGeoLogin` in `apps/web/core/wallet/wallet.tsx:50` tracks **unconditionally** — unlike `usePrivySignIn`, which arms on `requestedRef`.

So a successful `loginWithCode` here emits two events: one `manual_login` from the navbar, and one `manual_login` plus `link_source: explore_email_capture` from this hook.

The PR body notes the navbar's unconditional tracking for the *fallback* branch only. It fires in the happy path too.

bun.lock
```diff
@@ -172,8 +172,17 @@
         "wagmi": "^3.5.0",
```
ohohoreilly, line 172: **Adding `jsdom: ^28` here re-resolved the hoisted copy downward and split the two test environments.**

Hoisted `jsdom` moved **down** 29.1.1 → 28.1.0, pushing `apps/web` onto a nested `web/jsdom@29.1.1`.

The same publish also moved hoisted `@rolldown/pluginutils` **down** from `1.0.1` to `1.0.0-rc.3` — a release candidate — with `rolldown` getting a nested 1.0.1, and swapped hoisted `agent-base` 6.0.2 → 7.1.4 with `@sentry/cli` nested back to 5.0.1.

Pinning `jsdom` to `^29` to match `apps/web` avoids the churn and keeps the two test environments identical.

apps/web/partials/explore/email-capture-account-step.tsx
```diff
@@ -0,0 +1,213 @@
...
+import { useLoginWithEmail } from '@geogenesis/auth';
+
+import * as React from 'react';
+
+import cx from 'classnames';
+
+import { trackPrivyAuth } from '~/core/analytics';
+import { usePrivySignIn } from '~/core/hooks/use-privy-sign-in';
+
+import { SUBTEXT_CLASS } from './email-capture-popup';
```
ohohoreilly: **Circular import between this module and `email-capture-popup`.**

This imports `SUBTEXT_CLASS` from `./email-capture-popup`, which imports `AccountStep` from here. It works today only because the constant is read at render time rather than at module evaluation.

It also means `explore-page.test.tsx:21`'s `vi.mock('./email-capture-popup', …)` leaves `SUBTEXT_CLASS` as `undefined` for any future test that renders `AccountStep` in that suite.

The two shared style constants belong in their own module.

apps/web/partials/explore/email-capture-account-step.tsx
```diff
@@ -0,0 +1,213 @@
...
+          // looking, to the reader, like they pasted the wrong thing. The slice below does the same
+          // job on the normalized value, which is the only place it is correct.
+          pattern="\d*"
+          value={code}
+          onChange={event => setCode(event.currentTarget.value.replace(/\D/g, '').slice(0, CODE_LENGTH))}
+          placeholder="123456"
+          aria-label="Verification code"
+          aria-invalid={failed}
+          disabled={busy}
+          autoFocus
```
ohohoreilly: **`autoFocus` is defeated by `disabled={busy}`, so the reader lands on a field they have to click into.**

On first render `sending` is `false` and Privy's state is `initial`, so the input renders enabled and takes focus. The mount effect then immediately calls `requestCode()` → `setSending(true)`, the input becomes `disabled`, and the browser blurs it. When `sendCode` resolves the input is re-enabled but never re-focused.

That is the opposite of the "fastest path through this step" the comment above it describes.

Focus it in an effect when `busy` goes false, or don't disable the field while the code is being sent.

packages/auth/src/use-ensure-embedded-wallet.tsx
jwalkingjew: Confirmed and fixed — and this is the one I am most glad you caught, because it hits every existing user on every page load, not just the flow under review. `UseWalletsInterface` does expose `ready` and I was reading only `wallets`.

Gated on `ready` now. Test covers your exact window: not-ready with an empty list, assert nothing is asked for, then hydrate with the wallet present and assert it activates and `createWallet` was never called. Removing the gate fails it.

`098950eb4`

packages/auth/src/use-ensure-embedded-wallet.tsx
jwalkingjew: Fair, and the irony is the part that lands — the file argues in its own doc comment that its failures are invisible, then swallows them.

Both paths log now, matching `useSmartAccount`'s `[SMART-ACCOUNT]` convention, and so does exhausting the attempt budget. That last one is the state that actually strands a session, so it was the worst one to keep quiet.

`098950eb4`

apps/web/partials/explore/email-capture-account-step.tsx
jwalkingjew: You are right, and my reasoning was worse than an oversight — I had already found that unconditional tracking and written it up, but only reasoned about the *fallback* branch. It fires on the happy path too, so every signup through this flow was reported twice.

Removed the tracker here. The `link_source` that would distinguish this flow is not worth an inflated signup count, and getting it honestly means the navbar arming its own tracking the way `usePrivySignIn` does — a change to the path every existing user signs in through, which belongs on its own rather than smuggled in here.

Test asserts the hook is called with no arguments, so a tracker cannot come back unnoticed.

`098950eb4`

bun.lock
jwalkingjew: Good catch, and the knock-on effects were not something I looked at. Pinned to `^29.1.1` to match `apps/web` — one hoisted `jsdom@29.1.1` now, no nested copy, so the two environments are identical.

The `@rolldown/pluginutils` slide to `1.0.0-rc.3` came from `@vitejs/plugin-react`, which I added out of habit and the config never used. Removed, and that churn and the `agent-base` swap went with it.

`098950eb4`

apps/web/partials/explore/email-capture-account-step.tsx
jwalkingjew: Correct on both counts, including why it works today — the constants are read at render rather than at module evaluation, which is exactly the kind of thing that holds until an import order changes.

They live in `email-capture-styles.ts` now, imported by both. That also closes the `explore-page.test.tsx` trap you describe, where mocking the popup would leave them `undefined`.

`098950eb4`

apps/web/partials/explore/email-capture-account-step.tsx
jwalkingjew: Confirmed by reading the sequence you describe, which is exactly right: enabled for one frame, focused, then disabled by the send that mount starts, and nothing re-focuses it when that resolves.

`autoFocus` cannot express "focus when this becomes usable", so it is an effect on `busy` going false instead, with a ref on the input. Test asserts the field holds focus once the step is ready; removing the effect fails it.

`098950eb4`

jwalkingjew:
Thanks — this was a genuinely better review than either of the automated rounds, and two of the seven would have reached production.

All seven addressed in `098950eb4`. The two that mattered most:

**The `ready` gate.** This one is not about the flow under review at all — it fires for every existing user on every page load, because a restored session flips `authenticated` before `wallets` hydrates. An account with a wallet looks wallet-less, `createWallet` is called, Privy errors, and a slow hydration could spend the whole retry budget before its own wallet turned up.

**The double-counted signups.** Worse than an oversight on my part: I had already found the navbar's unconditional tracking and written it up — but reasoned about it only for the fallback branch. It fires on the happy path too. Every signup through this flow was being counted twice, inflating exactly the number the flow exists to move.

On **CI**: you are right that the suite was wired to nothing. Filter widened to include `@geogenesis/auth`, and I ran the exact pipeline command rather than trusting the change — `4 successful, 4 total`, both packages. (And thanks for noting the `vitest` → `vitest run` fix; that one would have hung a runner silently.)

The **lockfile** analysis I would not have found on my own. `jsdom` is pinned to `^29.1.1` to match `apps/web`, and the `@rolldown/pluginutils` slide to a release candidate came from `@vitejs/plugin-react` — which I had added out of habit and the config never used. Removing it took the `agent-base` swap with it.

494 files, 5892 tests in the app plus 12 in `packages/auth`. Typecheck and both builds clean.

ohohoreilly (commented):
Re-review at `098950eb4`, after the fixes for the previous round. All seven earlier findings are addressed — the hydration gate, the logging on both failure paths plus a new give-up log, the CI filter, the extracted `email-capture-styles.ts`, the lockfile unwind, the analytics de-dup and the focus handling.

This pass was run at higher effort and verified by execution rather than by reading:

- `next build` clean — so the new `useSearchParams()` call sites pulled in via `usePrepareOnboarding` do **not** break prerendering.
- `apps/web` vitest suite: exit 0.
- `packages/auth` vitest suite: 12 passed; `tsc` build exit 0.
- Checked the `packages/auth/tsconfig.json` `"exclude"` override empirically — it replaces TS's defaults, so `node_modules` is nominally un-excluded, but TypeScript 6.0.3 still skips it for wildcard includes. Not a bug; dropped.

Six findings below. Finding 1 is the one I would not ship without, and findings 2 and 4 both have signup-loss consequences.

Two nits not worth their own threads: the `EmbeddedWalletSync` import in `providers.tsx:11` sits above `./analytics-user-identifier`, outside the repo's `@trivago` sort order (no CI format check, so harmless), and `use-privy-sign-in.ts:6-7` has a stray double blank line left by the removed imports.

packages/auth/src/use-ensure-embedded-wallet.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const addressToActivate = walletToActivate?.address;
+  const attemptsForAddress = addressToActivate ? (activateAttempts[addressToActivate] ?? 0) : 0;
+
+  useEffect(() => {
+    // `walletsReady` is the guard that stops this firing at everybody who already has a wallet.
+    // A restored session sets `authenticated` before `wallets` has hydrated, so on every page load
+    // there is a window where the list is empty and the user looks wallet-less. Asking then means
+    // `createWallet` rejecting -- Privy errors when one already exists -- and if hydration outlasts
+    // the retry delay, an existing account can burn all three attempts before its wallet appears.
+    if (!authenticated || !walletsReady || embeddedWallet || createAttempts >= MAX_ATTEMPTS) return;
```
ohohoreilly: **The hydration gate closes the page-load window but not the fresh-login one, so this still races Privy's own `createOnLogin`.**

`apps/web/core/wallet/privy.tsx` sets `embeddedWallets.ethereum.createOnLogin: 'all-users'`. Now that this hook is mounted app-wide in `providers.tsx`, an ordinary navbar/modal login hits a window where `authenticated` is true, `useWallets().ready` is **also** true, and the wallet Privy is provisioning simply has not landed yet. The hook fires its own `createWallet()` concurrently.

`walletsReady` guards hydration on page load. It does not guard provisioning on a fresh login.

- Best case: Privy rejects the duplicate, the hook logs `[embedded-wallet] could not create a wallet` (now Sentry noise) and burns one of three attempts **on every login**.
- Worst case: the account ends up with two wallets whose `walletClientType === 'privy'`, and `useSmartAccount` picks with `wallets.find(w => w.walletClientType === 'privy')` (`use-smart-account.ts:32`) — which can select the one the user's existing smart account and personal space were **not** derived from.

Separately, line 96 has no in-flight flag. The cleanup's `cancelled` suppresses the error handler but does not stop the request, so any dependency change while a create is pending — `walletsReady` toggling after a wallet-list refresh, for instance — issues a second concurrent `createWallet()` with `createAttempts` still 0.

Suggest gating on `user.linkedAccounts` containing an embedded wallet rather than only on the connected-wallet list, plus an in-flight ref.

apps/web/core/hooks/use-prepare-onboarding.ts
```diff
@@ -0,0 +1,85 @@
...
+        const search = searchParams?.toString();
+        setPostOnboardingRedirect(returnTo === null ? null : (returnTo ?? `${pathname}${search ? `?${search}` : ''}`));
+      }
+
+      setName('');
+      setTopicId('');
+      setAvatar('');
+      setSpaceId('');
+      setStep('start');
+      setSelectedTopicIds([]);
```
ohohoreilly, line 88: **The new `setSelectedTopicIds([])` reaches two callers that fire while the user is already authenticated, and this PR widens the window in which that happens.**

`use-ranking-compose-access.ts:66` and `sign-in-prompt.tsx:50` both gate on `!smartAccount`, not on `!authenticated` (`use-ranking-block-state.ts:667`, `sign-in-prompt.tsx:27`).

A brand-new account from the email-capture flow is `authenticated` with **no** smart account until `useEnsureEmbeddedWallet` has created one and wagmi has picked it up — up to 3 attempts x 2s. If the user is in the onboarding dialog with interests already picked, and anything triggers the sign-in prompt or a ranking-compose link inside that window, `prepareOnboarding({ keepReturnTo: true })` now also wipes `selectedTopicIdsAtom`.

`PendingPersonalSpaceRunner` (`pending-personal-space-runner.tsx:58`) then submits no membership proposals for picks the user actually made.

The name/avatar/step reset was pre-existing. The topic wipe is new in this PR.

apps/web/partials/explore/email-capture-popup.tsx
```diff
@@ -173,7 +204,16 @@ function EmailCapturePopup() {
...
-  if (closed || !ready || authenticated || !scrolledEnough || anOverlayIsOpen) return null;
+  if (closed || !ready || authenticated || !scrolledEnough) return null;
+
+  // An overlay normally takes the card off the screen entirely. Not once a code has been sent:
+  // returning `null` unmounts the step, and mounting is what sends a code — so opening search or
+  // the chat panel mid-sign-up and closing it again would mail a second code and silently retire
+  // the one the reader was part-way through typing. Hidden instead of unmounted, so the attempt
+  // survives. `hidden` takes it out of the layout, the hit-testing and the accessibility tree, so
+  // it still yields the screen completely; it just does not forget where it was.
+  const hiddenByOverlay = anOverlayIsOpen && wantsAccount;
```
ohohoreilly, line 229: **The `hidden` fix covers overlays only; a route change still strands a live OTP with the card permanently dismissed.**

`hiddenByOverlay` keeps `AccountStep` mounted while `anOverlayIsOpen`, but a client-side navigation away from Explore unmounts `EmailCapturePopup` entirely, losing `wantsAccount` and `subscribedEmail`. Because `rememberDismissed()` already ran at subscribe (line 186), returning to Explore shows nothing.

Reader subscribes, clicks Create account, the code arrives, they click a link that navigates — and they now hold a valid Privy code with nowhere to type it and no way to bring the card back.

apps/web/partials/explore/email-capture-account-step.tsx
```diff
@@ -0,0 +1,223 @@
...
+  // button that leads here — which is what broke that button.
+  //
+  // The embedded wallet this login needs is not created here. It cannot be: this component is
+  // unmounted by the card's own visibility rule the instant `authenticated` turns true, which is
+  // the exact render in which the wallet becomes creatable. `useEnsureEmbeddedWallet`, mounted for
+  // the life of the app in `core/providers.tsx`, does it instead.
+  // Deliberately not tracking here, which is a correction rather than an omission.
+  //
+  // `useLoginWithEmail` and `useLogin` subscribe to the same Privy `login` event, and the navbar
+  // renders `GeoConnectButton` for every logged-out reader (`navbar-actions.tsx`, `if (!address)`)
```
ohohoreilly: **The "do not register a second tracker" decision depends on the navbar's `GeoConnectButton` being mounted at completion, and it is not always.**

The comment here, and the test at `email-capture-popup.test.tsx:756`, justify passing no callbacks to `useLoginWithEmail` on the grounds that the navbar's `useGeoLogin` tracks unconditionally.

But `navbar-actions.tsx:70` returns a skeleton — not `GeoConnectButton` — whenever `isUserLoading`, and that file's own comment notes `isUserLoading` "flips back to true mid-session (…a tab refocus or Privy re-init re-resolves it)".

If the navbar is in that state when Privy emits `login`, **nothing records the event at all** — `usePrivySignIn`'s tracker is disarmed by design. The failure mode is silent under-counting of exactly the signups this flow exists to produce.

packages/auth/index.ts
```diff
@@ -1,6 +1,12 @@
 export {
   usePrivy,
   useLogout,
+  // Privy's headless email login — `sendCode`/`loginWithCode` against an address already in hand,
+  // so someone who has just typed their email into another form is not asked for it twice
+  // (GEO-2948). The wallet wiring a modal login performs is handled app-wide by
```
ohohoreilly: **The comment and its test assert an invariant the code does not provide.**

`useLoginWithEmail` is re-exported from `@privy-io/react-auth` unchanged. It is *not* wrapped the way `useGeoLogin` (`src/use-login.tsx`) wraps `useLogin` with `setActiveWallet`.

The test "uses the login hook only once an account is asked for" pins the import path and nothing more; its stated rationale about "swapping back to the raw hook" describes a distinction that does not exist here. The real fix for the missing wallet is `EmbeddedWalletSync`.

As written, a future reader could conclude the wallet wiring lives in this re-export and delete `EmbeddedWalletSync`, reintroducing the exact bug this PR was opened to fix.

packages/auth/package.json
```diff
@@ -32,7 +32,15 @@
   },
   "devDependencies": {
     "prettier": "^3.8.1",
-    "typescript": "^6.0.2"
+    "typescript": "^6.0.2",
+    "react": "^19.2.8",
+    "@types/react": "^19.2.8",
+    "vitest": "^4.1.10",
+    "jsdom": "^29.1.1",
+    "@testing-library/react": "^16.3.0"
```
ohohoreilly: **`react-dom` is undeclared but required, and CI now depends on this suite.**

`@testing-library/react@16` imports `react-dom/client`. This repo uses bun's isolated linker (`node_modules/.bun`), and after a fresh `bun install --frozen-lockfile` I confirmed `packages/auth/node_modules` contains no `react-dom`. The suite passes today only because bun auto-installs the unmet peer.

Now that `ci.yml:64` gates on this suite, add `react-dom` and `@types/react-dom` to `devDependencies` so the gate cannot go dark on an install-config change.

packages/auth/src/use-ensure-embedded-wallet.tsx
jwalkingjew: Correct on both counts, and the wrong-wallet case is the one that would have hurt — `useSmartAccount` picking with `.find()` between two `walletClientType: "privy"` wallets could land on the one the account's smart account and personal space were never derived from.

Gated on `user.linkedAccounts` now, as you suggested: it is authoritative about what the *account* has, and arrives with the user object rather than with the connected list, so it is true through exactly the provisioning window `walletsReady` leaves open.

The in-flight ref is in too. You are right that `cancelled` only silences the handler — the request keeps going, so a dependency change mid-create started a second one with `createAttempts` still zero.

Three tests: a linked-but-unconnected wallet is not re-created, an account with genuinely none still gets one, and a wallet-list refresh mid-create does not start a second. Each fails with its guard removed.

`4e8fbb2d6`

apps/web/core/hooks/use-prepare-onboarding.ts
jwalkingjew: This one is a regression I introduced, and your read of why is exactly right — both callers gate on `!smartAccount` rather than `!authenticated`, and a brand-new account from this very flow sits authenticated with no smart account for as long as creation and activation take. My own feature widened the window its own reset then fires in.

The reset no-ops for an authenticated user now, and I made it cover the whole reset rather than just the topics. The leak it exists to stop requires a *different* person arriving on the browser, which means not authenticated — and `login()` is a no-op for someone already signed in, so there is no sign-in for the cleared state to belong to. That also protects the pre-existing name/avatar/step wipe you noted.

`4e8fbb2d6`

apps/web/partials/explore/email-capture-popup.tsx
jwalkingjew: Confirmed and fixed. `rememberDismissed()` runs at subscribe and the `status === "done"` exception is component state, so a navigation left `dismissed` true with nothing able to bring the card back — a valid code and nowhere to type it.

The attempt is kept in `sessionStorage` for roughly the life of a Privy code and resumed straight into the code step, skipping the scroll gate. Closing clears it, signing in clears it, and a stale one expires rather than offering a field for a code that no longer works. Three tests, each failing with its piece reverted.

`4e8fbb2d6`

apps/web/partials/explore/email-capture-account-step.tsx
jwalkingjew: You are right, and this is the third position I have taken on this — worth saying plainly. I added a tracker, removed it as a double count, and you have found the case that makes removing it wrong: the navbar is a skeleton whenever `isUserLoading`, which its own comment says flips back mid-session, and a completion landing there was recorded by nobody at all.

Fixed at the cause rather than by picking a side: the navbar arms its tracker the way `usePrivySignIn` does, so it reports the logins it starts and stops reporting session restores as manual ones. This flow then reports its own with `link_source`. No double count and no gap, and the arming fixes a pre-existing over-count too.

`4e8fbb2d6`

packages/auth/index.ts
jwalkingjew: Fair, and the danger you name is the real one — a reader concluding the wiring lives in the re-export and deleting `EmbeddedWalletSync` would reinstate exactly the bug this PR was opened for.

The comment says what is actually true now: re-exported unchanged, adds nothing, creates no wallet and activates nothing, and `EmbeddedWalletSync` is the only thing making an authenticated session end up with a usable wallet. I also rewrote the test's name and rationale, which described a distinction that does not exist.

`4e8fbb2d6`

packages/auth/package.json
jwalkingjew: Confirmed — `packages/auth/node_modules` has no `react-dom`, and you are right that a CI gate standing on bun auto-installing an unmet peer is not a gate.

Added `react-dom` and `@types/react-dom` to `devDependencies`, plus `react-dom` as a peer alongside `react`, since the package genuinely needs both.

`4e8fbb2d6`

PR #2457, opened 2026-09-18
feat(debates): availability calendar behind "Set my schedule" (GEO-2936)

o-p-o-p-o:
A week grid a person drags their availability onto, opened from a new callout at the top of the debates panel. Ported from the GEO-2936 interaction prototype, which settled the snap granularity and the three-kind model.

**This is the client half only.** Nothing is persisted: the availability endpoints carry the online toggle (`/me/debate-availability`) and nothing else, so Save holds the schedule for the session and no further. The payload the client *would* send is in `toPayload`, and the debug page below renders it live — worth agreeing on that shape before the backend is written.

## What is here

| | |
|---|---|
| `core/availability/blocks.ts` | The model and its rules: three block kinds, 30-minute snapping, merging, exception subtraction, slot expansion, payload. No React. |
| `partials/availability/availability-calendar.tsx` | The grid: drag to create, drag to move, drag the edges to resize, hover for the ×. |
| `partials/availability/availability-modal.tsx` | The dialog it opens in. Wider than the shared `Dialog` allows, so it drives the Radix primitives directly, as the debate share sheet does. |
| `core/debates/matchmaking/set-schedule-banner.tsx` | The lavender "Set your debate schedule" callout and its button, between the panel heading and the tabs. Dismissal reuses the explore banner's. |
| `app/space/[id]/(space)/debug-availability` | Somewhere to exercise it, beside `debug-debates`: the grid, the payload, and the slots the other person would be offered. |

Two one-line registrations: the route list in `core/utils/space-url.ts` (filesystem-checked by its own test — the route 404s without it) and the crawler exclusion in `app/robots.ts`, matching `debug-debates`.

## Three block kinds

Recurring weekly, this date only, and unavailable. The third is what makes "free Tuesdays, but not next Tuesday" sayable; without it that can only be said by deleting the recurring block and putting it back, which is the main way availability goes stale and sends invitations nobody can accept. `effectiveAvailability` is the only place the three resolve against each other, so the grid, the slot list and anything matching on availability cannot disagree.

## Interaction decisions worth a look

- **The mode buttons are the legend.** Each carries the colour its blocks are drawn in, so the colours are named once.
- **Flooring, not rounding.** A press takes the cell it is inside. Rounding to the nearest boundary meant the bottom pixels of the 8:30 cell created a 9:00 block.
- **Midnight to midnight**, opening scrolled to 7am. Night shifts and other time zones are availability too.
- **Overlaps step right 10px**, as Google Calendar does, and the block being dragged draws above them — a preview underneath the block it overlaps is invisible until it is let go of.
- **Touch requires a held press** (350ms). The grid shares its surface with scrolling, and a tap that created a block would fill the schedule in while someone was looking around. `touch-action` stays put for the whole gesture and the pan is suppressed by a non-passive listener; switching `touch-action` mid-gesture cancels the pointer, which ended the drag the moment the hold fired.
- **Mobile shows ~3.5 days** in a horizontal scroller that carries the headings with it.

## Testing

- 31 unit tests over the model, 5 over the grid, 4 over the banner; 633 pass across `core/availability`, `partials/availability`, `core/debates/matchmaking` and the route registry.
- Checked in Chrome against testnet: dragging, moving, resizing, deleting, clearing, the week arrows and Today, the payload and slot panels.
- Checked on an emulated iPhone 13: 3.47 days in view, scrolls both ways, tap and swipe leave the schedule alone, a held press draws.
- Not checked on a real phone — that is what this preview is for.


ohohoreilly:
I have merged `master` in and wired the calendar to the backend, since the half this branch was waiting on landed today. Pushed as `9d2ad6f7`. **Merged, not rebased**, so nothing here is rewritten and anything local of yours still pulls cleanly.

Your own comment called this out:

> Where the schedule lives until there is somewhere to send it… the wiring lands with the backend half of GEO-2936.

That backend is now on testnet, so this is that wiring.

## The conflict, which was not a textual one

`z-layers.ts` was the only clash, and neither side was right on its own. This branch chose **10003/10004** for the dialog; master has since moved `statusBar` to **10004** and `toast` to **10005**. Taking your numbers verbatim would have put the calendar *below* the status bar and the toast — the exact thing your comment says the layer exists to clear, and it would have looked fine in review.

Renumbered to **10006/10007**, with the class map to match, so the stacking intent survives master's renumbering.

## What the wiring adds

- **`fromPayload`**, inverting `toPayload`. Two things have to be undone exactly: the wire's `weekday` is one-based while the editor's is zero-based, and `exceptions` is *absent* rather than empty when there are none. The first is the dangerous one — get it wrong and every schedule shifts a day with nothing throwing — so it has its own test plus a full round trip.
- **`getDebateSchedule` / `replaceDebateSchedule`** against `/me/debate-schedule`, documented right beside `/me/debate-availability`, because those two paths differ by one word and mean completely unrelated things (calendar vs the "available right now" toggle).
- **`useDebateSchedule` / `useSaveDebateSchedule`**. The query hands back blocks rather than the wire payload, so no component ever handles the one-based weekday itself. The mutation reads the timezone at save time — a schedule means "18:00 where I am" — and takes the server's stored answer back rather than re-deriving it.
- **The banner** reads from the query and saves through the mutation. `blocks` is undefined until the first read answers, which the modal already treats as an empty calendar, so opening it early behaves exactly as before.

## Test changes you should know about

Seven sibling tests in `core/debates/matchmaking` mock `../hooks` wholesale and needed the two new exports added to their mocks. The banner's own test needed them stubbed outright — it rendered the component with no `QueryClientProvider`, which was fine until it started issuing a query.

Verified: **108 files / 1649 tests** pass across `core/debates`, `core/availability` and `partials/availability`; `tsc` clean on the touched areas.

Happy to back any of this out if you would rather wire it differently — particularly the decision to save on **Save schedule** rather than debounce as the user drags.

PR #2483, opened 2026-09-20
fix(api): meter the Mapbox proxies, and close the inet_aton hole in the SSRF block

ohohoreilly:
From an audit of the API surface. Every `chat/*` route checks same-origin, a wallet cookie and a rate limit; **nothing outside `chat/*` has a rate limit at all.** Most of those turned out to be fine on inspection — the cron sweep requires `CRON_SECRET`, the community-call catch-all is a CORS shim whose upstream does its own authz, the newsletter route has two limiters. Two were not.

## 1. The Mapbox proxies were unauthenticated and unmetered

`places/search` and `places/coordinates` are anonymous GETs that spend `MAPBOX_TOKEN`. Mapbox Search Box bills per session and per request, so these were a free metered API anyone could run in a loop at our expense — no origin check, no auth, no limit.

Both now go through a shared guard: same-origin, then one IP-bucketed limit across the pair (a search is followed by a retrieve for the chosen suggestion, so they are two halves of one user action and a shared budget is the honest accounting).

**Refused in production when the limiter is unusable**, matching `newsletter/subscribe` — an unlimited version of a billable anonymous endpoint is not a thing to serve while we work out whether Redis is misconfigured or briefly down. Outside production there is no Redis and no bill, so it carries on. Worth a reviewer's opinion: the alternative is failing open and eating the abuse risk during an Upstash blip.

Being straight about what same-origin buys: a browser sets `Origin` itself so this stops other sites driving these endpoints from a visitor's browser, and it stops nothing scripted. The rate limit is the control that matters.

## 2. `coordinates` injected user input into the upstream path

```ts
`https://api.mapbox.com/search/searchbox/v1/retrieve/${mapboxId}?session_token=${sessionToken}&access_token=…`
```

`mapboxId` went into the **path** unencoded, so a caller could steer the request off `/retrieve/` with `../`, or begin their own query with `?`. `sessionToken` went into the query string unencoded in both routes. Its sibling correctly wrapped its user input in `encodeURIComponent`; this one did not.

Both now build the upstream URL with `URL`/`searchParams`, validate `mapboxId` against its actual shape (it lands in a path, so pinning it beats escaping it) and validate the session token rather than pasting it through.

`search` also mapped straight off `data.suggestions`, which a Mapbox error response does not carry — the resulting TypeError surfaced as our 500 for their 4xx. It now reports a 502 when upstream failed and an empty list when it simply had nothing.

## 3. `isPrivateHost` missed every non-dotted-decimal IPv4 encoding

`new URL()` keeps a hostname as written, and the checks only read dotted-decimal:

| host | is | before |
| -- | -- | -- |
| `2852039166` | 169.254.169.254 | **allowed** |
| `0xA9FEA9FE` | 169.254.169.254 | **allowed** |
| `0251.0376.0251.0376` | 169.254.169.254 | **allowed** |
| `2130706433` | 127.0.0.1 | **allowed** |
| `0x7f000001` | 127.0.0.1 | **allowed** |
| `0177.0.0.1` | 127.0.0.1 | **allowed** |

Three of those are the cloud metadata address. Hostnames that could be numeric are now normalised through the inet_aton rules — one to four parts, hex, octal, decimal, last part absorbing the rest — before the existing checks run.

Public addresses in the same encodings still pass: `0x08080808` and `134744072` are both 8.8.8.8 and both allowed. So are `1e100.net`, `fac.ad`, `0x.ax`, `256.1.1.1` and `099.1.1.1`, which look numeric to varying degrees and are not addresses.

**Not reachable today** — the user-supplied URL goes to Anthropic's webFetch, not a server-side fetch, and `safeFetch` is unused on that path. But this is the defence-in-depth control its own comment says it is ("any future direct fetch path must not be steered at loopback / RFC1918 / link-local hosts"), and it had a hole.

## Housekeeping

`isSameOrigin` is promoted next to `client-ip.ts`, for the reason that file's comment already gives — eleven chat routes and `ranking-og/route-utils` each carried a copy. `ranking-og` re-exports it so its callers are untouched. I did not migrate the eleven chat copies; that is churn for a separate change.

## Noted, not changed

`ranking-og/rate-limit.ts` gates on `UPSTASH_REDIS_REST_URL`/`TOKEN` by name — the exact mistake `newsletter/rate-limit.ts` documents, since Vercel's Upstash integration provisions `KV_REST_API_*` which `Redis.fromEnv()` also accepts. On such a deploy its limiters stay null and OG rendering is unmetered. It fails open deliberately, so this is a gap rather than a break, and changing it is a behaviour decision for whoever owns that route.

## Validation

- 33 new tests — the inet_aton encodings in both directions, and the guard's same-origin / limit / degradation behaviour.
- **6508/6508 across the whole suite** (548 files).
- Typecheck and lint clean.


ohohoreilly:

PR #2484, opened 2026-09-20
fix(web): guard the three storage accesses that were not

ohohoreilly:
From the same audit as #2483. A browser with site data blocked throws `SecurityError` on the `window.localStorage` / `sessionStorage` **accessor itself**, not just on write. Eighteen files in the app touch storage; fifteen already wrap it. These three did not.

## The one that matters

`SearchDialog`'s `readCanonicalOnly` is a lazy `useState` initialiser, so it runs **during render**:

```ts
const readCanonicalOnly = (): boolean =>
  typeof window === 'undefined' || window.localStorage.getItem(SEARCH_CANONICAL_ONLY_KEY) !== 'false';
```

A throw there takes the dialog down rather than losing a preference, so search is unreachable for anyone with site data blocked — Firefox's "block all cookies", Chrome's per-site equivalent.

## The other two

- **`entity-feed.tsx`** reads and writes its type filter in effects. A throw surfaces as a component error on the app's main page. The filter now starts at its default instead.
- **`personal-profile-suggested.ts`** clears per-wallet dismiss flags on disconnect, where a throw would take the sign-out path with it — and a browser that stored nothing has nothing to clear.

## Shape

The codebase already has an idiom for this (`core/space/daily-activities-storage.ts`): `typeof window` for SSR, try/catch for the browser refusing, and a sensible default rather than a crash. I followed it rather than introducing a helper the other fifteen files do not use.

## Validation

6475/6475 across the suite, typecheck and lint clean. No new tests: these are catch-and-default paths around a browser behaviour the suite has no way to produce, and a test asserting "we call localStorage in a try" would assert the implementation rather than anything real.


ohohoreilly:

PR #2485, opened 2026-09-20
fix(web): two components called a hook after an early return, and nothing was checking

ohohoreilly:
## Nothing has been checking React hooks in this app

`eslint.config.mjs` extends `eslint-config-next/typescript` — the TypeScript subset, which carries no react-hooks rules. Printing the resolved config for a file confirms it rather than inferring it:

```
react-hooks rules resolved: NONE
total rules active: 23
```

That is worth sitting with given what this codebase has been debugging lately. A resume writing through a stale closure (GEO-2783). An effect discarding good URLs every time a card re-activated (GEO-2895). A fetch claiming a key it never released (GEO-2950). All of them the class `exhaustive-deps` exists to catch, in an app where the rule has never run.

Running the rules over all 2,182 files: **4 `rules-of-hooks` errors, 93 `exhaustive-deps` warnings across 53 files.**

## The two real bugs

**`CommentList`** returns early when `depth === 0`, then calls `useCommentDensity()` 35 lines later. `depth` is a prop, so a list rendered at one depth and then another calls a different number of hooks and React throws.

**`PowerToolsScreen`** returns early for a Relations data source, then calls `React.useMemo`. `data.sourceType` changes when the block's source does, which is a live user action, so this one is more than theoretical.

Both fixed by hoisting the hook above the return. Both are cheap enough to run on every render, so nothing is lost by doing so.

## The two false positives, now explicit

- A TanStack `cell` **is** mounted as a component — `flexRender` `createElement`s a function definition — so `useSpaceAwareValue` there is legal. The rule only sees a lowercase property called `cell`. That site carries a disable with the reason.
- A test `wrapper` **is** a component, passed to `renderHook({ wrapper })`. Renamed `Wrapper` so the rule can recognise it, rather than suppressed.

Neither was silenced blind; both were checked before being dismissed.

## Why the rules are set where they are

`rules-of-hooks` as **error** — there is no such thing as an acceptable violation, and the count is now zero.

`exhaustive-deps` as **warning**, deliberately. 93 existing violations across 53 files: as an error it would either block every PR or force a 93-item sweep by whoever happens to be next, on code they did not write. As a warning it stops the count growing in code being written now, and someone can burn down the backlog by area with context. `eslint .` passes no `--max-warnings`, so CI is unaffected either way — verified below.

The worst offenders, if anyone wants to start: `debate-room-page-client.tsx` (12), `table-block-filter-creation-prompt.tsx` (5), `review-changes.tsx` (4).

## Dependency

`eslint-plugin-react-hooks` is added as a devDependency because bun's isolated install does not hoist it out of `eslint-config-next`, so a config importing it by name fails to resolve. One line in `package.json`, one net line in `bun.lock`.

## Validation

- `eslint .` → **0 errors**, 586 warnings (93 of them the new rule).
- **6475/6475** across the suite, typecheck clean.


ohohoreilly:

PR #2486, opened 2026-09-20
chore: ignore the local-env family at the repo root, not just in apps/web

ohohoreilly:
`apps/web/.gitignore` ignores the local-env family:

```
.env.local
.env.development.local
.env.test.local
.env.production.local
```

The **root** `.gitignore` only listed `.env`. Running turbo from the repo root is the normal way to work here, and that is where a root `.env.local` ends up — **untracked but not ignored**, so `git add .` or `git add -A` would stage it along with whatever credentials are in it.

Found during an audit: there is one sitting in my working tree right now, showing as `??` rather than being ignored, which is how I noticed.

One-line class of change, no behaviour attached — it just closes the gap between the two files.


ohohoreilly:

PR #2494, opened 2026-09-20
fix(power-tools): declare a prop the apply callback reads

ohohoreilly:
Fifth slice of the `exhaustive-deps` backlog. Seven warnings across power tools, and one is a real bug.

## The bug: an apply that can use the wrong scope

`applyNewProperty` passes `applyToAllEntities: newPropertyOnly` and never declared `newPropertyOnly`, so the callback holds whatever that prop was when its other dependencies last changed.

The component treats the prop as changeable — there is an effect keyed on it alone a few lines below — so a change while the popover is open leaves the apply running against the previous scope. It is a boolean prop; declaring it costs nothing.

## An effect running every render

`options?.extraColumnIds ?? []` minted a new array each render, re-running the effect keyed on it every time. Named the empty case — the `excludedColumnIds` memo immediately below already avoids exactly this with a `useMemo`.

## Four that are deliberate, now stated

Each array carries `source.type` plus `sourceValue`, the narrowed `'value' in source ? source.value : null`, which between them cover everything about the source that matters. Depending on `source` itself is a hazard the codebase already documents, in `core/blocks/data/source.ts`:

> Identity for React effect deps — `getSource` returns new object literals each render, so comparing `source` by reference loops effects that call setState.

So the rule is asking for the thing that caused a previous bug. Disable with the reason and a pointer to that comment.

## One snapshot

The remaining effect snapshots `initialPropertiesMarkedForRemoval` when the popover opens, on purpose. Following the prop would re-snapshot whenever it changed and discard whatever the reader had ticked since opening.

## Coverage

**Power tools has no tests — zero files under `partials/power-tools`.** The suite passing says nothing about the apply-scope change, so I am verifying that one in the Vercel preview rather than leaning on green CI.

## Validation

- Power tools at **zero**; **81 → 74** against master.
- **6526/6526**, `tsc` clean, `eslint .` 0 errors.


ohohoreilly:
**Verified in the preview, not just in CI.** I said in the PR body that power tools has no tests so the suite proves nothing here — so I loaded the screen and drove it.

Power tools does not need auth to *read*, it just needs the URL parameters. Found a real data block through the API and opened it:

```
/space/89bd89bf28ff8a0963faf92a8c905e20/47cd90d4fe574f849aa1c1074d6b25c8/power-tools
  ?relationId=ebe02eea7f3245c7969402a86a0ebd0e
```

That renders the Claims table in World affairs with real rows and columns — which means `use-power-tools-data` and `power-tools-screen`, the two files this PR changes, are both actually running.

**The risk with these changes is a render loop**, since they alter memo and effect dependencies. Measured rather than eyeballed, by counting resource entries:

| | requests | after 6s | delta |
| -- | -- | -- | -- |
| idle after load | 216 | 216 | **0** |
| after typing a `Name is NATO` filter | 233 | 233 | **0** |

The filter drives new state through `source` and the data hook — 17 requests to apply, then completely settled. If `extraColumnIds` were still minting a new array per render, or a `source.value` dependency had been added and started the documented setState loop, this is where it would show.

Console across load and interaction: two wallet messages (`chains are not supported by Coinbase Smart Wallet`, `Detected injected providers`), both unrelated and present regardless. No React warnings, no "Maximum update depth exceeded".

**What I did not verify:** the `newPropertyOnly` fix itself. Reaching the edit-entities popover needs edit mode, which means signing in as an editor and writing to the graph — not something to do casually on a shared space. That one still rests on reading the code: the callback passes `applyToAllEntities: newPropertyOnly` and did not declare it, and the component has an effect keyed on that prop alone, so it can change while the popover is open.

PR #2495, opened 2026-09-20
perf(debates): two arrays rebuilt every render, and eight dependencies that needed saying

ohohoreilly:
Sixth slice of the `exhaustive-deps` backlog. Ten warnings across debates — the area with the bug history, and the one where the rule would have helped most. Two are real; eight are correct code that reads as an oversight.

## The two real ones

**`debate-feed`** built `visibleDebates` with an unmemoised ternary:

```ts
const visibleDebates = anchorPending ? [] : debates.slice(0, visibleCount);
```

Both branches produce a new array every render, and the effect keyed on it re-ran every render of the feed.

**`use-debate-playback`** did `transcriptQuery.data?.segments ?? []`, so the subtitle memo recomputed on every render of a playing video — which is every frame the playhead advances.

## Free, and now declared

The Jotai setters in `claims-tab` (stable by Jotai's contract), and `awaitingTap` in `debate-feed-player` — which is exactly `userPaused || autoplayBlocked`, both already in that dependency list. It replaces them rather than joining them, so the list now names what the effect reads.

## Deliberate, and now carrying the reason

These are the interesting ones, because taking the rule's advice would break three of them:

- **`collapse-answered`** runs with no dependency list on purpose and already explained why in a comment: the callbacks come from the caller, so a list would either churn on inline ones or go stale on memoized ones. The body is idempotent and `setHolding` bails on identity, so it cannot loop. The rule's suggested list would be worse.
- **`use-stable-list-order`** keeps `resetKey` even though the memo never reads it. It is consumed during render to clear `orderRef`, and the memo must recompute against the cleared order. **Dropping it as "unnecessary" would return a stale order on every reset** — a bug introduced by fixing a warning.
- **`recording-page-client`** omits the react-query mutation object, whose identity changes across renders. Depending on it would re-fire the mutation on each one.
- **Both `hooks.ts` effects** depend on `query.refetch` rather than `query`. The query object is rebuilt whenever its data changes, so depending on it would re-run the effect on every refetch it had just caused.

## Validation

- Debates at **zero**; **71 → 61** against master.
- **6526/6526**, `tsc` clean, lint 0 errors.
- Verified in the preview as well — see the comment below. This touches playback, and the suite passing is not the same as a video playing.


ohohoreilly:
**Verified in the preview.** This touches playback, and a green suite is not the same as a video playing — so I drove the explore feed rather than relying on CI.

Checked the tab was actually visible and focused first (`visibilityState: "visible"`, `hasFocus: true`) — a backgrounded tab makes media look broken and has produced a wrong diagnosis on this surface before.

**Playback is real and smooth.** Four `<video>` elements at `readyState: 4`; the on-screen pair playing, the off-screen pair correctly idle at 0. Over eight seconds of wall clock both advanced **exactly 8.0s** — no stutter, no restart to zero.

**No render loop**, which is the risk these changes carry. Counting resource entries:

| | requests | +8s | delta |
| -- | -- | -- | -- |
| playing, idle scroll | 250 | 250 | **0** |
| after scrolling the feed | 250 | 250 | **0** |

Zero in both windows. That is the direct test of the two real fixes here: an unmemoised `visibleDebates` would re-run its effect on every render, and a `transcriptSegments` array rebuilt each render would recompute the subtitle memo on every frame the playhead advances. Scrolling is the case that changes `visibleCount`, so it is the one that would expose the feed memo, and it stays flat.

Feed rendering is intact after scroll — debate cards with video, vote bars and percentages, claims and news stories all present.

Console across load, playback and scroll: two wallet messages (`chains are not supported by Coinbase Smart Wallet`, `Detected injected providers`), both unrelated and present regardless of this branch. No React warnings, no "Maximum update depth exceeded".

PR #2511, opened 2026-09-21
refactor: share media query subscriptions

jwalkingjew:
Split from #2465 so the mobile-navbar feature stays scoped.

## What changed

- Adds one shared `useMediaQuery` hook.
- Reuses it in the two pre-existing viewport hooks for app layout and call layout.
- Adds focused subscription, update, and cleanup coverage.
- Contains no navbar or account-menu behavior.

## Validation

- Focused `useMediaQuery` test passes.
- Targeted ESLint passes.
- TypeScript no-emit check passes.
- Hosted build, lint, typecheck, full tests, and Vercel deployment pass.


ohohoreilly:
Keep this one. It's behaviour-preserving and the duplication was real — two byte-identical `subscribe` / `getSnapshot` / `getServerSnapshot` triples, now one hook, with focused coverage of subscribe, update and cleanup.

Nothing I'd change. The added `typeof window === 'undefined'` guards are redundant given `getServerSnapshot` already handles SSR, and `getSnapshot` still constructs a `MediaQueryList` on every call rather than reusing one — but both of those are inherited from the code it replaces, so neither is a regression this introduces.

One cosmetic thing: the description has literal `\n` characters in it instead of newlines, so it renders as a single run-on line.

jwalkingjew:
Addressed the cosmetic issue: the PR description now uses real newlines and renders as separate sections. I also revalidated the code recommendation after updating to current master. I agree no hook change is warranted here—the SSR guards and repeated matchMedia lookup are inherited behavior, not regressions introduced by this refactor. Focused test, targeted ESLint, and TypeScript no-emit all pass.

PR #2526, opened 2026-09-22
feat(claims): lay the claim page out like the Explore claim card, with a Topics tab

o-p-o-p-o:
The claim page's hero now uses the Explore claim card's layout at page scale, so a claim looks the same wherever you meet it. Topics move to their own tab. Most of the change is to the claim page; the shared components that gain options are listed under **Outside the claim page**.

## Hero
- **Layout:** the Explore claim card's two zones. On the left are the claim, its description and the Agree/Disagree pills (capped at 360px, as on Explore), with the debate offer under the pills. On the right is Explore's own 220px verdict column (share, bar and the two sides) behind a full-height rule.
- **Alignment:** the verdict column starts on the title's row, so the share lines up with the claim's first line.
- **Narrow widths:** under a 520px container (`claim-card-narrow`, the card's own breakpoint) it stacks the way the card does, with the verdict below the pills and no rule. The hero is its own `@container`, so the switch follows the hero's width, not the window's. That's what matters in the side panel.
- **Unanswered claims:** a claim with no responses has no verdict column and takes the full width, as on Explore. The page's "No responses yet" invitation goes with it.
- **Chips:** the type and tag chips are gone. Controversial, when it applies, is a red chip above the title, in the chips' size and type.
- **Title:** unchanged from `master` (24px, 28px from a 560px column).
- **Spacing:** 48px from the hero to the tabs at every width. Below 560px that used to be the page's 24px gap.
- **Width:** the column is 840px, for 800px of content. The cover header shares the constant and widens with it.

## Topics
- **Topics tab:** a new tab at `/topics`, after Related claims. It lists every topic, uncapped, and only appears when the claim has topics, the same way Sources does. It uses the shared `ClaimRecordPage` like the other record routes.
- **Overview:** no longer repeats Topics. It holds Activity and Comments.
- **Hero topics row:** a row of up to five topics above the claim, with "See all" leading to the tab, is built but switched off (`SHOW_HERO_TOPICS = false`) while we decide on it. In the side panel, "See all" switches the panel's tab rather than navigating the page behind.

## Responder list (the popover under "6 agree")
- **Title:** "6 people agree" above the rows, replacing the "6 people · Agree" footer. The verb agrees with the count ("1 person agrees").
- **Order:** people with a profile picture first. The order within each group is unchanged.
- **Hover:** rows highlight like the browse sidebar's, with `rounded-lg` and `hover:bg-grey-01` inside a 4px inset.

## Debates tab
- **Width:** a debate fills the content column rather than the card's viewport-fitted cap of up to 560px. The two stacked videos make the card tall, so it no longer fits one screen at shorter window heights.

## Activity section
- **Header from #2514:** brings in the Activity header restyle from #2514: Debates/Claims pills, a "View all" pill, header arrows and 12px corners on the compact player. It's taken as-is from that branch, so whichever of the two PRs merges second will have nothing to do in these files.

## Outside the claim page
- **`ClaimVerdictColumn`:** now exported from `claim-explore-feed-card.tsx` for the hero. Nothing else in that file changes.
- **`ClaimVerdict` removed:** `claim-verdict.tsx` and its tests are deleted. The claim page was its only caller.
- **`RelationChipSection`:** gains `cap`, defaulting to the existing 8. The Topics tab passes `Infinity`.
- **`ClaimSideResponders`:** the title, sort and hover changes apply everywhere the list opens, so the Explore claim cards get them too.
- **`MemberRow`:** gains an optional `className`. The space members and editors lists don't pass one and are unchanged.
- **`DebateExploreFeedCard`:** gains `fullWidth`, passed through `ExploreFeedCard` (`fullWidthDebate`) and `PersonRecordFeed` (`fullWidthDebates`). All off by default; only the claim page's Debates tab turns it on.
- **`button.tsx` and `debate-feed-player.tsx`:** these changes come with the #2514 Activity header.

## Checks
- `tsc --noEmit` is clean.
- ESLint and Prettier are clean on every changed file.
- The full web suite passes: 589 files, 6,959 tests. The claim page tests cover the hero order (claim, pills, verdict, tabs), no verdict column on an unanswered claim, the Topics tab (uncapped, only topic relations, offered only when there are topics) and the Overview no longer showing topics.
- Checked in a desktop browser at 1440px against testnet data. The share and the title's first line both measure y=144, and the hero is 800px. Not yet checked at phone widths or in the entity side panel, where the stacked arrangement takes over.


jwalkingjew:
Closing this as superseded by #2530, which carries this work forward rather than replacing it.

#2530 started as a branch on top of `claim-page-hero` — it takes the Topics tab this PR added and turns it into an explore feed of topic cards, so it depends on the route, the tab registration and the `ClaimTabPanel` branch from here. Rather than keep the two stacked, we've folded this in. **All four commits are preserved with their original authorship**, rebased onto current `master`.

Two things changed in the move, both worth knowing about:

**The Activity-header conflict is resolved in favour of what merged.** This PR took the Activity header from #2514's branch, noting that "whichever of the two PRs merges second will have nothing to do in these files". #2514 merged on `master` a few hours ago, and the copy here had drifted from what actually landed — `flex` where master now has `flex-wrap`, the arrow guard on `navigation` where master uses `selected.rows.length`, and a scroll fade that #2514 dropped before merging. Following the intent above, `profile-activity-section.tsx` and its test are now byte-identical to `master` on #2530, so those files leave the diff entirely.

**Three findings from a review of the combined branch are fixed on #2530**, all in code from this PR:

- The hero derived its grid template from `hasVerdict`, which cannot be true until the response counts answer — so every claim painted one-column and then reflowed to `1fr 220px`, re-wrapping the title at the top of the page on every load. The template now reserves both tracks while the counts are out and gives the second back only on a settled zero.
- The hero pins its parts to explicit rows so the verdict can start on the title's, but row 1 belongs to the controversial/topics strip, which draws for almost nothing while `SHOW_HERO_TOPICS` is off. With it empty the rows still started at 2, leaving `claim-card-narrow:gap-y-4` — 16px — above the claim in the side panel and at phone widths. Measured 16px → 0px at 420px.
- `fullWidth` on `DebateExploreFeedCard` dropped the `max-w` *and* the `clamp(320px, calc(83dvh - 178px), 560px)` viewport-height budget. On the claim page's 840px column that made a debate ~1140px tall (two `aspect-480/289` tiles at ~800px), so on a 900px viewport the second debater and the interaction bar couldn't be seen together. It now raises the ceiling to the container instead of removing it — 800px at a 1300px viewport, 569px at 900px — so it still fills the wider column when there's room for it.

The hero, the responder-list changes, the `ClaimVerdict` deletion and the Topics tab are otherwise unchanged. Please do review #2530 — the hero is your design and the calls above are ones you may want to take differently.

PR #2536, opened 2026-09-22
fix(debates): rebuild a recording whose media pipeline dies

jwalkingjew:
## The bug

One debater's video never loads in the explore feed, while the same debate plays fine full screen.

Reproduced on the master preview against the AI space's **"Generative AI is making culture less original"** — the exact card in the report, blank top tile and all:

Instrumenting the page:

```
23765 suspend   1790097849448  rs4 t=0
23766 error     1790097849448  rs4 t=0  code 2
          PipelineStatus::PIPELINE_ERROR_READ: FFmpegDemuxer: demuxer seek failed
```

Slot 1 dies at `t=0`, **before either element has played** and before anything in the app seeks it. Slot 2 plays on beside it; the player's `data-debate-*` attributes read `ready=true playing=true`, because nothing noticed.

## Why

The recording is fine. A detached `<video>` handed the same signed URL:

```
loaded:            rs3 t=0.00  err=-  vw=1280
played 3s:         rs4 t=2.96  err=-  vw=1280
seeked to 20, +3s: rs4 t=22.63 err=-  vw=1280
```

What kills it is the element sitting in the feed's look-ahead preload long enough for Chrome to suspend its fetch and then resume it — a cue-less MediaRecorder WebM, so the resume is a demuxer seek, and it fails. Full screen is fine for the same debate because a card there is reached within seconds of being mounted, so that window barely exists.

Nothing recovers it, and nothing could: a `<video>` that reports an error is finished, and every correction in `useDebatePlayback` is about *where* the two elements are, not whether one of them is still alive.

## The fix

1. **`DebaterVideo` rebuilds the element on `error`** — detach, re-attach, `load()`. Verified in the live browser to take the dead element from `error.code === 2` back to `HAVE_ENOUGH_DATA`. `load()` alone cannot: resource selection would run against the src it already has. Bounded to three attempts per source, spaced, so a genuinely unplayable recording cannot loop on a multi-megabyte fetch.
2. **`resyncSlot` rejoins the pair** at the partner's position — the only remaining record of it — and only once the rebuilt element reports `loadeddata`. Decoded data, not merely a header: these files carry no Cues, so a seek is a parse walk (GEO-2828), and one made over data the element does not hold is how the pipeline died to begin with. A paused partner (the observed preload case) is left alone entirely; the card's next `resumeBoth` aligns both anyway.
3. **`pairIsSettled` stands down during a rebuild**, as it already does for a resume in flight. A rebuild is a split pair by construction — detaching the source pauses that element while its partner plays — and the existing correction reads a split pair as "the browser stopped playback on us": it would pause the healthy half and record a *user* pause, trading a blank tile for a stopped debate.

## What a failed rebuild must not do

A rebuild can fail as often as it is allowed to, and what is left then is an element that cannot play and cannot be made to. Chrome leaves such an element `paused === false` — the accident that let the original bug present as *one blank tile beside a debate still playing* — and detaching the source to rebuild it flips that to `true`. The pair's corrections would then read the split as "the browser stopped playback on us", pause the healthy half and record a **user** pause: a blank tile traded for a stopped debate, strictly worse than the bug.

So an element holding a `MediaError` is excluded from the play/pause correction, and the turn gate reads whichever element can still speak for playback — the half that works keeps its audio.

Two more, in the same spirit:

- The guard is **held until the rebuilt element is running**, not released at the seek. `play()` does not start an element synchronously, so a tick landing between the call and the first frame saw exactly the split pair the guard exists to hide — the reason `resumesInFlightRef` already spans its confirm window.
- A rebuild can now be **called off**. It outlives the render that started it, and the feed keys its cards by claim, so a re-rank hands a different debate to the same hook and the same `<video>` node while one is pending — that rejoin would have fired against the new recording, seeked it with the *previous* debate's offsets and started it, holding the new debate's pair unpoliced until the backstop. Cancelled on a recordings change and on unmount, and released the moment the element reports another error rather than waiting out the 15s backstop.

## When the rebuild is not enough

`error.code === 2` is `MEDIA_ERR_NETWORK`, and it covers a signature that has lapsed as well as the pipeline death diagnosed above. Every rebuild re-fetches the same bytes from the same signature, so it provably cannot answer the first — a card left open long enough would spend its whole budget deterministically and give up on a recording that is still there.

So the budget running out **escalates once to a freshly signed URL** (`refreshSlotUrl`) rather than giving up. One slot, not the pair: the pair-level fetch opens by blanking both URLs, which would drop the healthy tile to "Loading…" and release a video the viewer is watching. Capped at one re-sign per recording per card, so a recording that is simply gone cannot mint a URL per round.

**And past that the tile says so.** Before this existed the state was undetectable, so silence was the only option; it is detected now, and a blank half of a playing debate that accounts for itself in no way is the report that opened this ticket. Verified on this PR's preview by pointing one recording at a URL that can never load and letting the real path run to the end:

```
{"label":"Retry Bernie H>'s video","copy":"This recording didn't load"}
```

The same round `RetrySmall` control the player uses for replay, with its own label, over the scrim and the debater's name so it is clear *whose* video is missing — and a viewer's tap is worth a fresh budget.

## Tests

22 new tests, all of which fail with the fix backed out (verified by mutation):

- `debate-feed-player.test.tsx` — re-attaches and asks the controller to resync; each tile independently; bounded attempts; no stacked repairs; drops a pending repair when the tile is handed a different recording; escalates to a freshly signed URL once the budget is spent, exactly once; spends a fresh budget on the re-signed recording; offers the viewer a retry once it has run out of its own, and not over a recording that repaired itself.
- `use-debate-playback.test.tsx` — seeks the rebuilt recording to its partner and starts it; leaves a paused pair alone; waits for decoded data; **does not let the split pair pause the half that is still playing**; holds the guard until the rebuild is actually running; polices the pair again once it is; lets go as soon as the rebuild fails again; drops a pending rejoin when the card is handed a different debate; **leaves the healthy half playing when a recording cannot be revived**; guards the rebuild even while the partner is paused (so `resumeBoth` doesn't report "Could not play both videos" about an element that is a second from opening); re-signs one recording without disturbing the other, once per card, and afresh for a different debate.

`bun vitest run core/debates/ core/utils/video` → 1842 passed. `tsc` and `eslint` clean on the changed files.


ohohoreilly (commented):
This is good work. The diagnosis is properly evidenced — a detached element loading the same URL to `HAVE_ENOUGH_DATA` is what turns "the video is broken" into "the element is broken", and that distinction is the whole PR. The test names map one-to-one onto the claims in the description, including the subtle ones (`holds the guard until the rebuilt recording is actually running`, `drops a pending rejoin when the card is handed a different debate`), and mutation-verifying them is the right bar.

The sharpest thing here is this:

> Before the rebuild existed this was true by accident: Chrome leaves a failed element `paused === false` ... Detaching the source to rebuild it flips that to `true`, so what was accidental has to be said.

That's the observation that would have bitten, and it's the kind that only turns up by actually watching the element. Trading a blank tile for a stopped debate would have been a worse bug than the one being fixed, reported as a regression by someone who never saw the blank tile.

Three things, none blocking.

---

## 1. One class of `code 2` the rebuild provably cannot fix

`error.code === 2` is `MEDIA_ERR_NETWORK`, which covers the demuxer read failure you diagnosed **and** a signed URL that has stopped working. The rebuild re-fetches the *same* URL — `getRecordingPlaybackUrl` runs once per `(debate.id, filenames)` effect and nothing re-mints it — so an expired presign fails all three attempts deterministically and lands in the "genuinely unplayable" bucket.

The presign is 6 hours (`MEDIA_CONTENT_PRESIGN_EXPIRES` in geo-chat), so this is rare rather than likely, and it is clearly **not** what you reproduced. But it's the one failure mode the repair can't reach, and a long-lived tab in a feed is exactly where it would show up. Worth one line in the `MAX_MEDIA_RECOVERY_ATTEMPTS` comment saying the budget covers transient pipeline death and not a dead URL — or, if you want it to cover both, having the final failed attempt invalidate the URL query rather than re-fetching the same bytes.

## 2. After the budget is exhausted, the viewer is told nothing

> A genuinely unplayable source exhausts the budget in about a second and is then left alone, showing what it showed before this existed.

Which is a blank tile beside a playing debate, silently. This file already imports `RetrySmall` — is there a reason an exhausted rebuild shouldn't surface that affordance, so the state is legible and manually recoverable? Happy to hear "out of scope, separate ticket"; I mention it because the PR moves this from *undetectable* to *detected and deliberately not shown*, and once we know, not saying becomes a choice.

## 3. `clearTimeout(giveUpTimer)` reads a `const` declared below it

```ts
const release = () => {
  ...
  clearTimeout(giveUpTimer);   // ← TDZ if release() ever runs before the line below
  ...
};
recoveryReleasesRef.current.add(release);
function rejoin() { ... }
const giveUpTimer = setTimeout(release, RECOVERY_REJOIN_TIMEOUT_MS);
```

Safe as written — nothing invokes `release` between the `add` and the `setTimeout`, and `rejoin()` is only called after. But `release` is in a shared `Set` before `giveUpTimer` exists, so the safety is a property of statement order rather than of the code. Anyone adding an early bail — or hoisting the `add` — gets a `ReferenceError` inside a cleanup path, which is a miserable place to find one. `let giveUpTimer: ReturnType<typeof setTimeout>` above `release`, or moving the `setTimeout` up, costs nothing.

---

## One question I could not settle

`resyncSlot` returns early when the partner is paused, **before** incrementing `recoveringSlotsRef` — so during a preload-case rebuild the pair is unguarded. That's right while both are paused, since there's no split to misread.

What I couldn't convince myself of either way: a rebuild is in flight with the partner paused (no guard), and then the card goes active and `resumeBoth` fires mid-load. I *think* it's fine, because `resumeBoth` calls `play()` on both elements and `play()` flips `paused` to `false` synchronously even while the element is still opening — so there's no split for the correction to see. But that rests on `resumeBoth` never skipping the element that is mid-rebuild, and I didn't trace every path into it.

Posing it as a question rather than a finding: is that window real, and if so does it want the guard taken before the paused-partner early return rather than after?

jwalkingjew:
Thanks — all three taken, and the question turned out to have a real answer. Pushed in ee51b7a.

### 1. The dead-URL class

Taken the stronger option rather than the comment. Exhausting the budget now escalates once to `refreshSlotUrl`, which re-signs **that one recording**; the new `src` resets the tile and the budget is spent again on a source that is genuinely new.

One slot rather than the pair, because the pair-level fetch is the wrong instrument here: `fetchedForRef` guards it against re-fetching on every re-activation (GEO-2895) and the effect behind it opens by blanking both URLs — which would drop the healthy tile to "Loading…" and release a video the viewer is watching. Nothing about a dead recording is a reason to interrupt the live one.

Capped at one re-sign per recording per card, reset with the recordings, so a recording that is simply *gone* can't mint a URL per round — the tile spends a fresh budget on every new `src`, which would otherwise make the two ceilings cancel each other out.

I've deliberately not put the 6-hour number in the comment: I can't verify `MEDIA_CONTENT_PRESIGN_EXPIRES` from this repo, and a number that drifts is worse than none. The comment says what's true from here — `code === 2` covers both, and re-fetching the same signature can only answer one of them.

### 2. Telling the viewer

You're right that this is the part that changed: it was undetectable, so silence was the only option; now it's detected, and silence is a choice. Implemented rather than ticketed.

An exhausted tile shows the same round `RetrySmall` control the player uses for replay — stacked with a line reading "This recording didn't load", over the scrim and the debater's name so it's clear *whose* video is missing. A viewer's tap resets the budget and rebuilds.

Verified on this PR's preview by pointing one recording at a URL that can never load and letting the real path run end to end — three rebuilds, the escalation, then the control:

```
{"label":"Retry Bernie H>'s video","copy":"This recording didn't load"}
```

I also split `PLAYBACK_CONTROL_CIRCLE_CLASS` out of `CENTERED_PLAYBACK_CONTROL_CLASS` so the two round controls can't drift apart.

### 3. `giveUpTimer`

Fixed — `let giveUpTimer` above `release`, with a comment saying why it's a `let`, since the obvious tidy-up is to make it a `const` again. (`prefer-const` does fire on the uninitialised form, so it's `= undefined`.)

### The question: yes, the window is real — but not for the reason either of us was looking at

Your reasoning about the split pair is right, and I checked it: `play()` sets `paused` to `false` synchronously, before any media is ready, so there's no split for the correction to see and no path into `resumeBoth` skips the element mid-rebuild.

What is exposed is the layer above. `playBothWithMutedFallback` doesn't ask whether the elements are *paused*, it asks whether they're **running** — and a rebuilt cue-less recording is the slowest thing in that file to open. So the confirm poll times out, no refusal is recorded, both elements are muted, and it returns `'blocked'` → **"Could not play both videos. Try Play again."** on a card that is a second away from playing. Which is this ticket's own bug wearing different clothes: the app reporting a fault that isn't one, about a video that is fine.

So the guard is now taken **before any early return**, including the paused-partner case — as you suggested, though for the second reason rather than the first. It's a fact about the pair whatever the partner is doing, and `resumeBoth` reads it: on `'blocked'` with a rebuild in flight it suppresses the sentence and returns. The retry that outcome exists to allow still happens — the autoplay effect re-enters on `!playing` exactly as before — so the only thing removed is the claim that something went wrong.

Pinned by `guards the rebuild even while the partner is paused`, which fails (error surfaced) with the suppression backed out.

---

22 new tests now, each mutation-verified. `core/debates/ core/utils/video` → 1842 passed; tsc and eslint clean.

One thing worth flagging separately: this heals the symptom, not the cause. Writing Cues into these recordings would make every seek an index lookup instead of a parse walk, which would also retire the GEO-2828 drift workarounds. Happy to open that as its own ticket if it's not already somewhere.

PR #2546, opened 2026-09-23
fix(debates): recover a recording that dies on a metadata preload

jwalkingjew:
## The report

In Yaniv's personal space, one of the two debate cards on Overview shows **"This recording didn't load"**. Opening the same debate from the claim page plays it in both tiles.

## What is actually happening

`preload="metadata"` is what kills these recordings.

Debate recordings are MediaRecorder WebM — no duration, no cues — so the demuxer has to seek to work out how long the file is, and in metadata mode Chrome has already stopped fetching by the time it asks. The element dies with `PIPELINE_ERROR_READ: FFmpegDemuxer: demuxer seek failed` on a URL that is perfectly good.

Playing a tile lifts it out of metadata mode, which is why the failure only ever appears on tiles nobody is watching. That is the whole difference between the two pages:

- **Claim page** — one debate in the gallery, so it is the one the playback gate allows. It plays, so it loads.
- **Profile Overview** — two debates, the gate allows one, and the other sits at preload forever. Deterministic failure.

Reproduced against production with Playwright, then isolated to a bare page with a single `<video>` and no app code at all:

| preload | result |
|---|---|
| `metadata` | `PIPELINE_ERROR_READ` — 3/3 runs |
| `auto` | loads, buffers 2.2s, suspends |
| `none` | nothing loads (as asked) |

## Why the existing rebuild never healed it

GEO-2985 added a bounded rebuild for exactly this error, and it could never work: `reattachVideoSource` re-attached the source with the same `preload="metadata"`, so every rebuild reproduced the failure byte for byte. Three attempts, one re-signed URL, three more, then the message. The trace on the live page shows all seven.

## The fix

Raise `preload` to `auto` before the re-attach. The same element loads to `HAVE_ENOUGH_DATA` on the first rebuild.

Measured against the 89MB slot-1 recording of debate `01a0ca25`:

- rebuild keeping `metadata` — fails 3/3, budget exhausted, tile shows the message
- rebuild raising to `auto` — recovers on attempt 1
- the heavier fetch (~4.7MB against ~2MB) is paid only by recordings that actually failed

The raise lives at the call site in `rebuild`, next to the attempt budget and the backoff, rather than inside `reattachVideoSource` — it is a fact about these recordings, not about re-attaching a video source, and `processed-debate-player.tsx` deliberately uses `preload="none"`.

Nothing puts `preload` back, and that is deliberate — see the comment on the reset effect. A genuinely different recording never reaches a live element: `useDebatePlayback` blanks both URLs before fetching the new pair, which unmounts the `<video>` and takes the raised `preload` with it, so the replacement is built from the JSX default. The only `src` change a live element sees is `onExhausted` re-signing *this* recording — the one that has just failed every rebuild it was allowed under `metadata` — and handing that fresh URL the same mode would buy another wasted attempt and another blank half-second.

### The trade-off, stated plainly

Idle tiles still start at `metadata`, so a recording of this kind pays one wasted ~2MB load and about 400ms of grey before it heals. The alternative is `preload="auto"` everywhere, which removes that beat entirely but costs every idle tile ~4.7MB against ~2MB — and the Overview gallery mounts up to 6 cards, so 12 videos, 56MB against 24MB. Escalating on failure keeps the cost on the recordings that need it. Easy to flip if the flash matters more.

Scoped to the one site that plays raw `.local.webm`. The other `<video preload="metadata">` (`debate-coordinator.tsx:485`) plays a transcoded MP4, which carries its duration in the moov atom and never needs the seek that fails.

## Verified end to end

Built locally and loaded Yaniv's actual profile, repeatedly — which card wins the playback gate varies per load, and the bug only shows when the bad recording lands on the idle one. On the runs where it does:

```
4438  01a0ca25/slot-1  error    preload=metadata  err=2   ← the reported failure
4859  01a0ca25/slot-1  canplay  preload=auto              ← healed in 421ms
FINAL  [metadata, metadata, auto, metadata]   no errors   FAILED TILES: 0
```

Only the tile that failed carries `preload=auto`; the other three keep the cheap fetch.

## Tests

Three cases in `debate-feed-player.test.tsx`, each failing without its half of the change:

- raises preload past the mode that killed the pipeline, and leaves the healthy tile alone
- keeps the raised preload across a re-signed URL
- starts a different recording from a fresh element

`tsc --noEmit` and `eslint` clean; `core/debates`, `core/utils/video` and `partials/explore` pass (127 files, 2063 tests).


jwalkingjew:
## Re: Copilot review (`be489c33d`)

No inline threads were posted — all three points landed in the overview only, so replying here. Addressed in `c8e7b1e4a`.

---

### ✅ "Two comments need corrected cross-references and writer attribution" — right, both fixed

Both had gone stale in `be489c33d`, the commit that moved the preload raise from `reattachVideoSource` to the call site. I enumerated every cross-reference in the diff rather than fixing just what was flagged; these were the two:

1. The `rebuild` docblock sent the reader to "`reattachVideoSource`, which carries the measurement" — the measurement moved into `rebuild` in that same commit.
2. The reset effect said "`reattachVideoSource` writes the DOM property directly" — it stopped writing it in that same commit. `rebuild` does.

Exactly the class of defect the second review round caught (a stale comment whose obvious resolution re-adds the bug), so thanks for catching the repeat.

---

### ⚠️ "The regression test does not enforce the operation ordering required for the fix" — right that nothing was pinned, wrong about what

The stated requirement — preload raised *before* `load()` runs — is not what the fix depends on. Measured in Chrome against the 89MB slot-1 recording of debate `01a0ca25`, which is the one that fails:

| when `preload` is raised | result |
|---|---|
| sync, before the re-attach (what this PR does) | recovers |
| sync, **after** both `load()` calls | recovers |
| deferred one task (`setTimeout 0`) | recovers |
| deferred 500ms | **fails** — `error.code === 2` returns |
| never (what GEO-2985 shipped) | **fails** |

`load()` queues resource selection and awaits a stable state, so a same-tick write after the call is still picked up. The real constraint is *promptness*, not position.

But the instinct was correct: there **is** a timing property the fix depends on, and no test held the fix inside it. Every assertion looked at where `preload` landed, and it lands on `auto` however late it gets there — so the 500ms case would have gone green.

So the test is added, watching the property at the instant each `load()` runs. That is stricter than the browser demands, deliberately: nothing wants the raise anywhere else, the exact edge of the window is not worth depending on, and the drift it guards against — the raise migrating into an effect or a timer, which is where this PR once put a preload write — is invisible to every end-state assertion.

Verified by breaking it: reversing the two lines fails the new test and leaves the other 29 green.

The comments now state what was measured rather than what the spec implies — the previous wording ("the mode is read as resource selection begins") was the same overclaim in the source.

---

### Not changed

Idle tiles still start at `preload="metadata"`, so a recording of this kind pays one wasted ~2MB load and ~400ms of grey before healing. Raising it everywhere would cost every idle tile ~4.7MB against ~2MB, and the Overview gallery mounts up to 6 cards. Deliberate trade-off, documented in the PR body.

